PR #1, opened 2024-02-06
Adding GitHub Actions for building/testing/linting and deploying code documentation

fosdickio:
This PR adds some additional code documentation to `base/src/actions.rs` and adds 2 new GitHub Actions.  Those GitHub Actions are responsible for:

1. Building, testing, and linting the repository when any pushes or PRs occur on the `main` branch.  This will ensure that any failing tests are caught before being pushed.
2. Automatically uploading documentation generated by `cargo doc --no-deps` to GitHub Pages to serve as a static website.  This will allow contributors to quickly view up-to-date documentation.

In order to get GitHub Pages to serve the new documentation that is being pushed to the `gh-pages` branch, we need to also:

1. Go to the "Pages" section for the repository's settings on GitHub.
2. Select the `gh-pages` branch as the source for the GitHub Pages site.
3. Save the changes.

NOTE: The workflow uses `secrets.GITHUB_TOKEN` for authentication, which is automatically provided by GitHub Actions and does not need manual setup.

### DIFF
--- a/.github/workflows/deploy-cargo-docs.yaml
+++ b/.github/workflows/deploy-cargo-docs.yaml
@@ -20,7 +20,7 @@ jobs:
                   override: true
 
             - name: Generate Documentation
-              run: cargo doc --no-deps
+              run: make docs
 
             - name: Deploy to GitHub Pages
               uses: peaceiris/actions-gh-pages@v3
--- a/.github/workflows/deploy-cargo-docs.yaml
+++ b/.github/workflows/deploy-cargo-docs.yaml
@@ -1,0 +1,29 @@
+name: Deploy Docs to GitHub Pages
+
+on:
+    push:
+        branches:
+            - main
+
+jobs:
+    deploy-docs:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout Repository
+              uses: actions/checkout@v2
+
+            - name: Install Rust
+              uses: actions-rs/toolchain@v1
+              with:
+                  toolchain: stable
+                  profile: minimal
+                  override: true
+
+            - name: Generate Documentation
+              run: cargo doc --no-deps
+
+            - name: Deploy to GitHub Pages
+              uses: peaceiris/actions-gh-pages@v3
+              with:
+                  github_token: ${{ secrets.GITHUB_TOKEN }}
+                  publish_dir: ./target/doc
--- a/.github/workflows/rust-build-test.yaml
+++ b/.github/workflows/rust-build-test.yaml
@@ -1,0 +1,26 @@
+name: Rust
+
+on:
+    push:
+        branches: ["main"]
+    pull_request:
+        branches: ["main"]
+
+env:
+    CARGO_TERM_COLOR: always
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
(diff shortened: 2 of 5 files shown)


.github/workflows/deploy-cargo-docs.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+            - name: Install Rust
+              uses: actions-rs/toolchain@v1
+              with:
+                  toolchain: stable
+                  profile: minimal
+                  override: true
+
+            - name: Generate Documentation
+              run: cargo doc --no-deps
```
nhatcher: Can you run `make docs` instead? At this moment it is completely equivalent but in the future we might add things or options to the makefile

.github/workflows/rust-build-test.yaml
```diff
@@ -0,0 +1,26 @@
...
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v3
+
+            - name: Build
+              run: cargo build --release --verbose
+
+            - name: Run tests
+              run: cargo test --verbose
```
nhatcher: Can you run `make tests` here? Note that that will also run the linter and fail if there is a linting problem.

If you want you can add the `--verbose` flag to the Makefile

base/src/actions.rs
```diff
@@ -8,18 +8,36 @@ use crate::model::Model;
...
+    ///
+    /// * `sheet` - The sheet number to retrieve columns from.
+    /// * `row` - The row number to retrieve columns for.
+    /// * `descending` - If true, the columns are returned in descending order; otherwise, in ascending order.
+    ///
+    /// # Returns
+    ///
+    /// This function returns a `Result` containing either:
+    /// - `Ok(Vec<i32>)`: A vector of column indices for the specified row, sorted according to the `descending` flag.
+    /// - `Err(String)`: An error message if the sheet cannot be found.
```
nhatcher, line 35: Awesome!

base/src/actions.rs
```diff
@@ -54,18 +79,25 @@ impl Model {
...
 
+    /// Inserts one or more new columns into the model at the specified index.
+    ///
+    /// This method shifts existing columns to the right to make space for the new columns.
+    ///
+    /// # Arguments
+    ///
+    /// * `sheet` - The sheet number to retrieve columns from.
+    /// * `column` - The index at which the new columns should be inserted.
+    /// * `column_count` - The number of columns to insert.
```
nhatcher, line 95: This is great! 
Note that this is a public function and the sort of functions we should be documenting first.

I started doing this at the lexer:

https://github.com/ironcalc/IronCalc/blob/67e8b0ce145930b02048e2e5da4ed92384c9f859/base/src/expressions/lexer/util.rs#L17-L49

NB: no action required, we will add examples in a different PR

nhatcher (changes requested):
Hi @fosdickio,

Thanks for taking care of this!. I left you a couple of questions before merging.

base/src/actions.rs
fosdickio: Sounds good.  I'll focus on the public functions initially then. 🙂

PR #4, opened 2024-02-14
Adding additional lexer unit tests

fosdickio:
Adding additional unit tests to the lexer to improve test coverage gaps outlined in [Codecov](https://app.codecov.io/gh/ironcalc/IronCalc/blob/main/base%2Fsrc%2Fexpressions%2Flexer%2Fmod.rs).

### DIFF
--- a/base/src/expressions/lexer/test/test_common.rs
+++ b/base/src/expressions/lexer/test/test_common.rs
@@ -4,7 +4,7 @@ use crate::language::get_language;
 use crate::locale::get_locale;
 
 use crate::expressions::{
-    lexer::{Lexer, LexerMode},
+    lexer::{Lexer, LexerError, LexerMode},
     token::TokenType::*,
     token::{Error, OpCompare, OpProduct, OpSum},
     types::ParsedReference,
@@ -412,6 +412,20 @@ fn test_name_r1c1p() {
 }
 
 #[test]
+fn test_reference_r1c1_error() {
+    let mut lx = new_lexer("$A$4", false);
+    lx.mode = LexerMode::R1C1;
+    assert_eq!(
+        lx.next_token(),
+        Illegal(LexerError {
+            position: 1,
+            message: "Cannot parse A1 reference in R1C1 mode".to_string(),
+        })
+    );
+    assert_eq!(lx.next_token(), EOF);
+}
+
+#[test]
 fn test_name_wrong_ref() {
     let mut lx = new_lexer("Sheet1!2", false);
     assert!(matches!(lx.next_token(), Illegal(_)));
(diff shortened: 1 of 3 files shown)


base/src/expressions/lexer/test/test_common.rs
```diff
@@ -411,6 +411,20 @@ fn test_name_r1c1p() {
     assert_eq!(lx.next_token(), EOF);
 }
 
+#[test]
+fn test_reference_r1c1_error() {
```
nhatcher, line 415: Halfway through the code I changed my mind and I don't call test functions these days with names starting with `test`. I feel it is a little bit too much.

base/src/expressions/lexer/test/test_ranges.rs
```diff
@@ -191,7 +224,7 @@ fn test_range_column_absolute2() {
                 row: LAST_ROW,
                 absolute_column: true,
                 absolute_row: true,
-            }
+            },
```
nhatcher: I am confused, why was the linter passing on this?

base/src/expressions/test.rs
```diff
@@ -10,6 +10,11 @@ fn test_error_codes() {
         Error::NA,
         Error::NUM,
         Error::ERROR,
+        Error::NIMPL,
+        Error::SPILL,
+        Error::CALC,
+        Error::CIRC,
+        Error::NULL
```
nhatcher, line 17: Awesome, thanks!

nhatcher (changes requested):
Quick question, are tests passing in this branch?

fosdickio:
> Quick question, are tests passing in this branch?

@nhatcher Yes, the tests are passing locally.

```sh
test result: ok. 445 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 1.56s
```

base/src/expressions/lexer/test/test_ranges.rs
fosdickio: I am also curious about this, so I just did some testing.  `cargo fmt --` actually treats both as valid.  However, the integration in my IDE preferred trailing commas.  I've since changed some settings and it no longer seems to be preferring that format, so we should hopefully be fine moving forward.  I'm happy to revert the change back if you prefer.

base/src/expressions/lexer/test/test_common.rs
fosdickio: I think that makes sense with the `#[test]` macro above it.  I'll leave the `test_` prefix off moving forward.  I can quickly remove the prefix in the file altogether if you prefer?

base/src/expressions/lexer/test/test_ranges.rs
nhatcher: TIL! I was under the impression `cargo fmt` would allow one or the other. Yes, please revert those changes for now and lets make a more conscious decision on hour next call.

base/src/expressions/lexer/test/test_common.rs
nhatcher: It's ok in this file. I was just leting you know of my preferences. It's not a big deal.

nhatcher (approved):
Nice!

PR #5, opened 2024-02-14
Adding some additional lexer documentation

fosdickio:


fosdickio:
This is odd.  I ran `cargo fmt --` and now it thinks code coverage has decreased. 😒

nhatcher:
> This is odd. I ran `cargo fmt --` and now it thinks code coverage has decreased. 😒

You added more lines. what I don't understand is why the linting test was passing before

PR #6, opened 2024-02-20
Adding new example counting formulas and errors

fosdickio:
## Summary
I'm adding a new example to check whether a cell contains an error or formula.  I found this missing when trying to work through an example to read in a workbook and do something simple (count the total number of cells, formulas, and errors).

## External Usage
I originally worked on this from a separate project and pulled in the dependency to `Cargo.toml` like so (similar to `TironCalc`):

```toml
[dependencies]
ironcalc = { git = "https://github.com/ironcalc/IronCalc", version = "0.1.0"}
```

From there, I was able to load a local `.xlsx` file and perform the operation on it (`cargo run test.xlsx`).

```rust
use ironcalc::import::load_model_from_xlsx;

fn main() {
    let args: Vec<String> = std::env::args().collect();
    if args.len() != 2 {
        panic!("Usage: {} <file.xlsx>", args[0]);
    }
    let file_name = &args[1];

    let mut model = load_model_from_xlsx(file_name, "en", "UTC").unwrap();
    model.evaluate();
    let cells = model.get_all_cells();

    let mut cells_count = 0;
    let mut formula_count = 0;
    let mut error_count = 0;

    for cell in cells {
        if model.cell_contains_formula(cell.index, cell.row, cell.column) {
            formula_count += 1;
        }

        if model.cell_contains_error(cell.index, cell.row, cell.column) {
            error_count += 1;
        }

        cells_count += 1;
    }

    println!("Cell count: {}", cells_count);
    println!("Formula count: {}", formula_count);
    println!("Error count: {}", error_count);
}
```

## Other Considerations
I've moved this example into `base`, as it is an example mostly demonstrating functionality on `ironcalc_base::model::Model`.

## Testing
This code was ran locally and tested using `make tests` without error.


base/examples/formulas_and_errors.rs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        cells_count += 1;
+    }
+
+    assert_eq!(cells_count, 5);
+    assert_eq!(formula_count, 2);
+    assert_eq!(error_count, 1);
+
+    Ok(())
```
nhatcher, line 45: Great example!

base/src/model.rs
```diff
@@ -1172,6 +1172,18 @@ impl Model {
         }))
     }
 
+    /// Returns a boolean representing whether the cell contains a formula or not.
+    pub fn cell_contains_formula(&self, sheet_index: u32, row: i32, column: i32) -> bool {
```
nhatcher: I would prefer if you can add the example without having to add a method in the public API. This is something I would like to discuss with you, but I would like the public API to be as minimal as possible.

base/src/model.rs
```diff
@@ -1172,6 +1172,18 @@ impl Model {
...
+        let cell = &self.workbook.worksheets[sheet_index as usize].sheet_data[&row][&column];
```
nhatcher: Although I do this quite a lot, we shouldn't.  If the index is wrong this will panic. We should do things like:

```rust
let cell = self.workbook.worksheet(sheet)?.cell(row, column);
```
That will return an error if `(sheet, row, column)` are not valid parameters.
We shouldn't do this in any case, but we should specially avoid it in the public API.
So basically not call to the public API should be able to panic.

nhatcher (changes requested):
Thanks! 

I would prefer if we don't add members to the public API.

Once you are do, we can add this example to the documentation like in:
https://github.com/ironcalc/IronCalc/blob/main/base/src/lib.rs#L20

nhatcher:
> ## Summary
> 
> I'm adding a new public API endpoint to check whether a cell contains an error or formula. I found this missing when trying to work through an example to read in a workbook and do something simple (count the total number of cells, formulas, and errors). I'm adding an example to accompany it.
> ## External Usage
> 
> I originally worked on this from a separate project and pulled in the dependency to `Cargo.toml` like so (similar to `TironCalc`):

That's a nice idea but what you are doing here is better.
> 
> ```toml
> [dependencies]
> ironcalc = { git = "https://github.com/ironcalc/IronCalc", version = "0.1.0"}
> ```

I think you can drop the version (my mistake!)
> 
> From there, I was able to load a local `.xlsx` file and perform the operation on it (`cargo run test.xlsx`).
> 
> ```rust
> use ironcalc::import::load_model_from_xlsx;
> 
> fn main() {
>     let args: Vec<String> = std::env::args().collect();
>     if args.len() != 2 {
>         panic!("Usage: {} <file.xlsx>", args[0]);
>     }
>     let file_name = &args[1];
> 
>     let mut model = load_model_from_xlsx(file_name, "en", "UTC").unwrap();
>     model.evaluate();
>     let cells = model.get_all_cells();
> 
>     let mut cells_count = 0;
>     let mut formula_count = 0;
>     let mut error_count = 0;
> 
>     for cell in cells {
>         if model.cell_contains_formula(cell.index, cell.row, cell.column) {
>             formula_count += 1;
>         } else if model.cell_contains_error(cell.index, cell.row, cell.column) {
>             error_count += 1;
>         }
> 
>         cells_count += 1;
>     }
> 
>     println!("Cell count: {}", cells_count);
>     println!("Formula count: {}", formula_count);
>     println!("Error count: {}", error_count);
> }
> ```
> 
> ## Other Considerations
> 
> I've moved this example into `base`, as it is an example mostly demonstrating functionality on `ironcalc_base::model::Model`.
> ## Testing
> 
> This code was ran locally and tested using `make tests` without error.

base/examples/formulas_and_errors.rs
nhatcher, line 45: If you add this code in the documentation it will ran with `cargo tests`

fosdickio: Good to know!  Thanks for that.  I've updated `base/src/lib.rs` accordingly.

fosdickio:
@nhatcher This is ready for review again based on your feedback.  You might want to squash-and-merge this one, as there are quite a few small little changes.

nhatcher (approved):
Awesome, this is a useful example.

Issue #13, opened 2024-03-13
Garbage collector


nhatcher:
Currently some operations like adding and removing strings, adding and removing styles produces a lot of garbage.

To see a small example look at the branch:

https://github.com/ironcalc/IronCalc/tree/feature/nicolas-garbage-collector

There are other ways to achieve this. One would be to clean after ourselves each time we do anything that could potentially create garbage:

* delete cells, columns, rows
* change cells
* change styles
* change columns width, rows height,
* ....

jaycarlton:
It would be nice to be able to state categorically that when a user deletes a cell or its contents, that the information isn't retained outside of Undo history (and have a way to clear that history for sensitive documents, if it's persisted).

PR #18, opened 2024-03-13
Adding garbage collector for `model.shared_strings`

fosdickio:
## Background
`model.shared_strings` is an optimization for faster string look ups.  It is a mapping from a string value to a reference in memory.  Multiple cells that contain the same string value all use the same reference. 

## Change Summary
This change adds a garbage collector for `model.shared_strings`.  When cells are deleted using `model.delete_cell`, they are not removed from `model.shared_strings`.  This change cleans up unreachable string references when `model.evaluate` is run.  Before deleting a string, we must check the entire worksheet to make sure no other cells contain the deleted value before it can be deleted.

## Testing
- New tests were added in `base/src/test/test_garbage_collector.rs`
- The test suite was ran locally prior to submission using `make tests`

## Follow-Up Work
There is additional garbage collection (e.g., styles clean-up, etc.) that can be done in future PRs.

## References
- GitHub Issue: https://github.com/ironcalc/IronCalc/issues/13
- Branch with skeleton implementation: https://github.com/ironcalc/IronCalc/compare/main...feature/nicolas-garbage-collector

### DIFF
--- a/base/src/garbage_collector.rs
+++ b/base/src/garbage_collector.rs
@@ -7,19 +7,12 @@ impl Model {
     /// These include:
     /// - Shared strings that are no longer referenced by any cell
     pub fn garbage_collector(&mut self) -> Result<(), String> {
-        let cell_values = self.get_all_cell_values().unwrap();
-        let mut new_shared_strings = self.shared_strings.clone();
+        let cell_values = self.get_all_cell_values()?;
 
-        for (index, _) in self.shared_strings.iter() {
-            // A cell is referencing the string, so we continue
-            if cell_values.contains(index) {
-                continue;
-            }
-
-            new_shared_strings.remove(index);
-        }
-
-        self.shared_strings = new_shared_strings;
+        self.shared_strings.retain(|index, _| {
+            // A cell is referencing the string, so we keep it
+            cell_values.contains(index)
+        });
 
         Ok(())
     }
--- a/base/src/garbage_collector.rs
+++ b/base/src/garbage_collector.rs
@@ -1,12 +1,12 @@
 #![allow(clippy::unwrap_used)]
 
-use crate::model::Model;
+use crate::model::Model; // Add the missing error module
 
 impl Model {
     /// The garbage collector cleans up any lefover elements in the workbook that are no longer used.
     /// These include:
     /// - Shared strings that are no longer referenced by any cell
-    pub(crate) fn garbage_collector(&mut self) -> Result<(), String> {
+    pub fn garbage_collector(&mut self) -> Result<(), String> {
         let cell_values = self.get_all_cell_values().unwrap();
         let mut new_shared_strings = self.shared_strings.clone();
 
@@ -31,7 +31,7 @@ impl Model {
 
         for cell in cells {
             let cell_value = self.formatted_cell_value(cell.index, cell.row, cell.column);
-            cell_values.push(cell_value);
+            cell_values.push(cell_value?);
         }
 
         Ok(cell_values)
--- a/base/src/garbage_collector.rs
+++ b/base/src/garbage_collector.rs
@@ -30,9 +30,7 @@ impl Model {
         let mut cell_values = Vec::new();
 
         for cell in cells {
-            let cell_value = self
-                .formatted_cell_value(cell.index, cell.row, cell.column)
-                .unwrap();
+            let cell_value = self.formatted_cell_value(cell.index, cell.row, cell.column);
             cell_values.push(cell_value);
         }
 
--- a/base/src/garbage_collector.rs
+++ b/base/src/garbage_collector.rs
@@ -1,0 +1,39 @@
+use crate::model::Model;
+
+impl Model {
+    /// The garbage collector cleans up any lefover elements in the workbook that are no longer used.
+    /// These include:
+    /// - Shared strings that are no longer referenced by any cell
+    pub(crate) fn garbage_collector(&mut self) -> Result<(), String> {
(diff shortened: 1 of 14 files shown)


base/src/garbage_collector.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Returns a vector of all formatted cell values in the workbook.
+    fn get_all_cell_values(&self) -> Result<Vec<String>, String> {
+        let cells = self.get_all_cells();
+        let mut cell_values = Vec::new();
+
+        for cell in cells {
+            let cell_value = self
+                .formatted_cell_value(cell.index, cell.row, cell.column)
+                .unwrap();
```
nhatcher: We can't use unwrap in non test code. I will that to the linter, I removed it at some point.

base/src/model.rs
```diff
@@ -1645,6 +1645,8 @@ impl Model {
                 column: cell.column,
             });
         }
+
+        self.garbage_collector().unwrap();
```
nhatcher: We shouldn't call the gc on every evaluate call, this will make things extremely slow.
Also we shouldn't use `unwrap` in non test code as mentioned earlier.

You can use the gc when saving to disk or right before serializing to json.

base/src/garbage_collector.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use crate::model::Model;
+
+impl Model {
+    /// The garbage collector cleans up any lefover elements in the workbook that are no longer used.
+    /// These include:
+    /// - Shared strings that are no longer referenced by any cell
+    pub(crate) fn garbage_collector(&mut self) -> Result<(), String> {
+        let cell_values = self.get_all_cell_values().unwrap();
+        let mut new_shared_strings = self.shared_strings.clone();
```
nhatcher: We have two copies of the shared strings :/. This one that you are cloning and `self.workbook.shared_strings`.

The second one is just the list of shared strings and mimics the Excel structure. The first one `self.shared_strings` is just an optimization device. It was added in 67e8b0ce145930b02048e2e5da4ed92384c9f859.

Before fixing this issue could you add a test that will fail without the issue.

Also we might not need to clone the shared string. What is there is nothing to clean up?

You might not need this, but Rust offers some utilities to work in situations similar to those. It's called Clone On Write.

You can see an example in:

https://github.com/ironcalc/IronCalc/blob/main/xlsx/src/export/escape.rs
https://doc.rust-lang.org/std/borrow/enum.Cow.html

```
                self.workbook.shared_strings.push(value.to_string());
                self.shared_strings.insert(value.to_string(), string_index);
```

nhatcher (changes requested):
Hi @fosdickio,

I left a few comments, hopefully easy to address. I think we should do a PR with all the cases where memory should be freed.

On the top of my head:
* styles
* column structure (this is going to be a bit tough though)

xlsx/src/export/mod.rs
```diff
@@ -63,10 +64,14 @@ pub fn save_to_xlsx(model: &Model, file_name: &str) -> Result<(), XlsxError> {
     let writer = BufWriter::new(file);
     save_xlsx_to_writer(model, writer)?;
 
+    if let Err(err) = model.garbage_collector() {
```
fosdickio, line 67: Running the garbage collector clean up after saving of a file to avoid an issue where you can't save due to GC failure.

base/src/garbage_collector.rs
fosdickio: > Before fixing this issue could you add a test that will fail without the issue.

@nhatcher Will you elaborate on this?  What would this be testing exactly?

fosdickio:
> I think we should do a PR with all the cases where memory should be freed.

@nhatcher I was thinking it might be more clear to have separate PRs for garbage collection clean up of shared strings and style cleanup.  Are you suggesting having these in the same PR?

base/src/garbage_collector.rs
nhatcher: Hey, sorry I missed this. The idea is to write a test that will fail right now without writing any code. Then you fix the issue and make sure that the test is passing (a _regression_ test_. We can discuss this over a call if it is unclear.

nhatcher:
> > I think we should do a PR with all the cases where memory should be freed.
> 
> @nhatcher I was thinking it might be simpler to have separate PRs for garbage collection clean up of shared strings and style cleanup. Are you suggesting having these in the same PR?

Yeah, I was kind of thinking that we could only merge PRs that are feature complete. The problem with merging a PR that has missing features is that it is easy to put them aside and never et back to them. Let's talk about this on a call.

base/src/garbage_collector.rs
fosdickio: I think I might be missing something, as both of these tests should cover a regression:

- https://github.com/ironcalc/IronCalc/pull/18/files#diff-d3ca2f9ce2aa275260078785b08b1515090994045364f03035823d9f3fa52528R136-R154
- https://github.com/ironcalc/IronCalc/pull/18/files#diff-d738915e77f990f230abcc69bac45acb3e3b7536a5fbc395dc076ad7f27f8bb6R4-R31

fosdickio:
@nhatcher and I discussed this on a call.  Instead of having this as a small PR with only the shared strings cleanup, it should include all garbage collection functionality.  There are three additional things that should be included before the garbage collector functionality gets merged:

1. Shared styles
2. Shared formulas
3. [Optional] Cleanup of columns --> @nhatcher will file issue explaining in further detail

nhatcher:
Will need to be reimplemented

PR #19, opened 2024-03-13
Fixing failing test: `test_complex.fn_imcos`

fosdickio:
## Summary
This test was failing locally due to what looks to be a rounding error.  This fix changes the test to allow either value.

```
failures:

---- test::engineering::test_complex::fn_imcos stdout ----
thread 'test::engineering::test_complex::fn_imcos' panicked at base/src/test/engineering/test_complex.rs:82:5:
assertion `left == right` failed
  left: "-6.58066304055116+7.58155274274655i"
 right: "-6.58066304055116+7.58155274274654i"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test::engineering::test_complex::fn_imcos

test result: FAILED. 450 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 1.15s
```

### DIFF
--- a/base/src/test/engineering/test_complex.rs
+++ b/base/src/test/engineering/test_complex.rs
@@ -82,7 +82,7 @@ fn fn_imcos() {
     assert_complex_eq(
         &model._get_text("A1"),
         "-6.58066304055116+7.58155274274654i",
-        1e-14,
+        Some(1e-14),
     );
 }
 
--- a/base/src/test/engineering/test_complex.rs
+++ b/base/src/test/engineering/test_complex.rs
@@ -75,40 +75,20 @@ fn fn_imconjugate() {
 #[test]
 fn fn_imcos() {
     let mut model = new_empty_model();
-    model._set("A1", r#"=IMCOS("4+3i")"#);
+    model._set("A1", r#"0.11212+1.234523I"#);
+    model._set("A2", r#"=IMABS(IMCOS(3)-A1)"#);
 
     model.evaluate();
 
-    let cell_value = model._get_text("A1");
-    let (real_part, imag_part) =
-        parse_complex_number(&cell_value).expect("Failed to parse complex number");
+    let result_str = model._get_text("A2");
+    let float_value = match result_str.parse::<f64>() {
+        Ok(value) => value,
+        Err(err) => {
+            panic!("{}", err);
+        }
+    };
 
-    let expected_real = -6.58066304055116;
-    let expected_imag = 7.58155274274654;
-    let tolerance = 1e-14;
-
-    assert!(
-        (real_part - expected_real).abs() < tolerance
-            && (imag_part - expected_imag).abs() < tolerance,
-        "Expected approximately ({}, {}), got ({}, {})",
-        expected_real,
-        expected_imag,
-        real_part,
-        imag_part
-    );
-}
-
-fn parse_complex_number(s: &str) -> Result<(f64, f64), &'static str> {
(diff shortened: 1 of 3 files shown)


nhatcher:
Yes, I saw this. This depends on the processor architecture.

base/src/test/engineering/test_complex.rs
```diff
@@ -79,7 +79,11 @@ fn fn_imcos() {
 
     model.evaluate();
 
-    assert_eq!(model._get_text("A1"), "-6.58066304055116+7.58155274274654i");
+    let value = model._get_text("A1");
+    assert!(
+        value == "-6.58066304055116+7.58155274274654i"
+            || value == "-6.58066304055116+7.58155274274655i"
```
nhatcher: I think the correct fix would be to check on an approximation of the real and imaginary parts. This fix could get out of hand..

For an example see:

https://github.com/ironcalc/IronCalc/blob/main/base/src/test/engineering/test_complex.rs#L144-L162

nhatcher (changes requested):
Thanks for taking care of this, tests were failing on my M1 :/

nhatcher:
See also: d2bab6608bdd940d51ce520bb2a04d1f23973d60

base/src/test/engineering/test_complex.rs
fosdickio: Yikes!  I think this updated approximation version is even worse.  Got any suggestions?

https://github.com/ironcalc/IronCalc/pull/19/commits/e3c20b2ceaecb3d12d5f174d38b01480f5ece452

nhatcher: Hi, sorry for the late reply, the way to do that would be to test the function `=IMAGINARY(IMCOS(x))`

Or testing the absolute value is less than `f64::EPS`

nhatcher (changes requested):
Also the parse function wouldn't work if the second number is negative.
We could in the future have a utility `complex_numbers_are_close` or something

base/src/test/engineering/test_complex.rs
fosdickio: I think I'm missing something here when converting from a string to a float.  Updating to the following:

```rust
#[test]
fn fn_imcos() {
    let mut model = new_empty_model();
    model._set("A1", r#"0.11212+1.234523I"#);
    model._set("A2", r#"=IMABS(IMCOS(3)-A1)"#);

    model.evaluate();

    let result_str = model._get_text("A2");
    let float_value = match result_str.parse::<f64>() {
        Ok(value) => value,
        Err(err) => {
            panic!("{}", err);
        }
    };

    assert!(float_value < f64::EPSILON);
}
```

Results in:

```sh
running 1 test
thread 'test::engineering::test_complex::fn_imcos' panicked at base/src/test/engineering/test_complex.rs:87:13:
invalid float literal
```

nhatcher: What is happening is that `result_str` is `#VALUE!`. You can't use `+` and `-` to add or subtract complex numbers. In other words the arithmetic operators are not overloaded. Yo have to use functions like `IMADD` or `IMSUB`.

See the list of Excel functions here:
https://support.microsoft.com/en-au/office/excel-functions-by-category-5f91f4e9-7b42-46d2-9bd1-63f26a86c0eb

also you can test in the playground if it works:
https://playground.ironcalc.com/

fosdickio:
@nhatcher will take over this one.  Closing.

Issue #21, opened 2024-03-14
Document all implemented spreadsheet functions automatically 


nhatcher:
This should happen in the GitHub actions.

Maybe using something like:

https://github.com/marketplace/actions/publish-to-github-wiki

We should test that every function that is implemented is documented. For the MVP a simple list will suffice but in general:
* Each implemented function should have an example webpage
* maybe also link to the xlsx tests it appears on
* Lunks to Excel, G sheets, libre office documentation (and test that links are not broken)

nhatcher:
We have enough for the MVP now

PR #22, opened 2024-03-16
UPDATE: Adds minimal generated documented functions

nhatcher:
See #21


nhatcher:
Remark to self, the code coverage tool is amazing!

Issue #7, opened 2024-03-12
Create UserModel API


nhatcher:
The UserModel API adds undo/redo history and diffs on the top of Model:

This is the API intended to be used by applications like the Web UI or Tironcalc.

In pseudo code:

```
let model = UserModel::new();
model.set_input("A1", "=1+1");
assert(model.get("A1") == 2);
model.undo();
model.redo();
```

In this API changes like setting inputs will also evaluate the model.
This also should emit diffs with changes and be able to apply external diffs. Preparing the client for parallel editing

nhatcher:
Done!

PR #27, opened 2024-03-31
UPDATE: Adds 'user model' API

nhatcher:
* bump version for documentation
* Fixes wrong doc comment
* renames old APIs to be consistent

Hey @fosdickio, 

This is finally an example PR. There is a lot of ground covered in the PR.

This adds a wrapper around Model that will be used by spreadsheet applications (as opposed to engines).

The other API is the raw API. 
 
Setting a value in the 'User' API also evaluates the model and gives you an interface to undo the change if needed. In this way the model is never in a bad state. If you set a value with the raw API and forget to call `model.evaluate()` some values might be wrong.

We can talk about the PR on a call.


nhatcher:
Fixes #7

base/src/actions.rs
```diff
@@ -184,6 +189,64 @@ impl Model {
                 delta: -column_count,
             }),
         );
+        let worksheet = &mut self.workbook.worksheet_mut(sheet)?;
+
+        // deletes all the column styles
```
fosdickio, line 194: @nhatcher Do we still need to worry about garbage collecting styles or does this take care of it when we move cells?

PR #32, opened 2024-04-11
FIX: Fixes types for TypeScript manually

nhatcher:


bindings/wasm/fix_types.py
```diff
@@ -0,0 +1,97 @@
+# Regrettably at the time of writing there is not a perfect way to
+# generate the TypeScript types from Rust so we basically fix them manually
+# Hopefully this will suffice for our needs and one day will be automatic
```
fosdickio, line 3: @nhatcher 😢 Do you by chance know of any projects working on solving this right now?

nhatcher: No, there is [tsify](https://github.com/madonoharu/tsify), but causes more problems than it solves. What we are doing is not too bad and I am pretty sure that the technology will catch up eventually.
It is possible to inject some types with [typescript_type](https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/typescript_type.html), maybe that's an option.

PR #35, opened 2024-04-14
FIX: small diverse fixes

nhatcher:
* Uses `r#true`, `r#false` and `r#ref` instead of `false_value`, `true_value` and `ref_value`
* Uses binary internal representation and remove `json` serializarion/deserialization where not needed
* fixes wasm types issues


fosdickio:
@nhatcher Out of curiosity, what is the advantage of this change?

> Uses r#true, r#false and r#ref instead of false_value, true_value and ref_value

nhatcher:
> @nhatcher Out of curiosity, what is the advantage of this change?
> 
> > Uses r#true, r#false and r#ref instead of false_value, true_value and ref_value

It's just more consistent. When I wrote that code I didn't know you could use this notation to "sidestep" Rust reserved names. Without it I had to do a rename in the other side.

PR #72, opened 2024-05-28
FIX: Export views and showGridLines properly

nhatcher:


xlsx/src/export/worksheets.rs
```diff
@@ -247,15 +248,47 @@ pub(crate) fn get_worksheet_xml(
...
     format!(
         "{XML_DECLARATION}
 <worksheet \
 xmlns=\"http://schemas.openxmlformats.org/spreadsheetml/2006/main\" \
 xmlns:r=\"http://schemas.openxmlformats.org/officeDocument/2006/relationships\">\
   <dimension ref=\"{dimension}\"/>\
   <sheetViews>\
-    <sheetView workbookViewId=\"0\">\
-        <selection activeCell=\"A1\" sqref=\"A1\"/>\
+    <sheetView workbookViewId=\"0\"{show_grid_lines}{tab_selected}>\
```
nhatcher, line 290: Hi @nhatcher, this is great but the selected tab is also on the `xl/workbook.xml` file as:

```xml
<bookViews>
<workbookView xWindow="0" yWindow="500" windowWidth="28800" windowHeight="15800" activeTab="8" xr2:uid="{94D94FA5-4B40-4032-9EB1-15E3426A29B1}"/>
</bookViews>
```
I am not sure what the other options are, are they optional?

nhatcher: Good catch! Fixed, I didn't add all the other parameters. Also should we use this when importing?

PR #76, opened 2024-07-14
Exporting the Merged Cell : Exports the merged cell properly with testcase to verify the scenario

varuntumbe:
What is the enhancement ?

>> Currently the project can only load the merged cells from xlsx file, exporting the merged cells was not supported. This PR addresses this issue.

How it has been done ?

>> Small patch of code has been added to _get_worksheet_xml_ function where we export the merged cells in xml.

What is the UT ?

>> Testcase covers the basic scenario where, we load the already existed xlsx file ( containing the merged cell ) and we export the same file to different name. Later we compare the xml block of each files and validate.

### DIFF
--- a/xlsx/src/export/test/test_export.rs
+++ b/xlsx/src/export/test/test_export.rs
@@ -78,38 +78,6 @@ fn test_values() {
     }
 }
 
-// This test verifies whether exporting the
-// merged cells functionality is happening
-// properly or not.
-
-// This test is bit unconventional from the 
-// other tests as it first loads the excell
-// having the merged cell and exports it to 
-// anothe xlsx and verifies whether merged
-// cell node is same in both of the xlsx file
-#[test]
-fn test_exporting_merged_cells(){
-    let expected_merge_cell_ref;
-    let got_merge_cell_ref;
-    let temp_file_name = "temp_file_test_export_merged_cells.xlsx";
-    {
-        // loading the xlsx file containing merged cells
-        let example_file_name = "tests/example.xlsx";
-        let mut model = load_from_xlsx(example_file_name, "en", "UTC").unwrap();
-        expected_merge_cell_ref = model.workbook.worksheets.get(0).unwrap().merge_cells.clone();
-        // exporting and saving it in another xlsx
-        model.evaluate();
-        save_to_xlsx(&model, temp_file_name).unwrap();
-    }
-    //loading the previous file back and verifying whether
-    //merged cells got exported properly or not
-    let temp_model = load_from_xlsx(temp_file_name, "en", "UTC").unwrap();
-    got_merge_cell_ref = temp_model.workbook.worksheets.get(0).unwrap().merge_cells.clone();
-    assert_eq!(expected_merge_cell_ref, got_merge_cell_ref);
-
-    fs::remove_file(temp_file_name).unwrap();
-}
-
 #[test]
 fn test_formulas() {
     let mut model = new_empty_model();
--- a/xlsx/src/export/test/test_export.rs
+++ b/xlsx/src/export/test/test_export.rs
@@ -78,6 +78,38 @@ fn test_values() {
     }
 }
 
+// This test verifies whether exporting the
+// merged cells functionality is happening
+// properly or not.
+
+// This test is bit unconventional from the 
+// other tests as it first loads the excell
+// having the merged cell and exports it to 
+// anothe xlsx and verifies whether merged
+// cell node is same in both of the xlsx file
+#[test]
+fn test_exporting_merged_cells(){
+    let expected_merge_cell_ref;
+    let got_merge_cell_ref;
+    let temp_file_name = "temp_file_test_export_merged_cells.xlsx";
+    {
+        // loading the xlsx file containing merged cells
+        let example_file_name = "tests/example.xlsx";
+        let mut model = load_from_xlsx(example_file_name, "en", "UTC").unwrap();
+        expected_merge_cell_ref = model.workbook.worksheets.get(0).unwrap().merge_cells.clone();
+        // exporting and saving it in another xlsx
+        model.evaluate();
+        save_to_xlsx(&model, temp_file_name).unwrap();
+    }
+    //loading the previous file back and verifying whether
+    //merged cells got exported properly or not
+    let temp_model = load_from_xlsx(temp_file_name, "en", "UTC").unwrap();
+    got_merge_cell_ref = temp_model.workbook.worksheets.get(0).unwrap().merge_cells.clone();
+    assert_eq!(expected_merge_cell_ref, got_merge_cell_ref);
+
+    fs::remove_file(temp_file_name).unwrap();
+}
+
 #[test]
 fn test_formulas() {
     let mut model = new_empty_model();
--- a/xlsx/src/export/worksheets.rs
+++ b/xlsx/src/export/worksheets.rs
@@ -13,7 +13,7 @@
 //!   <v>1</v>
 //! </c>
 //! Formula in F6 would then be 'A6+C6'
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt::format};
 
 use itertools::Itertools;
 
@@ -59,6 +59,7 @@ pub(crate) fn get_worksheet_xml(
 ) -> String {
     let mut sheet_data_str: Vec<String> = vec![];
     let mut cols_str: Vec<String> = vec![];
+    let mut merged_cells_str: Vec<String> = vec![];
 
     for col in &worksheet.cols {
         // <col min="4" max="4" width="12" customWidth="1"/>
@@ -241,6 +242,15 @@ pub(crate) fn get_worksheet_xml(
         ))
     }
     let sheet_data = sheet_data_str.join("");
+
+    for (_ind,merge_cell_ref) in worksheet.merge_cells.iter().enumerate(){
+        merged_cells_str.push(format!(
+            "<mergeCell ref=\"{merge_cell_ref}\"/>"
+        ))
+    }
+    let merged_cells_count = merged_cells_str.len();
+    let merged_cells = merged_cells_str.join("");
+
     let cols = cols_str.join("");
     let cols = if cols.is_empty() {
         "".to_string()
@@ -295,6 +305,9 @@ xmlns:r=\"http://schemas.openxmlformats.org/officeDocument/2006/relationships\">
   <sheetData>\
   {sheet_data}\
   </sheetData>\
+  <mergeCells count=\"{merged_cells_count}\">\
+  {merged_cells}\
+  </mergeCells>\
 </worksheet>"
     )
 }
--- a/xlsx/src/export/worksheets.rs
+++ b/xlsx/src/export/worksheets.rs
@@ -243,11 +243,10 @@ pub(crate) fn get_worksheet_xml(
     }
     let sheet_data = sheet_data_str.join("");
 
-    for (_ind, merge_cell_ref) in worksheet.merge_cells.iter().enumerate() {
+    for merge_cell_ref in worksheet.merge_cells.clone() {
         merged_cells_str.push(format!("<mergeCell ref=\"{merge_cell_ref}\"/>"))
     }
     let merged_cells_count = merged_cells_str.len();
-    let merged_cells = merged_cells_str.join("");
 
     let cols = cols_str.join("");
     let cols = if cols.is_empty() {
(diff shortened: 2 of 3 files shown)


xlsx/src/export/worksheets.rs
```diff
@@ -295,6 +305,9 @@ xmlns:r=\"http://schemas.openxmlformats.org/officeDocument/2006/relationships\">
   <sheetData>\
   {sheet_data}\
   </sheetData>\
+  <mergeCells count=\"{merged_cells_count}\">\
```
nhatcher: if `merged_cell_count` is `0` I think we should not add this at all.

xlsx/src/export/test/test_export.rs
```diff
@@ -78,6 +78,38 @@ fn test_values() {
     }
 }
 
+// This test verifies whether exporting the
+// merged cells functionality is happening
+// properly or not.
+
+// This test is bit unconventional from the
```
nhatcher: Isn't this test quite similar to

https://github.com/ironcalc/IronCalc/blob/main/xlsx/tests/test.rs#L52-L95

?

nhatcher: Also I think this is more of an _integration_ test than a unit test, that's why it is probably better suited in that folder.

nhatcher (changes requested):
Brilliant, I added a couple of minor comments.

xlsx/src/export/test/test_export.rs
```diff
@@ -78,6 +78,38 @@ fn test_values() {
...
+
+// This test is bit unconventional from the 
+// other tests as it first loads the excell
+// having the merged cell and exports it to 
+// anothe xlsx and verifies whether merged
+// cell node is same in both of the xlsx file
+#[test]
+fn test_exporting_merged_cells(){
+    let expected_merge_cell_ref;
+    let got_merge_cell_ref;
```
nhatcher: You don't need to declare  `got_merge_cell_ref` here. You can just do this at the point where it is defined. You do need to declare `expected_merge_cell_ref` because it is defined in a place with a smaller scope. One handsome thing you can do is:

```rust
    let expected_merged_cell_ref = {
        // loading the xlsx file containing merged cells
        let example_file_name = "tests/example.xlsx";
        let mut model = load_from_xlsx(example_file_name, "en", "UTC").unwrap();
        let merged_cell_ref = model.workbook.worksheets.get(0).unwrap().merge_cells.clone();
        // exporting and saving it in another xlsx
        model.evaluate();
        save_to_xlsx(&model, temp_file_name).unwrap();
        merged_cell_ref
    }
```

But that's not needed (I'm not even sure I like it!)

xlsx/src/export/worksheets.rs
varuntumbe: Refactored the code to where only when merged_cell_count greater than 0, xml block gets added

xlsx/src/export/test/test_export.rs
varuntumbe: Got it, moved this testcase to test/ folder + added a case to which checks the 0 merged cell count scenarios

xlsx/src/export/test/test_export.rs
varuntumbe: Understood. This does look clean.

xlsx/src/export/worksheets.rs
```diff
@@ -280,6 +288,16 @@ pub(crate) fn get_worksheet_xml(
...
     }
 
+    let merge_cells_section = if merged_cells_count > 0 {
+        format!(
+            "<mergeCells count=\"{merged_cells_count}\">{merged_cells}</mergeCells>",
+            merged_cells_count = merged_cells_count,
+            merged_cells = merged_cells
+        )
+    } else {
+        String::new()
```
nhatcher: This is correct but `"".to_string()` is what we use elsewhere.

xlsx/tests/test.rs
```diff
@@ -334,3 +334,72 @@ fn no_export() {
...
+        // exporting and saving it in another xlsx
+        model.evaluate();
+        save_to_xlsx(&model, temp_file_name).unwrap();
+        expected_merge_cell_ref
+    };
+    {
+        let mut temp_model = load_from_xlsx(temp_file_name, "en", "UTC").unwrap();
+        {
+            //loading the previous file back and verifying whether
+            //merged cells got exported properly or not
```
nhatcher: Please add a space between `//` and `merged ..`. Same in the comment above.

xlsx/tests/test.rs
```diff
@@ -334,3 +334,72 @@ fn no_export() {
     }
     fs::remove_dir_all(&dir).unwrap();
 }
+
+// This test verifies whether exporting the
+// merged cells functionality is happening
+// properly or not.
+// it first loads the excell
```
nhatcher: Capital `It`. Also you are free to use more space horizontally.

xlsx/src/export/worksheets.rs
```diff
@@ -280,6 +288,16 @@ pub(crate) fn get_worksheet_xml(
         }
     }
 
+    let merge_cells_section = if merged_cells_count > 0 {
+        format!(
+            "<mergeCells count=\"{merged_cells_count}\">{merged_cells}</mergeCells>",
+            merged_cells_count = merged_cells_count,
+            merged_cells = merged_cells
```
nhatcher: This is great but it doesn't match the style I have been doing everywhere:
```rust
    let merge_cells_section = if merged_cells_count > 0 {
        format!(
            "<mergeCells count=\"{}\">{}</mergeCells>",
            merged_cells_count,  merged_cells_str.join("")  )
    } else {
        "".to_string()
    }
```
You don't  need to define `merged_cells` at all. Also I think we should escape all strings.
See producing xlm by hand is dangerous. All strings that come from a user should be escaped `escape_xml(...)`

xlsx/src/export/worksheets.rs
```diff
@@ -241,6 +242,13 @@ pub(crate) fn get_worksheet_xml(
         ))
     }
     let sheet_data = sheet_data_str.join("");
+
+    for (_ind, merge_cell_ref) in worksheet.merge_cells.iter().enumerate() {
```
nhatcher: Can't you just do  `for merge_cell_ref in worksheet.merge_cells {`?

nhatcher (changes requested):
Almost there!

xlsx/src/export/worksheets.rs
varuntumbe: Done

xlsx/tests/test.rs
varuntumbe: okay noted and done

xlsx/tests/test.rs
varuntumbe: Done

xlsx/src/export/worksheets.rs
varuntumbe: Understood.
removed merged_cells
string escaping added.

Noted the escape_xml() routine.

xlsx/src/export/worksheets.rs
varuntumbe: Cannot directly use` for merge_cell_ref in worksheet.merge_cells {` as merge_cells is Vec of String,  String gets moved. so doing` for merge_cell_ref in worksheet.merge_cells.clone() {` instead

xlsx/tests/test.rs
```diff
@@ -334,3 +334,68 @@ fn no_export() {
...
+                .merge_cells
+                .clone();
+            assert_eq!(expected_merge_cell_ref, *got_merge_cell_ref);
+            fs::remove_file(temp_file_name).unwrap();
+        }
+        {
+            // this block is to verify that if there are no
+            // merged cells, exported xml should not have the
+            // <mergeCells/> xml node
+            temp_model
```
nhatcher, line 379: This is fine, but I think we probably should use an (not yet existing) API  call for this

xlsx/src/export/worksheets.rs
nhatcher: Are the string escaped properly?

xlsx/src/export/worksheets.rs
nhatcher: Maybe you can use: `for merge_cell_ref in &worksheet.merge_cells{`? The idea is not to allocate if we don't have to.

nhatcher (commented):
Very last round!

nhatcher:
(ignore the coverage failure)

xlsx/src/export/worksheets.rs
varuntumbe: Oops yes, noted and changed

xlsx/tests/test.rs
varuntumbe: ohh okay. Let me know if its necessary, I can  implement get and clear functions for merged cells in worksheet

xlsx/src/export/worksheets.rs
varuntumbe: `        format!(
            "<mergeCells count=\"{}\">{}</mergeCells>",
            merged_cells_count,
            merged_cells_str.join("")
        )`

This has escaped string right ? '"' is escaped with ' \\' and other than this, I dont see any strings which needs escaping.

xlsx/tests/test.rs
varuntumbe: I have added a commit which has APIs for get and clear on merge cells. if not needed, will revert the last commit.

xlsx/src/export/worksheets.rs
nhatcher: No, `merged_cells_str.join("")` is itself a string and it is not properly escaped. The string could be something like (it's called xml injection):

```
</mergeCells><some_tag_that_does_something_nasty/>
```

I caused the problem in the first first when I decided to build the xml by hand. That is almost always a bad idea precisely because of this issue. I am not a friend of dependencies and while we do use an external parser to _load_ the xml we build the xml by hand. This does mean we have to be extra careful when building the xml. All strings must be correctly, you can use:

https://github.com/ironcalc/IronCalc/blob/main/xlsx/src/export/escape.rs#L81

for this mater.

xlsx/tests/test.rs
nhatcher: Hi @varuntumbe, no no need to do this just yet. I think it is fine if we use the json structure for the purposes of this PR.
Let's deal with the correct API for merge cell on a separate commit.  Thanks!

nhatcher (changes requested):
Hi @varuntumbe, we are almost there. But yes, please remove the API calls, we will add them shortly in your next PR!

One more question, have you tested that the generated files open correctly with Excel?

varuntumbe:
> Hi @varuntumbe, we are almost there. But yes, please remove the API calls, we will add them shortly in your next PR!
> 
> One more question, have you tested that the generated files open correctly with Excel?


Okay @nhatcher , removed the last 2 commits ( containing the API calls ). 

Yes I have opened the file and excel and verified.  Screen shot is below.

image

xlsx/tests/test.rs
varuntumbe: Got it Thanks @nhatcher

xlsx/src/export/worksheets.rs
varuntumbe: Apologies for dragging this further,

I tried using the above function ( escape_xml ) on 

`merged_cells_str.join("")`

escaped string o/p : ` "<mergeCells count=\"2\">&lt;mergeCell &quot;ref&quot;=&quot;K7:L10&quot;/&gt;&lt;mergeCell &quot;ref&quot;=&quot;H18:J20&quot;/&gt;</mergeCells>"` 

But getting below error in testcase, when I try to load the same xlsx file again

``Result::unwrap()` on an `Err` value: Xml("Missing \"ref\" XML attribute")`

Plus when I open the exported xlsx file, merged cells are not coming in excel.

I understood purpose of using escaping of xml ( thanks to you ! ), Will this be necessary here ? Because these strings wont be coming from user right ? These strings are made by application and exporting it. Plus the exported XML should have not have these symbols ( &lt, &qut etc ) right ?

Let me know how I can resolve this.

Thanks,
Varun

nhatcher: Oh yeah, that wouldn't work :). The `escape_xml` escapes the content inside xml, here we would want to escape an attribute. Now the attribute you are reading from an existing string that has already been imported from Excel. This string should be a range like: `A1:C3`. I have some thouhts about this:

* we shouldn't store strings in the internal structure, maybe we can store already the parsed range?
* this also has implications when importing. So we should import and parse the merge cells ranges

I think to move forward we can merge this as it is (no pun intended) since it is consistent with the importing mechanism.
In an upcoming PR we should fix this though. Either marking cells as part of a merged cells structure or having the merged cells.

By the way your mechanism will work, you just need to escape each of the strings in the refs:

```rust
    for merge_cell_ref in &worksheet.merge_cells {
        merged_cells_str.push(format!("<mergeCell ref=\"{}\"/>", escape_xml(merge_cell_ref)))
    }
```

But let's not do that this time around, we will eventually build that string ourselves from numbers, so there will be no escaping needed.
Many thanks!

nhatcher (approved):
LGTM!

xlsx/src/export/worksheets.rs
varuntumbe: Understood, I will resolve the conversation here. In the next PR, Lets add this work item as well .

PR #78, opened 2024-08-07
Merged Cells Implementation

varuntumbe:
### What are we trying to achieve with this PR ?

++ Currently IronCalc just imports and exports the merged cell without imposing rules of merged cells ( cell belonging to merge cells block cannot be modified etc ). This PR adds those conditions and restrictions which completes Merge cell implementation on IronCalc

### What changes are added in laymen terms ?

- New Merge cell struct has been added to cells.rs
- Check has been added to all public set fns where we check whether cell is being part of merged cells or not.
- Update and Delete merge cells ( Public API for handling merge cells ) have been added
- Exporting and importing of worksheet have been changed a bit to accommodate the above changes
- Testcase has been added
 

### What happens when user tries to create new merge cell which overlaps with other Merge cells ?

**Algorithm to Handle Overlapping Merges**

1.	Identify the Cells to Merge
Determine the range of cells you want to merge. For example, merge cells A1 and C1.
2.	Detect Existing Merged Areas
Check if any of the cells involved are part of existing merged cell ranges.
3.	Unmerge Overlapping Areas
Unmerge the areas where overlaps occur to avoid conflicts.
4.	Merge the New Range
Perform the merge operation on the new range.

### Manual testing and verification

case 1 : Merged cell in excell

Screenshot 2024-10-09 135425

case 2 : Trying to create a overlapped merge cell ( C1 to G4 )

Screenshot 2024-10-09 160344

++> In excell App Note that its not possible to select exactly C1 to G4, It will end up selecting all possible overlapped Merge cell

Screenshot 2024-10-09 160430

++> Finally when you create new one is formed

**Same scenarios created in IronCalc**

case 1 : Merged cell

Screenshot 2024-10-03 180501

case 2 : Trying to create overlapped merge cell ( C1 to G4 )

Screenshot 2024-10-03 180544

++> Here its possible as we are creating using programming interface not user interface. But the logic remains same, where new merge cell to be created if overlaps with any of the existing merge cell, those gets unmerged and new one gets created


xlsx/src/import/worksheets.rs
```diff
@@ -146,20 +146,23 @@ fn load_columns(ws: Node) -> Result<Vec<Col>, XlsxError> {
...
+    let mut merge_cells: Vec<MergeCell> = Vec::new();
     let merge_cells_nodes = ws
         .children()
         .filter(|n| n.has_tag_name("mergeCells"))
         .collect::<Vec<Node>>();
     if merge_cells_nodes.len() == 1 {
         for merge_cell in merge_cells_nodes[0].children() {
             let reference = get_attribute(&merge_cell, "ref")?.to_string();
-            merge_cells.push(reference);
+            let merge_cell_st = reference.split(":").nth(0).unwrap();
```
nhatcher: We should never use unwrap in production code. And reference needs proper parsing. There is a function `parse_range` in this file you can use

xlsx/tests/test.rs
```diff
@@ -399,3 +399,37 @@ fn test_exporting_merged_cells() {
...
+    let mut model = load_from_xlsx(example_file_name, "en", "UTC").unwrap();
+
+    model.set_user_input(0, 1, 5, "Hello".to_string());
+
+    // making sure that merge cell should not get updated
+    assert_ne!(model.get_cell_content(0, 1, 5).unwrap(), "Hello");
+    assert_ne!(model.get_cell_type(0, 1, 5).unwrap(), CellType::Text);
+
+    model.update_cell_with_bool(0, 1, 5, true);
+    assert_ne!(model.get_cell_content(0, 1, 5).unwrap(), "true");
```
nhatcher: Ha! I didn't know about the `assert_ne` macro! TIL.
But wouldn't it be easier to check for equality?

base/src/types.rs
```diff
@@ -347,6 +347,18 @@ pub enum FontScheme {
     None,
 }
 
+// Merge Cell type
+// There will be one MergeCell maintained for every Merge cells that we load
+// It will consist of range_ref : its range reference in String representation ( ex : A1-C2 )
+// merge_col_range : Its tuple having start and end values of columns ( ex: A1-C2 => (1,3))
+// merge row_range : Its tuple having start and end values of rows (ex: A1-C2 => (1,2))
+#[derive(Encode, Decode, Debug, PartialEq, Clone)]
+pub struct MergeCell {
```
nhatcher: On IronCalc both rows and columns are `i32`. Now it might not be the best option, and we might one day change the _whole_ code base to a different choice, but for now it is better if we stick to it (and you will not need to do `as u32` all over.

Also, in other places we use `range = [row_start, column_start, row_end, column_end]`. I suggest we use a similar tstructure here.

base/src/utils.rs
```diff
@@ -367,3 +367,40 @@ mod tests {
         assert!(!is_valid_hex_color("#ffffff00")); // with alpha channel
     }
 }
+
+pub(super) fn get_column_number(s: &str, str_len: usize) -> u32 {
```
nhatcher: These functions are already implemented elsewhere. Search for instance `number_to_column`

base/src/model.rs
```diff
@@ -1219,6 +1226,17 @@ impl Model {
     /// * [Model::update_cell_with_bool()]
     /// * [Model::update_cell_with_formula()]
     pub fn update_cell_with_text(&mut self, sheet: u32, row: i32, column: i32, value: &str) {
+        // Checking first whether cell we are updating is part of Merged cells
+        // If so, returning with print message
+        // TODO : Any better error handling ? ==> Should we start returning Result<> op ?
```
nhatcher: Correct, error handling is missing here. Maybe a first PR should tackle that? I'm not sure. What is the `sheet` doesn't exist? A public method should never panic.

nhatcher (changes requested):
Hi @varuntumbe, I know this is WIP but just came to leave some thoughts.

I think the general approach is correct. We might have some discussions about the correct struct for merged cells and some naming arguments :D.

As you detected proper error handling is missing in many methods. It might be a good idea to add those first in a separate PR.

Also we need to be able to merge and unmerge cells, that should be part of the public API.

Excellent first step! :rocket:

varuntumbe:
> Hi @varuntumbe, I know this is WIP but just came to leave some thoughts.
> 
> I think the general approach is correct. We might have some discussions about the correct struct for merged cells and some naming arguments :D.
> 
> As you detected proper error handling is missing in many methods. It might be a good idea to add those first in a separate PR.
> 
> Also we need to be able to merge and unmerge cells, that should be part of the public API.
> 
> Excellent first step! :rocket: 

@nhatcher thanks for the advice. Will add error handling and other things to the PR and make it ready to review soon.

Thanks,
Varun

xlsx/src/import/worksheets.rs
varuntumbe: Understood, will be using this

xlsx/tests/test.rs
varuntumbe: Yes, I can check for equality as well. Introduced negation here as its more intuitive, we just want to test that update_cell_with_bool does not set the cell as true as its a part of merge cell.

base/src/types.rs
varuntumbe: Understood keeping the above struct

base/src/utils.rs
varuntumbe: These are cleaned up, as parse_range itself will take care of these things

base/src/model.rs
varuntumbe: Here one doubt.
I see only **update_cell_with_formula** returns Result<(), String> where I can propogate the error back.

All the other updates does not return anything ( kept like that purposefully as its set functions ). If I change these in all of these public apis, I need to add error handling in all of the testcases as it starts to give warning ( saying we did not handle the error return values ).

Should I propagate the error back in these functions and modify the testcases ?

OR

Should I keep it as it is but just silently log it saying update did not go through ?

xlsx/tests/test.rs
varuntumbe: Added equality assertion instead of negation

base/src/model.rs
varuntumbe: Added basic error handling

base/src/model.rs
```diff
@@ -1218,7 +1225,29 @@ impl Model {
...
     /// * [Model::update_cell_with_bool()]
     /// * [Model::update_cell_with_formula()]
-    pub fn update_cell_with_text(&mut self, sheet: u32, row: i32, column: i32, value: &str) {
+    pub fn update_cell_with_text(
+        &mut self,
+        sheet: u32,
+        row: i32,
+        column: i32,
+        value: &str,
+    ) -> Result<(), String> {
```
nhatcher, line 1246: Can we do the error propagation in a different PR?

That will be a fantastic learning experience by the way. Maybe you can start with one method.

You will have to add tests:

* what if the sheet is incorrect 
* what if the row, column are out of range
* is the text always valid

Let em know if you think you can do that (this is technical deb)

varuntumbe: Okay you mean error handling of all set functions ( leaving the merge cell part ) should be handled separately in a new PR ?

Okay I remove my last commit from here, and handle the error handling in a separate PR

base/src/model.rs
```diff
@@ -1221,6 +1229,21 @@ impl Model {
...
+                    return Err(format!(
+                        "Cell row : {}, col : {} is part of merged cell block, so singular update to the cell is not possible",
+                        row, column
+                    ));
+                }
+            }
+            Err(err) => {
+                return Err(err);
+            }
+        }
```
nhatcher, line 1254: You might simplify this a bit:

```Rust
 if self.is_part_of_merge_cell(sheet, row, column)? {
                    return Err(format!(
                        "Cell row : {}, col : {} is part of merged cell block, so singular update to the cell is not possible",
                        row, column
                    ));
                }
        }
```

base/src/model.rs
```diff
@@ -1271,6 +1294,22 @@ impl Model {
...
```
nhatcher, line 1312: Same in here

base/src/types.rs
```diff
@@ -351,6 +351,16 @@ pub enum FontScheme {
...
+// Merge Cell type
+// There will be one MergeCell maintained for every Merge cells that we load
+// It will consist of range_ref : its range reference in String representation
+// merge_cell_range : Its tuple having [row_start, column_start, row_end, column_end]
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub struct MergeCell {
+    pub merge_cell_range: (i32, i32, i32, i32),
+    pub range_ref: String,
+}
+
```
nhatcher, line 391: I would suggest some slightly different naming:

```rust
pub struct MergedRange {
  range: (i32, i32, i32, i32),
  r#ref: String
}
```

It is not a cell in it a range of cells that is merged. Also the thing is a _merged_ range not a _merge_ range but we _merge_ a range of cells into a _merged_ range.

Second thing. Do we need the 'ref' at all? It is redundant, opening the door to bugs where the range and the ref fall out of sync. We could use it if we would need it for perfomance reasons. another reason to keep the string representation is if it differs of what can be computed. For instance if the user enters `a1:a3` but we would compute `A1:A3`. This is not the case in here.

I would just try:
```rust
struct MergedRange = (i32, i32, i32, i32);
```

(See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9fecb809b857b6643b044c4187db1274)

and provide some utility function to compute the str representation.

nhatcher: Note that `ref` is a keyword in Rust. You can use keywords in Rust by escaping them with `r#` in front of them.

base/src/worksheet.rs
```diff
@@ -522,6 +522,23 @@ impl Worksheet {
...
 
+    /// Returns true if cell part of Merged Cells.
+    /// First cell of Merge block is not considered as part of Merge cell
+    /// Ex : if Merge cells were A1-C3, A1 is not considered as part of Merge cell
+    pub fn is_part_of_merge_cell(&self, row: i32, column: i32) -> Result<bool, String> {
+        if !is_valid_column_number(column) || !is_valid_row(row) {
+            return Err("Incorrect row or column".to_string());
+        }
+
+        //traverse through Vector of Merged Cells and return (linear search)
```
nhatcher: Please always leave a space between the `//` and the rest of the comment. I will try to have a linter that imposes that rule.

base/src/types.rs
```diff
@@ -110,7 +110,7 @@ pub struct Worksheet {
     pub sheet_id: u32,
     pub state: SheetState,
     pub color: Option<String>,
-    pub merge_cells: Vec<String>,
+    pub merge_cells: Vec<MergeCell>,
```
nhatcher: I think the correct name would be:

```rust
pub merged_ranges: Vec<MergedRange>
```

nhatcher (changes requested):
Hi @varuntumbe , thanks you very much for you incredible patience.
Here is finally my review round.
I _think_ we need to change the name of the object, and probably simplify the structure into a tuple.

Let me know what you think.

base/src/types.rs
```diff
@@ -351,6 +352,64 @@ pub enum FontScheme {
...
+    }
+
+    // Method which returns range_ref from the tuple
+    // ex : (3,1,4,2) is interpreted as A3:B4
+    // This public method does not return the output wrapped in Result
+    // idea is, once MergedRange struct gets created, there shoud not be any problem in converting
+    // the tuple in to the range_ref string. so purposefully number_to_column fn is not handled
+    // if it panics here, then it will be a bug as in corner scenario which we had not thought while creating Merge cell
+    pub fn get_merge_range_as_str(&self) -> String {
+        let st_col = number_to_column(self.get_col_start()).expect(&format!(
```
nhatcher: We can't use `expect`, that is similar to `unwrap` or `panic`. we can't panic in the library

base/src/types.rs
```diff
@@ -351,6 +352,64 @@ pub enum FontScheme {
...
 }
 
+// MergedRange type
+// There will be one MergedRange struct maintained for every Merge cells that we load
+// merge_cell_range : Its tuple having [row_start, column_start, row_end, column_end]
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub struct MergedRange(i32, i32, i32, i32);
+
+// implementing accessor function
+impl MergedRange {
```
nhatcher: I see where you are coming from, and you are not wrong, but IronCalc doesn't do this. We have a solid convention:

`(row_start, column_start, row_end, column_end)`

and we access members directly. What you are doing is great, and has many advantages, but for now I would like to have a more uniform code style.

There are many advantages of what you are doing. For instance all those are valid choices:

```
(row_start, column_start, row_end, column_end)
(row_start, row_end, column_start, column_end)
(row_start, column_start, row_width, column_width)
...
```
and many variations. And we might end up refactoring things in that direction one day

base/src/worksheet.rs
```diff
@@ -522,6 +522,23 @@ impl Worksheet {
         Ok(is_empty)
     }
 
+    /// Returns true if cell part of Merged Cells.
+    /// First cell of Merge block is not considered as part of Merge cell
+    /// Ex : if Merge cells were A1-C3, A1 is not considered as part of Merge cell
+    pub fn is_part_of_merge_cell(&self, row: i32, column: i32) -> Result<bool, String> {
```
nhatcher: Remember there is not "merge cell" there is a "merged range" or "merged cells".

base/src/worksheet.rs
```diff
@@ -522,6 +522,23 @@ impl Worksheet {
...
+        if !is_valid_column_number(column) || !is_valid_row(row) {
+            return Err("Incorrect row or column".to_string());
+        }
+
+        // traverse through Vector of Merged Cells and return (linear search)
+        for merge_cell in &self.merge_cells {
```
nhatcher: `for merged_rage in &self.merged_ranges`

base/src/types.rs
nhatcher: Also, please don't use `col`, use `column` (I know brain dead, but I use it all over)

xlsx/src/import/worksheets.rs
```diff
@@ -146,20 +146,28 @@ fn load_columns(ws: Node) -> Result<Vec<Col>, XlsxError> {
     Ok(cols)
 }
 
-fn load_merge_cells(ws: Node) -> Result<Vec<String>, XlsxError> {
+fn load_merge_cells_nodes(ws: Node) -> Result<Vec<MergedRange>, XlsxError> {
```
nhatcher: merged_cells (or merged_range or merged_ranges). This was my mistake initially

base/src/types.rs
```diff
@@ -110,7 +111,7 @@ pub struct Worksheet {
     pub sheet_id: u32,
     pub state: SheetState,
     pub color: Option<String>,
-    pub merge_cells: Vec<String>,
+    pub merge_cells: Vec<MergedRange>,
```
nhatcher: I think I prefer `merged_range` and `merged_ranges` otherwise you would have have to go for `merged_cells` and `merged_cells_list` or something.

I am happy o discuss here, I'm not a native English speaker!

I see the point in calling a "merged cell" a cell that is it composed of many cells.

I think if we go with merged ranges we sidestep the problem completely. In that case it would be:

```rust
pub merged_ranges: Vec<MergedRange>
```

what do you think?

nhatcher (changes requested):
Hi @varuntumbe ,

I still have some naming issues, maybe we can discuss those on a call to make sure we are aligned?

Let me know what you think

base/src/model.rs
```diff
@@ -1982,6 +2046,164 @@ impl Model {
...
+    /// * [Model::update_cell_with_formula()]
+    /// * [Model::update_cell_with_number()]
+    /// * [Model::update_cell_with_bool()]
+    /// * [Model::update_cell_with_text()]
+    pub fn merge_cells(&mut self, sheet: u32, range_ref: &str) -> Result<(), String> {
+        match self.parse_merged_range(range_ref) {
+            Ok(parsed_merge_cell_range) => {
+                // ATTENTION 2: Below thing we can support here but keeping it simple
+                // Web or different client needs to keep this in mind
+                // User can give errored parse ranges like C3:A1
```
jaycarlton, line 2097: Can the error handling for the range specifications not happen in a central place somewhere? It seems odd to have to handle it here.

base/src/model.rs
```diff
@@ -1982,6 +2046,164 @@ impl Model {
...
+                if parsed_merge_cell_range.0 > parsed_merge_cell_range.2
+                    || parsed_merge_cell_range.1 > parsed_merge_cell_range.3
+                {
+                    return Err(
+                        "Invalid parse range. Merge Mother cell always be top left cell"
+                            .to_string(),
+                    );
+                }
+
+                let mut merged_cells_overlaped_list: Vec<bool> = Vec::new();
```
jaycarlton, line 2109: Why are you using a vector of bools here?

base/src/model.rs
```diff
@@ -1982,6 +2046,164 @@ impl Model {
...
+    /// ```
+    ///
+    /// See also:
+    /// * [Model::update_cell_with_formula()]
+    /// * [Model::update_cell_with_number()]
+    /// * [Model::update_cell_with_bool()]
+    /// * [Model::update_cell_with_text()]
+    pub fn merge_cells(&mut self, sheet: u32, range_ref: &str) -> Result<(), String> {
+        match self.parse_merged_range(range_ref) {
+            Ok(parsed_merge_cell_range) => {
```
jaycarlton, line 2094: Why are you using a tuple here instead of a type with names?

base/src/model.rs
```diff
@@ -1982,6 +2046,164 @@ impl Model {
...
+                            self.workbook.worksheet_mut(sheet)?.update_cell(
+                                row_index,
+                                col_index,
+                                Cell::EmptyCell { s: 0 },
+                            )?;
+                        }
+                    }
+                }
+
+                let new_merged_cells = MergedCells::new(parsed_merge_cell_range);
```
jaycarlton, line 2161: It might be possible to have a single type that represents a cell or range of cells or a merged cell, following the Composite Pattern https://en.wikipedia.org/wiki/Composite_pattern. Not sure if that will work in Rust though.

base/src/model.rs
```diff
@@ -1982,6 +2046,164 @@ impl Model {
...
+    /// # use ironcalc_base::Model;
+    /// # use ironcalc_base::cell::CellValue;
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let mut model = Model::new_empty("model", "en", "UTC")?;
+    /// model.merge_cells(0, "D4:F6");
+    /// model.unmerge_cells(0, "D4:F6");
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn unmerge_cells(&mut self, sheet: u32, range_ref: &str) -> Result<(), String> {
```
jaycarlton, line 2190: Is it not possible to unmerge a single merged cell instead of all of them in the sheet? I'd expect this function to be broken up into just the indexing and a call to another funciton to unmerge each merged cell.

base/src/new_empty.rs
```diff
@@ -57,7 +57,7 @@ impl Model {
             rows: vec![],
             comments: vec![],
             dimension: "A1".to_string(),
-            merge_cells: vec![],
+            merged_cells_list: vec![],
```
jaycarlton, line 60: None of the other things have `_list` in their name. It seems redundant in a safely typed language.

base/src/test/test_model_merge_cell_fns.rs
```diff
@@ -0,0 +1,155 @@
...
+
+#[test]
+fn test_model_set_fns_related_to_merge_cells() {
+    let mut model = new_empty_model();
+
+    //creating a merge cell of D1:F2
+    model.merge_cells(0, "D1:F2").unwrap();
+
+    //Updating the mother cell of Merge cells and expecting the update to go through
+    model.set_user_input(0, 1, 4, "Hello".to_string()).unwrap();
```
jaycarlton, line 13: It's kind of awkward to have to remember which cell is the Mother. Should editing any cell in the merged cell maybe just overwrite the mother cell? Likewise I'd expect retrieving any of the subcells' values or styles to return that of the mother cell.

base/src/test/test_model_merge_cell_fns.rs
```diff
@@ -0,0 +1,155 @@
...
+    // Updating cell which is not in Merge cell block
+    assert_eq!(model.set_user_input(0, 1, 3, "Hello".to_string()), Ok(()));
+    assert_eq!(model.get_cell_content(0, 1, 3), Ok("Hello".to_string()));
+    assert_eq!(model.get_cell_type(0, 1, 3), Ok(CellType::Text));
+
+    // 1: testing with set_user_input()
+    assert_eq!(
+        model
+            .set_user_input(0, 1, 5, "Hello".to_string()),
+        Err("Cell row : 1, col : 5 is part of merged cells block, so singular update to the cell is not possible".to_string())
```
jaycarlton, line 26: Make a string constant to use in all these cases where it's the same.

base/src/test/test_model_merge_cell_fns.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+#[test]
+fn test_model_merge_cells_crud_api() {
+    let mut model = new_empty_model();
+
+    //creating a merge cell of D4:F6
+    model.merge_cells(0, "D4:F6").unwrap();
+    model
+        .set_user_input(0, 4, 4, "Merge Block".to_string())
```
jaycarlton, line 74: Do these setters and getters also support spreadsheet-style indexing (like "D4")? If so, it would be more readable to refer to the cell the same way in all the calls.

base/src/test/test_worksheet.rs
```diff
@@ -283,3 +283,87 @@ fn test_worksheet_navigate_to_edge_in_direction() {
...
+    model.merge_cells(0, "D1:E3").unwrap();
+
+    // Lets check whether D1 (Mother Merge cell) is part of Merge block or not
+    // It should not be considered as part of Merge cell
+    assert!(!model
+        .workbook
+        .worksheet(0)
+        .unwrap()
+        .is_part_of_merged_cells(1, 4)
+        .unwrap(),);
```
jaycarlton, line 303: Do you need the trailing comma here?

base/src/types.rs
```diff
@@ -351,6 +352,43 @@ pub enum FontScheme {
     None,
 }
 
+// MergedCells type
+// There will be one MergedCells struct maintained for every Merged cells that we load
+// merge_cell_range : Its tuple having [row_start, column_start, row_end, column_end]
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub struct MergedCells(pub i32, pub i32, pub i32, pub i32);
```
jaycarlton, line 359: I'd name this `MergedCell` since it's more of a single entity than a collection. Or maybe `MergedCellGroup`?

jaycarlton, line 359: Can we not have named fields like `top_row, top_col, bottom_row, bottom_col`? Or even just two cell structs? I find the literal numbers difficult to read and likely error prone.

base/src/worksheet.rs
```diff
@@ -522,6 +522,23 @@ impl Worksheet {
         Ok(is_empty)
     }
 
+    /// Returns true if cell part of Merged Cells.
+    /// First cell of Merged cells block is not considered as part of Merged cells
+    /// Ex : if Merged cells were A1-C3, A1 is not considered as part of Merged cells block
+    pub fn is_part_of_merged_cells(&self, row: i32, column: i32) -> Result<bool, String> {
+        if !is_valid_column_number(column) || !is_valid_row(row) {
```
jaycarlton, line 529: This check seems very common; maybe factor out a method like `ValidateCellIndices(row, col)`.

base/src/worksheet.rs
```diff
@@ -522,6 +522,23 @@ impl Worksheet {
...
+            return Err("Incorrect row or column".to_string());
+        }
+
+        // traverse through Vector of Merged Cells and return (linear search)
```
jaycarlton, line 533: It wouldn't be that much more work to put these in a hash map for quicker lookup. Unless we only expect very few merges.

base/src/worksheet.rs
```diff
@@ -575,6 +592,16 @@ impl Worksheet {
             }
         }
     }
+
+    /// Returns mutable reference to Vector of Merged cells list
+    pub fn get_merged_cells_list_mut(&mut self) -> &mut Vec<MergedCells> {
```
jaycarlton, line 597: `get_merged_ranges()`

xlsx/tests/test.rs
```diff
@@ -403,7 +403,7 @@ fn test_exporting_merged_cells() {
             .worksheets
             .first()
             .unwrap()
-            .merge_cells
+            .merged_cells_list
```
jaycarlton, line 406: `merged_ranges`

jaycarlton (commented):
Some suggestions. I mainly read this for my own edification.

base/src/test/test_model_merge_cell_fns.rs
nhatcher: This API is very low level on purpose. What you are proposing is not a bad idea, and perhaps we should implement it one day.

base/src/test/test_model_merge_cell_fns.rs
nhatcher: It's also ok in tests to repeat code and keep the different bits independent.

base/src/test/test_model_merge_cell_fns.rs
```diff
@@ -0,0 +1,155 @@
+#![allow(clippy::unwrap_used)]
+
+use crate::{test::util::new_empty_model, types::CellType};
+
+#[test]
+fn test_model_set_fns_related_to_merge_cells() {
+    let mut model = new_empty_model();
+
+    //creating a merge cell of D1:F2
```
nhatcher: I'm sorry for doing this, but could you leave a space after the `// `? (OCD)

AndrooFrowns: Might be good to use this (or a similar github action) to require rust fmt standards (or whatever standards you prefer)

https://github.com/marketplace/actions/rust-rustfmt-check

varuntumbe: No problem, corrected everywhere. yes having this automated formatter would be nice.

nhatcher:
This will need to be done again at this point

Issue #39, opened 2024-04-18
Create Web Application


nhatcher:
First version of Web skin done in React/TypeScript.

Basic spreadsheet features. Everything in <https://playground.ironcalc.com/> plus

    Borders
    Vertical Align
    Copy Styles
    Text overflow goes to the next cell if empty
    Text coverage

nhatcher:
This was implemented in #79 .

Open issues will be on separate tickets:

* Browse mode
* Frontend tests
* Storyboard
* Text overflows to next cell if empty

PR #79, opened 2024-08-11
UPDATE: Adds web app

nhatcher:
This PR finally introduces the frontend application

There are several changes in the API to support all that is needed in the frontend.

TODO:

* browse mode is not present
* test coverage is zero (but test infrastructure is in place)
* storyboard removed
* It is not a react component (so cannot simply be installed with npm)
* Test coverage in the rust side is insufficient
* Documentation in the Rust side in insufficient
* Documentation in the frontend is non existent

Despite all those missing features I decided to merge this to get moving


bindings/wasm/types.js
```diff
@@ -0,0 +1,23 @@
+export var BorderType;
```
nhatcher: This file needs to be removed

webapp/.storybook/main.ts
```diff
@@ -0,0 +1,21 @@
+import type { StorybookConfig } from '@storybook/react-vite';
```
nhatcher: This file needs to be removed, belongs to the storyboard

webapp/.storybook/preview.tsx
```diff
@@ -0,0 +1,29 @@
+import type { Preview } from '@storybook/react';
```
nhatcher: This file needs to be removed, belongs to the storyboard

webapp/src/App.css
```diff
@@ -0,0 +1,6 @@
+#root {
+    position: absolute;
+    inset: 10px;
+    border: 1px solid #AAA;
+    border-radius: 4px;
```
nhatcher: Missing new line at the end of file, why is this not detected by biome?

nhatcher: Fixed.

webapp/.storybook/preview.tsx
nhatcher: Fixed

webapp/.storybook/main.ts
nhatcher: fixed.

bindings/wasm/types.js
nhatcher: Removed

Makefile
```diff
@@ -10,7 +11,10 @@ tests: lint
 	./target/debug/documentation
 	cmp functions.md wiki/functions.md || exit 1
 	make remove-artifacts
-	cd bindings/wasm/ && wasm-pack build --target nodejs && node tests/test.mjs
+	# Regretabbly we need to build the wasm twice, once for the nodejs tests
+	# and a second one for the vitest.
+	cd bindings/wasm/ && wasm-pack build --target nodejs && node tests/test.mjs && make
+	cd webapp && npm run test
```
nhatcher, line 17: I think we need to run `npm install` here too

nhatcher: Since this run first the `lint` branch and that runs the install we don't need this ATM.

nhatcher:
Closes #39

PR #118, opened 2024-10-26
Bugfix/nicolas cut and paste

nhatcher:


base/src/user_model/common.rs
```diff
@@ -882,6 +878,122 @@ impl UserModel {
                 });
             }
         }
+        // bottom of the cells above the first
```
jaycarlton, line 881: These sections seem to be repeating a good deal of shape/logic. Is it possible to pull out a helper method for setting these borders?

PR #122, opened 2024-11-05
stringify test cases and some comments

jaycarlton:
Added a few test cases to help learn bits of the system. Added some comments while exploring the code.

### DIFF
--- a/base/src/expressions/lexer/mod.rs
+++ b/base/src/expressions/lexer/mod.rs
@@ -149,16 +149,17 @@ impl Lexer {
 
     /// Checks the next token without advancing position
     /// See also [advance_token](Self::advance_token)
+    /// TODO: is there not a way to do this without mutating `self`?
     pub fn peek_token(&mut self) -> TokenType {
-        let position = self.position;
+        let orig_position = self.position;
         let tk = self.next_token();
         self.next_token_position = Some(self.position);
-        self.position = position;
+        self.position = orig_position;
         tk
     }
 
     /// Advances position. This is used in conjunction with [`peek_token`](Self::peek_token)
-    /// It is a noop if the has not been a previous peek_token
+    /// It is a noop if there has not been a previous peek_token
     pub fn advance_token(&mut self) {
         if let Some(position) = self.next_token_position {
             self.position = position;
@@ -446,6 +447,7 @@ impl Lexer {
 
     // Private methods
 
+    // Move position to the end of input and return a `LexerError` value.
     fn set_error(&mut self, message: &str, position: usize) -> LexerError {
         self.position = self.len;
         LexerError {
@@ -454,6 +456,7 @@ impl Lexer {
         }
     }
 
+    // Show the next char without advancing, if there is a next char.
     fn peek_char(&mut self) -> Option<char> {
         let position = self.position;
         if position < self.len {
@@ -463,6 +466,7 @@ impl Lexer {
         }
     }
 
+    // Returns an error unless `ch_expected` matches the char at the current position.
     fn expect_char(&mut self, ch_expected: char) -> Result<()> {
         let position = self.position;
         if position >= self.len {
@@ -483,17 +487,18 @@ impl Lexer {
         Ok(())
     }
 
+    // Consume (read) the character at the curent position.
     fn read_next_char(&mut self) -> Option<char> {
-        let position = self.position;
-        if position < self.len {
-            self.position = position + 1;
-            Some(self.chars[position])
+        let orig_position = self.position;
+        if orig_position < self.len {
+            self.position = orig_position + 1;
+            Some(self.chars[orig_position])
         } else {
             None
         }
     }
 
-    // Consumes an integer from the input stream
+    // Consumes an integer from the input stream. Does not support thousands separator or sign.
     fn consume_integer(&mut self, first: char) -> Result<i32> {
         let mut position = self.position;
         let len = self.len;
@@ -581,6 +586,7 @@ impl Lexer {
     // Consumes an identifier from the input stream
     fn consume_identifier(&mut self) -> String {
         let mut position = self.position;
+        // Advance to the final character in the identifier.
         while position < self.len {
             let next_char = self.chars[position];
             if next_char.is_alphanumeric() || next_char == '_' || next_char == '.' {
@@ -594,6 +600,7 @@ impl Lexer {
         chars
     }
 
+    // Consumes a double-quote delimited string value.
     fn consume_string(&mut self) -> String {
         let mut position = self.position;
         let len = self.len;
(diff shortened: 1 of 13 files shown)


base/src/expressions/lexer/mod.rs
```diff
@@ -149,16 +149,17 @@ impl Lexer {
 
     /// Checks the next token without advancing position
     /// See also [advance_token](Self::advance_token)
+    /// TODO: is there not a way to do this without mutating `self`?
```
nhatcher, line 152: This kind of questions shouldn't be in code. The reason is that `peek_token` advances te token position. Otherwise we would hace to parse it all over again if we "accept" it

base/src/expressions/lexer/mod.rs
```diff
@@ -149,16 +149,17 @@ impl Lexer {
...
     pub fn peek_token(&mut self) -> TokenType {
-        let position = self.position;
+        let orig_position = self.position;
```
nhatcher, line 154: Please try not to rename variables unless there is a need for it or a previous discussion.

base/src/expressions/lexer/mod.rs
```diff
@@ -149,16 +149,17 @@ impl Lexer {
...
         let tk = self.next_token();
         self.next_token_position = Some(self.position);
-        self.position = position;
+        self.position = orig_position;
         tk
     }
 
     /// Advances position. This is used in conjunction with [`peek_token`](Self::peek_token)
-    /// It is a noop if the has not been a previous peek_token
+    /// It is a noop if there has not been a previous peek_token
```
nhatcher, line 162: Thanks!
annoying but, could you do this is a separate PR?

base/src/expressions/lexer/mod.rs
```diff
@@ -446,6 +447,7 @@ impl Lexer {
 
     // Private methods
 
+    // Move position to the end of input and return a `LexerError` value.
```
nhatcher, line 450: Let's discuss how to best comment code first. This is an awesome contribution but a bit more difficult than it looks like.
Let's do that is separate PRs please.

base/src/expressions/lexer/test/test_common.rs
```diff
@@ -10,6 +10,7 @@ use crate::expressions::{
     types::ParsedReference,
 };
 
+// TODO: why not jsut pass in a `LexerMode`? It's more expressive than a bool arg.
```
nhatcher, line 13: What do you mean by that? Maybe we can  discuss?

base/src/expressions/lexer/test/test_common.rs
```diff
@@ -35,8 +36,8 @@ fn test_number_integer() {
 }
 #[test]
 fn test_number_pi() {
-    let mut lx = new_lexer("3.415", true);
-    assert_eq!(lx.next_token(), Number(3.415));
+    let mut lx = new_lexer("3.14159", true);
+    assert_eq!(lx.next_token(), Number(3.14159));
```
nhatcher, line 40: Why are you doing this change? Do you think it covers another case?

base/src/expressions/lexer/test/test_common.rs
```diff
@@ -685,3 +686,15 @@ fn test_comparisons() {
...
+fn test_whitespace() {
+    let mut lx = new_lexer("   ", true);
+    assert_eq!(lx.next_token(), EOF);
+}
+
+#[test]
+fn test_empty() {
+    let mut lx = new_lexer("", true);
+    assert_eq!(lx.next_token(), EOF);
+}
```
nhatcher, line 700: These two are nice to have, maybe in a different PR?

base/src/expressions/parser/mod.rs
```diff
@@ -111,7 +112,10 @@ pub enum Node {
     },
     RangeKind {
         sheet_name: Option<String>,
+        // index into vector (0-based)
```
nhatcher, line 115: Yes, they are, but I feel we should document this at the API level and not clutter code here.

Kind of the leitmotif is that a cell is referenced by three numbers `(sheet, row, column)` in that order when sheet is an index (0-indexed) and (row, column) are 1 based. But this is not the place to document this. Methinks

base/src/expressions/parser/mod.rs
```diff
@@ -111,7 +112,10 @@ pub enum Node {
...
         sheet_index: u32,
+        // coordinates of opposite corners of the range, e.g. A1:C4.
+        // true if cell 1 is anchored
```
nhatcher, line 118: `true` here means that the reference is something like `C$4` notice the `$` sign, making the reference 'absolute'

base/src/expressions/parser/mod.rs
```diff
@@ -202,8 +209,8 @@ impl Parser {
         let lexer = lexer::Lexer::new(
             "",
             lexer::LexerMode::A1,
-            get_locale("en").expect(""),
-            get_language("en").expect(""),
+            get_locale("en").expect("Failed to get locale"),
+            get_language("en").expect("Failed to get language"),
```
nhatcher, line 213: Thanks! We actually need to remove all expect from code, but that's a problem for a different day. Here expect, expect with message or unwrap are all equally bad. Let me take care of this.

base/src/expressions/parser/mod.rs
```diff
@@ -464,6 +474,7 @@ impl Parser {
                     Some(name) => self.get_sheet_index_by_name(name),
                     None => self.get_sheet_index_by_name(&context.sheet),
                 };
+                // todo: bust out helper function for this calculation
```
nhatcher, line 477: I don't know what you mean by this. Maybe let's talk?

base/src/expressions/types.rs
```diff
@@ -30,6 +30,7 @@ pub struct CellReference {
 #[derive(Clone)]
 pub struct CellReferenceRC {
     pub sheet: String,
+    // 1-based index
```
nhatcher, line 33: Again, rows and columns are *always* 1-based and sheets 0-based. I think we should only document it once.

base/src/types.rs
```diff
@@ -369,9 +369,11 @@ pub struct Font {
     #[serde(default = "default_as_false")]
     #[serde(skip_serializing_if = "is_false")]
     pub u: bool, // seems that Excel supports a bit more - double underline / account underline etc.
```
nhatcher, line 371: Correct, you could open a ticket for this. And actually this might be a nice first issue.

webapp/src/AppComponents/storage.ts
```diff
@@ -18,8 +18,8 @@ export function updateNameSelectedWorkbook(model: Model, newName: string) {
         console.warn("Failed saving new name");
       }
     }
-    const modeBytes = model.toBytes();
-    localStorage.setItem(uuid, bytesToBase64(modeBytes));
+    const modelBytes = model.toBytes();
```
nhatcher, line 21: ooops, thanks! Care to make a PR just for this?

base/src/expressions/parser/test_stringify.rs
```diff
@@ -0,0 +1,146 @@
+use super::Parser;
```
nhatcher, line 1: Hi @jaycarlton , these tests are amazing? But why did you add them? Did you add them because you want to learn how to use the API or because you are covering some case that wasn't covered before. If it is the second case we definitely need them in!

nhatcher (changes requested):
Hi @jaycarlton , thanks for all the work put in here and very sorry for my late reply :'(.

I _think_ our best course of action is to decline this PR and create some minor ones. I am happy to have a call with you to align our priorities.

nhatcher:
Non active. Will take this later

PR #124, opened 2024-11-10
Support csv import

AndrooFrowns:
Implement basic changes to permit CSV files to be converted to .ic

Modified paste csv function to not use a `&str` so that the data doesn't need to be entirely in memory, hopefully permitting large CSVs to be read.

not addressed yet, but noticed while implementing:

* export to CSV
* user/programmatic selection of CSV settings
* supporting strings rather than chars for comment indicator

### DIFF
--- a/xlsx/src/bin/xlsx_2_icalc.rs
+++ b/xlsx/src/bin/xlsx_2_icalc.rs
@@ -8,8 +8,10 @@
 
 use anyhow::{bail, Context, Result};
 use clap::Parser;
-use ironcalc::{export::save_to_icalc, import::load_from_xlsx};
-use ironcalc_base::Model;
+use ironcalc::{
+    export::{save_to_icalc, ModelType},
+    import::{load_from_csv, load_from_xlsx},
+};
 use std::path::{Path, PathBuf};
 
 #[derive(Parser)]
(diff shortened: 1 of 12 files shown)


nhatcher:
> I could try to switch strategies here or just make that a future enhancement, up to you.

There is not a time constraint, if you think there is a better way, go for the better way.

xlsx/src/bin/xlsx_2_icalc.rs
```diff
@@ -1,26 +1,121 @@
...
+    };
+
+    let output_path = create_output_path(args.output, &path)?;
+
+    let model = match kind {
+        FileKind::Xlsx => handle_xlsx(&path),
+        FileKind::Csv => handle_csv(&path),
+    }?;
+
+    save_to_icalc(&model, &output_path).with_context(|| "Failed to sasve file as .icalc")?;
```
jaycarlton, line 55: I thought the extension was `.ic`.

xlsx/tests/csv.rs
```diff
@@ -0,0 +1,40 @@
...
+use ironcalc::compare::{test_file, test_load_and_saving};
+use ironcalc::export::save_to_xlsx;
+use ironcalc::import::{load_from_icalc, load_from_xlsx, load_from_xlsx_bytes};
+use ironcalc_base::types::{HorizontalAlignment, VerticalAlignment};
+use ironcalc_base::Model;
+
+use std::process::{Command, ExitStatus};
+
+#[test]
+fn test_simple_csv() {
```
jaycarlton, line 14: There are a huge number of edge cases in CSV files that you'll want to test. Escaped (or unescaped) commas inside strings, function calls, misaligned column count, etc.

AndrooFrowns: This is a good point, however as I'm relying on the implementation we already have for copy/paste purposes, I tried to stay focused the parts relating to getting a file in rather than improving the csv parsing itself.

If it's a blocker I can add it as well when I get time to work on this, but the distinction made sense in my head.

xlsx/src/bin/xlsx_2_icalc.rs
AndrooFrowns: That might be, I just kept what this file was using. Is there a standard naming convention we should ensure is used everywhere?

nhatcher:
This will probably be needing a full redo

PR #128, opened 2024-11-10
Implement UNICODE function

ExtraBB:
First contribution, thought I'd start with something simple. Let me know if I missed anything.


nhatcher:
This is great!!

For what I can see you are just missing a test workbook. You just need to create a workbook with Excel called UNICODE.xlsx and place it here:

The heavy lifting of the tests should be in that file

nhatcher (changes requested):
Missing test workbook

ExtraBB:
Thanks, I completely missed the `xlsx` tests but they are very useful! Caught a bug writing to the wrong Excel formatting and missing the `_xlfn` variant of the function. The branch is updated

nhatcher:
Hi @ExtraBB , don't worry about the test coverage.
I will get around doing a proper review on my evening (Berlin time)

nhatcher:
Awesome, sorry for the late reply. The test function is great but doesn't contain error cases:

image

Also can you color those cells that have a formula:

image

Then it is easy to see for instance that A20 has a formula (as opposed to be empty)

nhatcher:
You also need to add `UNICODE` here:

https://github.com/ironcalc/IronCalc/blob/main/base/src/functions/mod.rs#L452-L481

ExtraBB:
@nhatcher updated with error cases.

base/src/test/test_fn_unicode.rs
```diff
@@ -0,0 +1,61 @@
...
+    assert_eq!(model._get_text("A3"), *"#NAME?");
+    assert_eq!(model._get_text("A4"), *"#VALUE!");
+    assert_eq!(model._get_text("A5"), *"#REF!");
+    assert_eq!(model._get_text("A6"), *"#DIV/0!");
+}
+
+#[test]
+fn wrong_number_of_arguments() {
+    let mut model = new_empty_model();
+    model._set("A1", "=UNICODE()");
```
nhatcher, line 57: One last thing, two arguments?

nhatcher (changes requested):

base/src/test/test_fn_unicode.rs
exbb123bb: Updated 👍

base/src/functions/text.rs
```diff
@@ -342,6 +342,50 @@ impl Model {
...
+
+            if s.len() == 0 {
+                return CalcResult::Error {
+                    error: Error::VALUE,
+                    origin: cell,
+                    message: "Empty cell".to_string(),
+                };
+            }
+
+            let unicode_number = s.chars().nth(0).unwrap() as u32;
```
nhatcher: Sorry, I missed this. We can't use `unwrap` (or `panic`) in the library code.

nhatcher: Maybe something like:

```Rust
match s.chars().next() {
    Some(c) => {
        return CalcResult::Number(c as f64);
    },
    None => {
        return CalcResult::Error {
            error: Error::VALUE,
            origin: cell,
            message: "Empty cell".to_string(),
        };
    }
}
```

nhatcher (changes requested):
Sorry, I missed the unwrap.

I will add this to clippy eventually to make it impossible soon enough.

base/src/functions/text.rs
exbb123bb: Updated! Can't cast `char` directly to `f64`, has to go through `u32` first

nhatcher:
Approved and merged!

Issue #14, opened 2024-03-13
Use a streaming parser instead of roxmltree


nhatcher:
Currently we use the fantastic roxmltree parser to import from Excel and read xml files.

This is great but choques on really large files.

Maybe we can use https://github.com/tafia/quick-xml (streaming parser) instead of the roxmltree to be able to import/export 1 million rows easily.

See https://data.cityofnewyork.us/Social-Services/DEP-311-Service-Requests/vwtg-azuw/about_data for instance to get some huge csv files

PR #134, opened 2024-11-14
Parse XML using streaming API instead of tree API.

ephraimkunz:
I saw #14 which looked fun so thought I'd take a crack at it. The goal was to reduce peak memory usage and execution time by using quick_xml for pull-based parsing, rather than roxmltree for tree-base parsing.

For now, I kept the old tree-based parser around in case we want to do further tests for performance / correctness. I don't think long term we'd want to keep both parsers though as it might be a maintenance burden. So happy to remove the old parser in this PR or later if we want to take the new one. The pull-based parser passes all tests and the output xml it writes is identical to what the library wrote before any of these changes. It's possible it regresses other things that the tests don't catch though, but I'd need some exotic XMLs to know for sure.

It looks like the worksheet xml files and the shared string xml files are the largest part of the xlsx and have the most to gain from stream parsing, so I only looked at those. Do we want to convert other uses of roxmltree to quick_xml? They don't show up in profiling so it would be just for consistency and not for performance reasons that we'd get rid of the remaining uses of roxmltree.

Here are the performance results. I tested the new and old parsers on all the files in xlsx/tests/calc_tests and verified no regressions in parsing speed were present on these small files. The majority of my testing was with a large file with the maximum number of rows Excel allows (1,048,576) and 39 columns (https://data.cityofnewyork.us/Social-Services/DEP-311-Service-Requests/vwtg-azuw/about_data). I tested reading in the Excel file (`load_from_xlsx`), calling `evaluate` on the model, then writing out the Excel file (`save_to_xlsx`). 

 **Tree (original)**
Peak RAM usage was 11.5 GB, took 51 seconds.

**Stream (new)**
Peak RAM usage was 3.3 GB, took 47 seconds.

Those are pretty good results, peak memory usage was reduced by almost 3.5x. That's a direct result of not pulling the whole XML file into memory at once. Time wasn't reduced by as much as I hoped, but I think it's because we are looking at most of XML tags in a file to parse it, and because we are mainly bottlenecked on unzipping the files.

Most of the remaining memory used in the stream parser is storing the nested HashMaps in `sheetData`. Reducing memory usage further here would involve changing the data layout or data structure of `sheetData`. 

Most of the time is spent zipping and unzipping. Further improvements here would be around exploring unzipping the files completely as separate files on disk, then parsing them. I think this would speed up parsing and keep the memory footprint low, but would temporarily increase the disk footprint and may leave files behind if the parser crashes, which wouldn't be an expected side effect of the call to `load_from_xlsx`. 

Anyway, happy to hear feedback here, it's my first time implementing a stream parser so I'm grateful for any suggestions.


xlsx/src/export/mod.rs
```diff
@@ -119,14 +119,12 @@ pub fn save_xlsx_to_writer<W: Write + Seek>(model: &Model, writer: W) -> Result<
...
-                sheet_dimension_str,
-                is_sheet_selected,
-            )
-            .as_bytes(),
+        worksheets::write_worksheet_xml(
+            worksheet,
+            &model.parsed_formulas[sheet_index],
+            sheet_dimension_str,
+            is_sheet_selected,
+            &mut zip,
```
ephraimkunz, line 129: I noticed that writing out the worksheet caused a memory spike of around 1 GB. That's because we were generating the whole worksheet as a String, then writing it in one go. Instead, pass in the writer and write it out incrementally to avoid the spike.

xlsx/src/export/worksheets.rs
```diff
@@ -51,15 +53,60 @@ fn get_formula_attribute(
...
+        let column_start = number_to_column(range[1]).unwrap_or("A".to_string());
+        let column_end = number_to_column(range[3]).unwrap_or("A".to_string());
+        if range[0] == range[2] && range[1] == range[3] {
+            sqref = format!("{column_start}{}", range[0]);
+        } else {
+            sqref = format!("{}{}:{}{}", column_start, range[0], column_end, range[2]);
+        }
+    }
+
+    write!(
```
ephraimkunz, line 98: Write to the zip file in chunks, to avoid memory spikes. But we want the chunks big enough to avoid overhead in `write!` so we'll do some batching below.

xlsx/src/export/worksheets.rs
```diff
@@ -90,8 +148,11 @@ pub(crate) fn get_worksheet_xml(
         row_style_dict.insert(row.r, row.clone());
     }
 
+    const ROW_STRING_COUNT_BEFORE_WRITE: usize = 10;
```
ephraimkunz, line 154: Empirically determined batch size for calling `write!` on rows. So we'll accumulate 10 rows worth of data before writing it. Smaller and I saw longer run times, larger and we use more memory with the same runtime.

xlsx/src/export/worksheets.rs
```diff
@@ -90,8 +148,11 @@ pub(crate) fn get_worksheet_xml(
...
+    let mut row_strings: Vec<String> = Vec::with_capacity(ROW_STRING_COUNT_BEFORE_WRITE);
+
     for (row_index, row_data) in worksheet.sheet_data.iter().sorted_by_key(|x| x.0) {
-        let mut row_data_str: Vec<String> = vec![];
+        let mut row_data_str: Vec<String> = Vec::with_capacity(row_data.len());
```
ephraimkunz, line 158: Avoid lots of reallocations in the hot path as the vector grows.

xlsx/src/import/shared_strings.rs
```diff
@@ -1,39 +1,121 @@
...
+    }
+}
+
+#[derive(Debug, PartialEq)]
+enum ParserState {
+    OutsideSi,
+    InsideSi,
+    T,
+}
+
```
ephraimkunz, line 65: Pretty simple parser for the shared strings file. It isn't too big, but this gives us some memory savings and some time savings.

xlsx/src/import/shared_strings.rs
```diff
@@ -1,39 +1,121 @@
...
-        let text = node
-            .descendants()
-            .filter(|n| n.has_tag_name("t"))
-            .map(|n| n.text().unwrap_or("").to_string())
-            .collect::<Vec<String>>()
-            .join("");
-        shared_strings.push(text);
+fn read_shared_strings_from_reader<R: Read>(reader: &mut R) -> Result<Vec<String>, XlsxError> {
+    let streaming = true;
+    if streaming {
```
ephraimkunz: Right now this is how I'm keeping both implementations around. Should we do this in a different way, or remove the old implementation? I left them both here for now in case we want to do further perf / correctness testing. Changing `streaming` to false gives the old tree-based parsing behavior for shared strings.

xlsx/src/import/shared_strings.rs
```diff
@@ -67,7 +149,8 @@ mod tests {
         </r>
     </si>
 </sst>"#;
-        let shared_strings = read_shared_strings_from_string(xml_string.trim()).unwrap();
+        let shared_strings =
+            read_shared_strings_from_reader(&mut xml_string.trim().as_bytes()).unwrap();
```
ephraimkunz, line 133: Takes in a reader now, to hide the fact that the old implementation reads the whole file into a string before processing it. The new implementation streams it in, so make this function take a reader to hide that.

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,535 @@
+use core::str;
+use std::{collections::HashMap, io::Read, num::ParseIntError};
```
ephraimkunz: I put all the common parsing code between both implementations here. I think everything in this file just came directly from the old tree-based parser, but we need a common place so the new streaming parser can call into these functions.

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,535 @@
...
+                format!("xl/{path}")
+            };
+            let settings = SheetSettings {
+                name: sheet_name.to_string(),
+                id: sheet.sheet_id,
+                state: state.clone(),
+                comments: comments.get(rel_id).expect("").to_vec(),
+            };
+
+            let streaming = true;
```
ephraimkunz: Here's where we can switch between the tree and streaming parsers. Same questions apply to this variable as did to the variable by the same name in shared_strings.rs.

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,535 @@
...
+    #[error("RowError: {0}")]
+    RowError(ParseIntError),
+    #[error("ColumnError: {0}")]
+    ColumnError(String),
+}
+
+// This parses Sheet1!AS23 into sheet, column and row
+// FIXME: This is buggy. Does not check that is a valid sheet name
+// There is a similar named function in ironcalc_base. We probably should fix both at the same time.
+// NB: Maybe use regexes for this?
```
ephraimkunz: Kept all the old comments.

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz, line 1: This file contains the new streaming parser implementation.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1049 @@
...
+    shared_strings: &mut Vec<String>,
+) -> Result<(Worksheet, bool), XlsxError> {
+    let mut sheet_parser = SheetParser::new(settings, shared_strings, worksheets, tables);
+
+    let zipfile = archive.by_name(path)?;
+    let xmlfile = BufReader::new(zipfile);
+    let mut xmlfile = quick_xml::Reader::from_reader(xmlfile);
+    xmlfile.config_mut().expand_empty_elements = true;
+
+    const BUF_SIZE: usize = 700;
```
ephraimkunz: Empirically determined. This was the max element size in the xlsx I was testing. If it's too small, not a big deal as quick_xml will enlarge it if needed.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1049 @@
...
+    cell_value: Option<String>,
+    value_metadata: Option<String>,
+    cell_type: Option<String>,
+    cell_style: i32,
+    formula_index: i32,
+    cell_ref: String,
+    column: i32,
+
+    formula_data: FormulaData,
+}
```
ephraimkunz: Hold temporary data while we parse until we have enough to construct a cell.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1049 @@
...
+            tag.unescape()
+                .ok()
+                .map(|t| t.to_string())
+                .unwrap_or_default(),
+        );
+
+        Ok(())
+    }
+
+    fn process(&mut self, ev: Event) -> Result<(), XlsxError> {
```
ephraimkunz: The state machine that we use the parse in a streaming manner.

xlsx/src/import/worksheets/worksheet_tree.rs
```diff
@@ -1,99 +1,19 @@
...
-        token::{get_error_by_english_name, Error},
-        types::CellReferenceRC,
-        utils::{column_to_number, parse_reference_a1},
-    },
-    types::{
-        Cell, Col, Comment, DefinedName, Row, SheetData, SheetState, Table, Worksheet,
-        WorksheetView,
-    },
+    expressions::utils::column_to_number,
+    types::{Col, Row, SheetData, Table, Worksheet, WorksheetView},
```
ephraimkunz: I've done nothing to change the tree parser, just moved into a new file and changed some imports.

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,535 @@
...
+    Ok(CellReferenceRC {
+        sheet: sheet_name,
+        row: row.parse::<i32>().map_err(ParseReferenceError::RowError)?,
+        column: column_to_number(&column).map_err(ParseReferenceError::ColumnError)?,
+    })
+}
+
+fn get_column_from_ref(s: &str) -> String {
+    s.chars().filter(|c| !c.is_ascii_digit()).collect()
+}
```
ephraimkunz: This surprisingly was a big perf win compared to the old version:
```
    let cs = s.chars();
    let mut column = Vec::<char>::new();
    for c in cs {
        if !c.is_ascii_digit() {
            column.push(c);
        }
    }
    column.into_iter().collect()
    ```
    
    That old version had a lot more allocations to grow the vector, and this method is called very often.

**Comment by nhatcher** [comment]
This is amazing @ephraimkunz ! Looking forward to testing this myself!

Did you do any comparison with (sane author as quick-xml)
https://github.com/tafia/calamine/

Whatever magic Johann Tuffe is doing there I'm sure we can learn a lot.

Thanks!

**Comment by ephraimkunz** [comment]
I wasn't aware of that crate. Let me do some comparisons and report back here.

**Comment by nhatcher** [comment]
Hi @ephraimkunz , had a quick glance at the PR. I don't think we should keep around the old parser. And definitely shouldn't keep roxmltree as a dependency. I realize that is a larger change, so we can go step by step.

File: xlsx/src/import/worksheets/mod.rs [hunk 16]
```diff
@@ -0,0 +1,552 @@
...
+            };
+            comments.insert(
+                &sheet.id,
+                load_sheet_rels(archive, &path, tables, &sheet.name)?,
+            );
+        }
+    }
+
+    // load all sheets
+    let worksheets: &Vec<String> = &workbook.worksheets.iter().map(|s| s.name.clone()).collect();
```
jaycarlton: maybe name this `worksheet_names` for clarity

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+    let mut sheets = Vec::new();
+    let mut selected_sheet = 0;
+    let mut sheet_index = 0;
+    for sheet in &workbook.worksheets {
+        let sheet_name = &sheet.name;
+        let rel_id = &sheet.id;
+        let state = &sheet.state;
+        let rel = &rels[rel_id];
+        if rel.rel_type.ends_with("worksheet") {
```
jaycarlton: This chunk looks to be repeated from above. Is there a way to avoid this duplication?

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+            let path = if let Some(p) = path.strip_prefix('/') {
+                p.to_string()
+            } else {
+                format!("xl/{path}")
+            };
+            let settings = SheetSettings {
+                name: sheet_name.to_string(),
+                id: sheet.sheet_id,
+                state: state.clone(),
+                comments: comments.get(rel_id).expect("").to_vec(),
```
jaycarlton: Put in an error message in `expect()`.

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+
+use super::{
+    tables::load_table,
+    util::{get_attribute, get_color},
+};
+
+mod worksheet_streaming;
+mod worksheet_tree;
+
+pub(crate) struct Sheet {
```
jaycarlton: I'd probably name this `SheetInfo` or something to emphasize tha tit doesn't contain the rows in the sheet.

xlsx/src/import/worksheets/mod.rs
jaycarlton: Maybe declare it as `const` to emphasize its compile-time const-ness.

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+) -> Result<Vec<Comment>, XlsxError> {
+    // ...xl/worksheets/sheet6.xml -> xl/worksheets/_rels/sheet6.xml.rels
+    let mut comments = Vec::new();
+    let v: Vec<&str> = path.split("/worksheets/").collect();
+    let mut path = v[0].to_string();
+    path.push_str("/worksheets/_rels/");
+    path.push_str(v[1]);
+    path.push_str(".rels");
+    let file = archive.by_name(&path);
+    if file.is_err() {
```
jaycarlton: Why do we return `Ok` if there's an error?

xlsx/src/import/worksheets/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+    if comment_list.len() == 1 {
+        for comment in comment_list[0].children() {
+            let text = comment
+                .descendants()
+                .filter(|n| n.has_tag_name("t"))
+                .map(|n| n.text().unwrap().to_string())
+                .collect::<Vec<String>>()
+                .join("");
+            let cell_ref = get_attribute(&comment, "ref")?.to_string();
+            // TODO: Read author_name from the list of authors
```
jaycarlton: Is it worth doing this in the current PR?

xlsx/src/import/shared_strings.rs
jaycarlton: If you keep it, I'd put in a compile-time switch so you don't have to edit the file.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+            column: Default::default(),
+            formula_data: Default::default(),
+        }
+    }
+}
+
+#[derive(Debug, Default)]
+struct FormulaData {
+    formula_type: String,
+    formula_si: Option<String>,
```
jaycarlton: What does SI stand for?

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+    current_sheet_view: SheetView,
+    colors: Vec<Option<String>>,
+    sheet_data: SheetData,
+    current_data_row: HashMap<i32, Cell>,
+    current_cell_data: CellData,
+    current_row_index: i32,
+    rows: Vec<Row>,
+    cols: Vec<Col>,
+    shared_formulas: Vec<String>,
+    merge_cells: Vec<String>,
```
jaycarlton: I'd name this `merged_cells`.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+
+        Ok(())
+    }
+
+    fn load_current_sheet_view_pane(&mut self, tag: BytesStart) -> Result<(), XlsxError> {
+        // 18.18.53 ST_PaneState (Pane State)
+        // frozen, frozenSplit, split
+        if let Some(state) = get_optional_attribute_streaming(&tag, "state")? {
+            if state == "frozen" {
+                // TODO: Should we assert that topLeft is consistent?
```
jaycarlton: Yeah, checking that the range isn't inside-out sounds like a good practice.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+            _ => None,
+        };
+
+        let sqref = match get_optional_attribute_streaming(&tag, "sqref")?.map(|s| parse_range(&s))
+        {
+            Some(Ok(s)) => Some(s),
+            _ => None,
+        };
+
+        let (selected_row, selected_column, row1, column1, row2, column2) =
```
jaycarlton: This part would be easier to read as a struct than as tuples.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+                            "Unable to decode attribute: \"{:?}\": {:?}",
+                            attr_name,
+                            e.to_string()
+                        ))
+                    })?;
+
+                    height_attribute = Some(value.parse().unwrap_or(default_row_height));
+                }
+                attr_name @ b"customHeight" => {
+                    let value = std::str::from_utf8(&attribute.value).map_err(|e| {
```
jaycarlton: This line is repeated several times. Maybe factor out a small helper function (assuming it can be inlined)?

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+            .insert(self.current_cell_data.column, cell);
+        self.current_cell_data.set_to_default_values();
+
+        Ok(())
+    }
+
+    fn load_cell(&mut self, tag: BytesStart) -> Result<(), XlsxError> {
+        // 18.3.1.4 c (Cell)
+        // Child Elements:
+        // * v: Cell value
```
jaycarlton: Would it make sense to make an enum for these element types?

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+                } else {
+                    // It's a daughter cell
+                    match self.index_map.get(&si) {
+                        Some(index) => {
+                            self.current_cell_data.formula_index = *index;
+                        }
+                        None => {
+                            // Haven't bumped into the mother cell yet. We insert a placeholder.
+                            // Note that it is perfectly possible that the formula of the mother cell
+                            // is already in the set of array formulas. This will lead to the above mention duplicity.
```
jaycarlton: *mentioned

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+                let formula = self.current_cell_data.formula_data.formula_text.clone();
+                let context = format!("{}!{}", self.settings.name, self.current_cell_data.cell_ref);
+                let formula =
+                    from_a1_to_rc(formula, self.worksheets, context, self.tables.clone())?;
+
+                match get_formula_index(&formula, &self.shared_formulas) {
+                    Some(index) => self.current_cell_data.formula_index = index,
+                    None => {
+                        self.shared_formulas.push(formula);
+                        self.current_cell_data.formula_index =
```
jaycarlton: This line seems to come up often. You could make a little `append_formula()` method that pushed and updated the index in one shot.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+                attr_name,
+                e.to_string()
+            ))
+        })
+}
+
+fn get_number_streaming(tag: &BytesStart, attr_name: &str) -> i32 {
+    get_optional_attribute_streaming(tag, attr_name)
+        .ok()
+        .unwrap_or(Some("0".into()))
```
jaycarlton: Why do we need to specify the default three times? I'm new to Rust.

xlsx/src/import/worksheets/worksheet_streaming.rs
```diff
@@ -0,0 +1,1034 @@
...
+        .unwrap_or("0".into())
+        .parse::<i32>()
+        .unwrap_or(0)
+}
+
+pub(super) fn get_color_streaming(tag: BytesStart) -> Result<Option<String>, XlsxError> {
+    // 18.3.1.15 color (Data Bar Color)
+    if let Some(mut val) = get_optional_attribute_streaming(&tag, "rbg")? {
+        // FIXME the two first values is normally the alpha.
```
jaycarlton: Does this need to be fixed before submitting?

jaycarlton (commented):
Nice submission. I learned a lot going through it.

ephraimkunz:
Sorry for the delay here - I just started a new job and it's taking a lot of my time recently. 

Thanks for the helpful review. I'll take a look soon. 
I am also still planning to do some benchmarking vs [calamine](https://lib.rs/crates/calamine), [umya-spreadsheet](https://lib.rs/crates/umya-spreadsheet) and [ooxml](https://lib.rs/crates/ooxml). 
I may also look at the test coverage of this new parser and the possibilities around adding more tests or fuzzing it.

xlsx/src/import/worksheets/mod.rs
ephraimkunz: I'm keeping this consistent with lots of other code that refers to the list of worksheet names as "worksheets".

xlsx/src/import/worksheets/mod.rs
ephraimkunz: I don't think so, the PR is already big enough. This was an existing TODO comment I copied over.

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz: This was an existing FIXME in the code, I just left it. So I would say not in this PR.

xlsx/src/import/shared_strings.rs
ephraimkunz: Removed it in latest commit.

xlsx/src/import/worksheets/mod.rs
ephraimkunz: Removed in latest PR

xlsx/src/import/worksheets/mod.rs
ephraimkunz: Existing struct name in the code. I think we should handle all TODO fixes and renames in a future PR.

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz: Existing code, but probably.

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz: Existing comment

xlsx/src/import/worksheets/mod.rs
ephraimkunz: Looks like the existing code decides to return an empty vec in the case of an error. Maybe this should be fixed, but not in this PR (existing issue).

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz: According to ChatGPT, "shared string item". This property stores the "si" attribute in the formula XML.

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz: `merge_cells` is the existing name in the codebase. Maybe this can happen in a future PR.

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz: Existing comment

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz: Existing code in this file.

xlsx/src/import/worksheets/worksheet_streaming.rs
ephraimkunz: We don't, I just don't always remember my Rust combinators.

xlsx/src/import/worksheets/mod.rs
ephraimkunz: Existing code duplication, probably a way to avoid this but since it only repeats in 2 places I think its ok for now.

ephraimkunz:
> Hi @ephraimkunz , had a quick glance at the PR. I don't think we should keep around the old parser. And definitely shouldn't keep roxmltree as a dependency. I realize that is a larger change, so we can go step by step.

Agreed. I removed the tree-based implementation in this last push which had the unfortunate side effect of changing the diff by a lot, because I combined everything back into 1 file since we don't have two different implementations anymore.

Roxmltree is still used in a couple places but those should be easy to fix, following the model of `shared_strings.rs`, where I implemented a simple state machine to replace it there.

ephraimkunz:
I also did a bunch of perf tests against other crates (what looked to be the top XLSX parsers on [lib.rs](lib.rs)) on my same large file with > 1 million rows and 39 columns. Here are the results:

[OOXML](https://lib.rs/crates/ooxml) was significantly worse than IronCalc. It doesn't do streaming so it reads the whole file into memory. Super slow as well, I gave up after it ran for 700 seconds.

[Umya-spreadsheet](https://lib.rs/crates/umya-spreadsheet) wasn't great. It was the worse for RAM usage out of any crates I tested, clocking in at a max of 16 GB to read the file. It took 2 minutes, 30 seconds to complete.

[Calamine](https://lib.rs/crates/calamine) was of course the crate to beat, and comparison to it was what kicked off this whole performance benchmark. It seemed to perform slightly worse in terms of time (17 seconds vs this PR's 14 seconds) and slightly better in terms of memory (2.7 GB peak vs this PR's 3.2 GB). However, that's not the full story. It's a lazy parser and doesn't cache the results of its lazy evaluation. This means you can grab cell data in one call, but if you try to grab all cell data again, it re-parses all the cell data again. Furthermore, fetching formulas is a 2nd call that also re-parses all data. So, if I want it to parse all the data in the xlsx file and return it in a single struct like this xlsx does, it would require around 30 seconds although peak memory would still be slightly lower.

In conclusion, I think we're competitive with Calamine as well as supporting more xlsx features than that library does (for the most part, they seem to support images which we don't). Switching to Calamine would be a major change since it is lazy and depending on our access patterns might require multiple re-parses of the data. It would be interesting to explore more about how it maintains a lower memory footprint while parsing, which I think it mostly due to how the cell data is stored (using a vector of (rowID, colID, data) rather than what we do, which is doubly-nested hashmap of rowID:colID:Cell).

ephraimkunz:
@jaycarlton, @nhatcher This is ready to be reviewed again.

nhatcher:
Hi @ephraimkunz , this is great news!!!
Amazing job really 

I will have a look as soon as possible.

In the meantime, why are we using roxmltree in a couple of places? If its a matter of time I can probably help out

ephraimkunz:
I had originally converted just the performance sensitive parts to use quick-xml. I should be able to look at converting the rest to quick-xml and removing roxmltree today.

ephraimkunz:
@nhatcher I pushed changes to make a few more places use streaming parsers. It looks like there are still several places left to convert:
- Comments
- Sheet relationships
- Workbook
- Tables
- Styles

Several of these have pretty complicated XML meaning it will take longer. I wonder if we should consider reviewing and merging this PR as-is, then file separate tasks for each of those, with the goal of removing the roxmltree dependency once all those are converted to using streaming parsers?

I don't have a lot of time to work on this and I think there's a lot more work to remove the dependency completely in this PR.

nhatcher:
> I don't have a lot of time to work on this and I think there's a lot more work to remove the dependency completely in this PR.

Hi @ephraimkunz , that's tough. The problem is that in situations like this is not difficult to end up in a hybrid state where half of the work is done following one framework (or library or whatever)  and the other half the other. Then a third party wants to add something or fix a bug and it would be difficult for them to know what to do. I think you have done more than half of the work already and I am tempted to go ahead and bite the bullet and not let this good work go to waste.

As for the more technical side. This is really amazing work and leaves us in an unbeatable situation. In we are in the same ballpark as Calamine we are already wining. Remember that we actually parse formulas!

I want to give this a little bit more though, but I will likely just merge this as is by the end of the day and open tickets for the remaining tasks.

Once again thank you for your work. This has an enormous impact. If someone in a couple of months form now opens a beefy workbook and IronCalc opens it like a champ it might be because of this PR. And they might choose the technology because of your work here.

ephraimkunz:
No problem, it's been fun learning more. I should be able to help with some of those tasks as I get time.

PR #135, opened 2024-11-14
Add Tiron to this repo

ephraimkunz:
As discussed in https://github.com/ironcalc/TironCalc/pull/3, I moved the contents to this repo and got it building in the Cargo Workspace. Once this lands, we can remove that repo.


base/src/actions.rs
```diff
@@ -383,11 +383,13 @@ impl Model {
     ///    * All cell references to initial_column will go to target_column
     ///    * All cell references to columns in between (initial_column, target_column] will be displaced one to the left
     ///    * All other cell references are left unchanged
+    ///
```
nhatcher, line 386: Let's not do this format related changes in this repo

nhatcher (changes requested):
Hi @ephraimkunz ,

Thanks! I I feel this is a bigger project. We will need to:

* Get code up to speed (removing unwraps, for instance)
* Figure out a way to have tests
* Maybe we want to preserve git history?

Let me know when you think. We can also have a call and discuss

nhatcher:
Hi @ephraimkunz , closing this for now

PR #146, opened 2024-11-27
Feature/nicolas formulatext

nhatcher:


docs/src/.vitepress/theme/style.css
```diff
@@ -175,20 +175,20 @@ li::marker {
   top: 36px;
 }
 
-.VPSwitch[data-v-70a26bb8] {
```
nhatcher, line 178: Hi @dg-ac , you can't do this. This selector works in your own build. The tag `70a26bb8` is different each time and it is different in the build. That's why these few selectors weren't working. I fixed with a HACK, using `important!`.

What this does is effectively "escaping" the specificity rules of CSS. It is generally a bad idea because you break the rules and people will not be able to modify your content. It's a bit like a bad cook will hack their way with some Maggi cubes instead of using proper ingredients.

dg-ac: Hi @nhatcher,

Gotcha! Will keep in mind, thanks for the explanation!

PR #148, opened 2024-11-28
UPDATE: Webapp cosmetics

dg-ac:
Hi @nhatcher,

To save you some time, I'm trying to polish a bunch of things in the webapp myself. I've started with the File menu, but if everything goes well I'll continue with other components like the Import popup or the top bar itself.

Anyways, this time I've changed only 2 things:
- Adjusted divider on fileMenu
- Replaced the big dot symbol we were using for selected selected workbooks with a check icon

Let me know what you think!

D

PR #154, opened 2024-11-29
Bugfix/dani file menu

nhatcher:
Hi Dani, this PR contains two commits:

1. Is yours from #148 
2. Is mine by running "npm run check-write"

So this is basically #148 without the WIP commit


dg-ac:
Hi @nhatcher,

Thank you! In that case, I don't there's anything else I need to do here, right?

dg-ac (approved):
Just checked the changes, everything looks as expected 👍

nhatcher:
Your approval!

PR #160, opened 2024-12-04
UPDATE: Added a 'Help' button in the top header that leads to the docs

dg-ac:
Hi @nhatcher,

Small update here. Based on our Figmas, I'm adding a new button to the top bar of the webapp, this time leading to our recently released Docs. LEt me know what you think!

D

### DIFF
--- a/webapp/src/AppComponents/FileBar.tsx
+++ b/webapp/src/AppComponents/FileBar.tsx
@@ -35,6 +35,11 @@ export function FileBar(properties: {
         }}
         onDelete={properties.onDelete}
       />
+      <HelpButton
+        onClick={() => window.open("https://docs.ironcalc.com", "_blank")}
+      >
+        Help
+      </HelpButton>
       <WorkbookTitle
         name={properties.model.getName()}
         onNameChange={(name) => {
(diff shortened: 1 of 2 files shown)


webapp/src/AppComponents/FileBar.tsx
```diff
@@ -107,7 +125,7 @@ const Toast = styled("div")`
 `;
 
 const Divider = styled("div")`
-  margin: 10px;
+  margin: 0px 8px 0px 16px;
```
nhatcher, line 128: Left over from another PR?

webapp/src/AppComponents/FileBar.tsx
```diff
@@ -35,6 +35,11 @@ export function FileBar(properties: {
         }}
         onDelete={properties.onDelete}
       />
+      <HelpButton
+        onClick={() => window.open("https://docs.ironcalc.com", "_blank")}
```
nhatcher, line 39: Are you sure you want:

1. hard code the URL (if the docs change place the link will be broken)
2. Point to the general documentation. Wouldn't the documentation for web app be better? (https://docs.ironcalc.com/web-application/about.html)

nhatcher (changes requested):
A couple of minor questions, otherwise LGTM

webapp/src/AppComponents/FileBar.tsx
dg-ac: 1. Is there a better way for that?
2. Makes sense

webapp/src/AppComponents/FileBar.tsx
dg-ac: Hah no. Even if it's looks the opposite, it's a way to make logo, divider and File button look like they are at the same distance...

nhatcher:
Hi @dg-ac , the linter is failing, you need to run:

```bash
npm run check-write
```

Also you added a new commit with new changes, what are those?

nhatcher:
> Hi @dg-ac , the linter is failing, you need to run:
> 
> ```shell
> npm run check-write
> ```
> 
> Also you added a new commit with new changes, what are those?

Hi @dg-ac , I fixed this myself.

PR #161, opened 2024-12-04
UPDATE: Small fixes in Upload file Dialog

dg-ac:
@nhatcher Another PR with very small updates:

- Makes the link to documentation usable
- Changes a few tiny bits so the implementation is closer to what we had in the Figma mockups

Let me know if this works for you.

D


nhatcher (approved):
LGTM!

PR #162, opened 2024-12-05
Update fv files

stevethesleeve:
Files updated. All appears to be well apart from error when attempting to access the example xlsx file:
    "app IronCalc says
    Example file not found, or failed to load".

Nicolás, have I done something wrong?


nhatcher:
> Files updated. All appears to be well apart from error when attempting to access the example xlsx file: "app IronCalc says Example file not found, or failed to load".
> 
> Nicolás, have I done something wrong?

Hi @stevethesleeve , that's expected the workbook is still not in place. It will not be there until I do a deployment

stevethesleeve:
No problem.

Shall I continue preparing the function pages and raising PRs or would you prefer me to wait?

nhatcher:
You can continue with the PR, in general the models in app.ironcalc.com will never be available until deployment (PR merged)

docs/src/functions/financial/fv.md
```diff
@@ -4,8 +4,46 @@ outline: deep
...
 
-::: warning
-🚧 This function is implemented but currently lacks detailed documentation. For guidance, you may refer to the equivalent functionality in [Microsoft Excel documentation](https://support.microsoft.com/en-us/office/excel-functions-by-category-5f91f4e9-7b42-46d2-9bd1-63f26a86c0eb).
-:::
\ No newline at end of file
+FV can be used to calculate future value over a specified number of compounding periods. A fixed interest rate or yield is assumed over all periods, and a fixed payment or deposit can be applied at the start or end of every period.
+
+If your interest rate varies between periods, use the [FVSCHEDULE](./FVSCHEDULE) function instead of FV.
+<!--@include: ../markdown-snippets/standard-headings.md{2,2}-->
+<!--@include: ../markdown-snippets/standard-headings.md{3,3}-->
```
nhatcher: Are you sure you want to do that? IMHO this makes the Markdown much more difficult to read.
You could always have a template that you copy/paste every time, no?

docs/src/functions/financial/fv.md
```diff
@@ -4,8 +4,46 @@ outline: deep
...
+* *nper*. The number of compounding periods to be taken into account. While this will often be an integer, non-integer values are accepted and processed.
+* *pmt*. The fixed amount paid or deposited each compounding period.
+* *pv* (optional). The present value or starting amount of the asset (default 0).
+* *type* (optional). A logical value indicating whether the payment due dates are at the end (0) of the compounding periods or at the beginning (any non-zero value). The default is 0 when omitted.
+<!--@include: ../markdown-snippets/standard-headings.md{5,5}-->
+* Make sure that the *rate* argument specifies the interest rate or yield applicable to the compounding period, based on the value chosen for *nper*.
+* The *pmt* and *pv* arguments should be expressed in the same currency unit. The value returned is expressed in the same currency unit.
+* To ensure a worthwhile result, one of the *pmt* and *pv* arguments should be non-zero.
+* The setting of the *type* argument only affects the calculation for non-zero values of the *pmt* argument.
+<!--@include: ../markdown-snippets/error-type-details.md-->
```
nhatcher, line 28: This, on the other hand, is great.

stevethesleeve:
With regards to using an include file as the source of the individual page headings...

I agree that a page template should be used to avoid repeated typing and I will be doing this (with or without the heading includes).

However, there was some logic in the proposed mechanism for defining the text of each sub-heading on a page. I was thinking about what might happen if we propose a change to a page heading after we have generated the 500+ pages for all functions. My thought was we could accomodate such a change easily by defining the headings separately, so that a change of heading text is achieved by editing only one file.

On the other hand, I'm sure we could write a script to change the heading texts in multiple files without having to do it by hand!

I'm happy to do it either way. Let me know your thoughts.

stevethesleeve:
Updated original files to:
(1) Remove markdown includes for text used in headings
(2) Remove spaces used in example xlsx file

PR #164, opened 2024-12-06
Implement SIGN()

sylvinus:
Hello!

I tried to implement a very simple function :-)

Let me know if I missed anything!

Thanks,


nhatcher (changes requested):
Hi @sylvinus , this is great!

You need to add a test file (created by Excel) in:

It should cover the happy paths and some of the errors. See, for instance the function ABS.xlsx note how cells with a formula are formatted differently

Issue #165, opened 2024-12-07
Fix: Webapp columns can be resized to "negative"


enjoythecode:
It is possible to drag the right-border of the column to the left of the left border (as if the width was "negative", so-to-say), which breaks the UI in multiple ways.

**Demonstration:**

![ironcalc-negative_column_width](https://github.com/user-attachments/assets/27ad4744-3042-44b7-9373-144aa40deca9)

**Expected behavior:**

Limit the minimum column size.

enjoythecode:
Hmm, the rust back-end should probably reject a negative width? @nhatcher do you think this should be handled on the UI or the "backend"? I am happy to make my first PR on this task.

enjoythecode:
Same issue exists with rows.

nhatcher:
Hi @enjoythecode , this is a great first task. It should be handled in both places.

Around:

https://github.com/ironcalc/IronCalc/blob/main/base/src/worksheet.rs#L300

insert:

```rust
if width < 0.0 {
    return Err(format!("Cannot set a negative width: {width}"));
}
```

you will need to add a regression test in:

https://github.com/ironcalc/IronCalc/blob/main/base/src/test/test_column_width.rs

In the front:

https://github.com/ironcalc/IronCalc/blob/main/webapp/src/components/worksheet.tsx#L107

Maybe something like:

```js
if (width > 0) {
      model.setColumnWidth(sheet, column, width);
      worksheetCanvas.current?.renderSheet();
}
```

Something similar for rows.

Please add an entry to the CHANGELOG under "fixed".

Many thanks!

nhatcher:
This is fixed now

PR #167, opened 2024-12-07
FIX: Prevent negative column width, row height

enjoythecode:
fixes !165

### DIFF
--- a/base/src/test/test_column_width.rs
+++ b/base/src/test/test_column_width.rs
@@ -86,22 +86,14 @@ fn test_column_width_higher_edge() {
 #[test]
 fn test_column_width_negative() {
     let mut model = new_empty_model();
-    let cols = vec![Col {
-        custom_width: false,
-        max: 16,
-        min: 5,
-        style: Some(1),
-        width: 10.0,
-    }];
-    model.workbook.worksheets[0].cols = cols;
     let result = model
         .workbook
         .worksheet_mut(0)
         .unwrap()
         .set_column_width(16, -1.0);
-    assert!(result.is_err());
-    assert_eq!(model.workbook.worksheets[0].cols.len(), 1);
+    assert_eq!(result, Err("Can not set a negative width: -1".to_string()));
+    assert_eq!(model.workbook.worksheets[0].cols.len(), 0);
     let worksheet = model.workbook.worksheet(0).unwrap();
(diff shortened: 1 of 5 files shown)


enjoythecode:
fixes ironcalc/ironcalc!165

enjoythecode:
hmm, I can't reference that issue apparently.

enjoythecode:
fixes #165

there we go.

enjoythecode:
@nhatcher  do you have a preferred way of reaching out to you when a PR is ready for review?

this is ready for your review, thanks for your guidance.

base/src/test/test_column_width.rs
```diff
@@ -82,3 +82,26 @@ fn test_column_width_higher_edge() {
...
+fn test_column_width_negative() {
+    let mut model = new_empty_model();
+    let cols = vec![Col {
+        custom_width: false,
+        max: 16,
+        min: 5,
+        style: Some(1),
+        width: 10.0,
+    }];
+    model.workbook.worksheets[0].cols = cols;
```
nhatcher: You don't need to do this, the default column widths will do.

base/src/test/test_column_width.rs
```diff
@@ -82,3 +82,26 @@ fn test_column_width_higher_edge() {
...
+    let result = model
+        .workbook
+        .worksheet_mut(0)
+        .unwrap()
+        .set_column_width(16, -1.0);
+    assert!(result.is_err());
```
nhatcher: Could you do something like:

```rust
assert_eq!(result, Err("some error message".to_string());
```

The reason is hat sometimes we assume it is an error but there is actually a different one

webapp/src/components/worksheet.tsx
```diff
@@ -104,10 +104,12 @@ function Worksheet(props: {
         editor: editor,
       },
       onColumnWidthChanges(sheet, column, width) {
+        if (width < 0) return;
```
nhatcher: Could you use:
```js
if (width < 0) {
  return;
}
```

I will add this rule to biome some day :S

nhatcher (changes requested):
Excellent job.

I let a couple of minor comments

enjoythecode:
Thanks, addressed the comments.

nhatcher:
Build is failing because of a formatting issue (just run `cargo fmt`)

nhatcher (approved):
LGTM!

PR #176, opened 2024-12-10
Add GEOMEAN Function

berjonah:
Add GEOMEAN Function and tests.


nhatcher:
Awesome work! Thank you!

There are a few missing pieces:

You need to add a test file (created by Excel) in:

It should cover the happy paths and some of the errors. See, for instance the function ABS.xlsx note how cells with a formula are formatted differently

You also need to update the CHANGELOG.md and the documentation functions. See:

https://github.com/ironcalc/IronCalc/pull/164/files

Were something similar is done. Also remember to run `cargo fmt` before pushing to make sure we pass the tests.

Let me know if you have any questions

nhatcher (changes requested):
The heavy lifting of the testing should be done by a file `GEOMEAN.xlsx` covering all corner cases.

nhatcher (approved):
LGTM!

Great contribution, many thanks!

nhatcher:
Hi @berjonah, have a look at my PR in #184 . Regrettably I disabled the tests by mistake.

PR #179, opened 2024-12-12
FIX: Makes styling in Format Menu consistent with Figmas

dg-ac:
Hi again,

I've also made some cosmetic adjustments in the Format Menu, so it looks as expected. See the difference below:

Before:
image

After:
image

Best,
D

### DIFF
--- a/webapp/src/components/formatMenu.tsx
+++ b/webapp/src/components/formatMenu.tsx
@@ -38,7 +38,22 @@ const FormatMenu = (properties: FormatMenuProps) => {
         open={isMenuOpen}
         onClose={(): void => setMenuOpen(false)}
         anchorEl={anchorElement.current}
-        anchorOrigin={properties.anchorOrigin}
+        anchorOrigin={{
+          vertical: "bottom",
+          horizontal: "left",
+        }}
+        transformOrigin={{
+          vertical: "top",
+          horizontal: "left",
+        }}
+        sx={{
+          "& .MuiPaper-root": {
+            borderRadius: "8px",
+            padding: "4px 0px",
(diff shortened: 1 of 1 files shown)


webapp/src/components/formatMenu.tsx
```diff
@@ -38,7 +38,22 @@ const FormatMenu = (properties: FormatMenuProps) => {
...
+        anchorOrigin={{
+          vertical: "bottom",
+          horizontal: "left",
+        }}
+        transformOrigin={{
+          vertical: "top",
+          horizontal: "left",
+        }}
+        sx={{
+          "& .MuiPaper-root": {
```
nhatcher, line 50: We should try hard NOT to do this. This is an example of a _leaky abstraction_. We shouldn't care about the internals.

Also you could do a `StyledMenu` for it

webapp/src/components/formatMenu.tsx
```diff
@@ -123,14 +138,25 @@ const MenuItemWrapper = styled(MenuItem)`
...
   display: flex;
 `;
 
-const MenuDivider = styled("div")``;
+const MenuDivider = styled("div")`
+  width: 100%;
+  margin: auto;
+  margin-top: 4px;
+  margin-bottom: 4px;
+  border-top: 1px solid #eeeeee;
```
nhatcher, line 158: I think we should be consistent on how we write colors. I _try_ to use always uppercase `#EEEEEE`

nhatcher (approved):
LGTM!

webapp/src/components/formatMenu.tsx
dg-ac: Thanks! Fixed in #183, I think.

PR #180, opened 2024-12-12
FIX: Makes footer/navigator scrollable when there are too many sheets

dg-ac:
Hi!

Last PR today. I've added some horizontal scrolling to the spreadsheet navigator. Until now it was impossible to reach all sheets when the screen gets too narrow and/or there are way too many sheets. See below the difference:

Before (sorry, for some reason the video is uploaded as attachment):
https://github.com/user-attachments/assets/b1ab681d-b671-4d73-9cdc-927a75e58d23

After:
https://github.com/user-attachments/assets/df8ef324-de0a-4045-af24-d6ad54a35156

Best,
D

### DIFF
--- a/webapp/src/components/navigation/navigation.tsx
+++ b/webapp/src/components/navigation/navigation.tsx
@@ -2,6 +2,7 @@ import { styled } from "@mui/material";
 import { Menu, Plus } from "lucide-react";
 import { useState } from "react";
 import { useTranslation } from "react-i18next";
+import { theme } from "../../theme";
 import { NAVIGATION_HEIGH } from "../constants";
 import { StyledButton } from "../toolbar";
 import type { WorkbookState } from "../workbookState";
@@ -101,12 +102,19 @@ const Container = styled("div")`
   padding-left: 12px;
   font-family: Inter;
   background-color: #fff;
-  border-top: 1px solid #E0E0E0;
+  border-top: 1px solid #e0e0e0;
 `;
 
 const Sheets = styled("div")`
   flex-grow: 2;
   overflow: hidden;
+  overflow-x: auto;
+
+  &::-webkit-scrollbar {
+    display: none;
+  }
+  -ms-overflow-scroll: none; /* IE and Edge */
+  scrollbar-width: none; /* Firefox */
 `;
 
 const SheetInner = styled("div")`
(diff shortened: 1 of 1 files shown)


webapp/src/components/navigation/navigation.tsx
```diff
@@ -101,23 +102,40 @@ const Container = styled("div")`
...
 
 const Sheets = styled("div")`
   flex-grow: 2;
   overflow: hidden;
+  overflow-x: auto;
+
+  &::-webkit-scrollbar {
+    display: none;
+  }
+  -ms-overflow-scroll: none; /* IE and Edge */
```
nhatcher, line 116: I don't think this is needed anymore

webapp/src/components/navigation/navigation.tsx
```diff
@@ -101,23 +102,40 @@ const Container = styled("div")`
...
-  border-top: 1px solid #E0E0E0;
+  border-top: 1px solid #e0e0e0;
 `;
 
 const Sheets = styled("div")`
   flex-grow: 2;
   overflow: hidden;
+  overflow-x: auto;
+
+  &::-webkit-scrollbar {
```
nhatcher, line 113: I don't think we need this at all.

nhatcher (approved):
Left a couple of comments but LGTM!

webapp/src/components/navigation/navigation.tsx
dg-ac: I'm curious, why? This is there to hide the scrollbar.

nhatcher: In the early 2000's it was typical that browsers behaved differently and many houses (Firefox, IE and Chrome) where having their own prefixes. These days browser vendors are much better at compliance and CSS specific rules are almost always a bad idea.

Think about this: if you put a browser specific CSS rule now all of the sudden we need to do specific testing for several browsers and we might start having specific fixes for each of them. That's a can of worms...

nhatcher: For instance this might just break Firefox, Edge and even Chrome (remember that chrome is not webkit anymore, although I suspect they will keep this particular rule).

Rule of thumb: never use browser specific rules unless you absolutely must, and then you have to figure out the rules for all browsers.

dg-ac: Thanks for the explanation! Super helpful. I'll make a new PR to remove that part.

PR #182, opened 2024-12-12
Or function fixes

### LINKED ISSUE #175: Excel Compatibility: or Function
# Summary
IronCalc allows `=or()` (no parameters), while Excel does not. This means it's possible to export xlsx files from IronCalc that Excel can't load correctly.

## IronCalc Behaviour
- `or` takes any number of parameters, `=or()` returns FALSE.

## Excel Behaviour
- The `or` formula takes a minimum of 1 parameters
- In Excel (365) if you type in `=or()` you get an error and it doesn't accept the formula:  
   image
- If you you load an xlsx containing `=or()` exported from IronCalc, Excel will treat the file as corrupted:   
   image
- If you recover the file, it replaces the formula with the text FALSE
   image

## LibreOffice Calc Behaviour
Like Excel, Calc requires at least one argument, however the UI doesn't prevent you from entering `=or()`. Instead you get an error in the cell `Err511`. If you try to save the file as an XLSX, you get an error:
image

# Underlying Issue
Similar situations can be encountered in other scenarios where Excel disallows an invalid formula, such as setting a cell to `=/`. The core issue is that Excel prevents some formulas from being created in the first place, while IronCalc does not.

# Potential Solutions
Solutions which come to mind for the the broader issue on how to handle invalid formulas that are incompatible with Excel:
1. Follow the Calc approach and disallow saving to XLSX when an incompatible formula is present.
2. Maintain the current export method. Excel will report an error, but the user can "recover" the file with the offending formulas removed and replaced with their calculated values.
3. At XLSX export time, remove incompatible formulas and replace them with calculated values, potentially with a warning message.
4. Implement logic similar to Excel, preventing the creation of these invalid formulas in the first place.

gian-hancock:
Validate argument count for `OR` function and add tests.

Fixes #175

### DIFF
--- a/base/src/functions/logical.rs
+++ b/base/src/functions/logical.rs
@@ -136,6 +136,9 @@ impl Model {
     }
 
     pub(crate) fn fn_or(&mut self, args: &[Node], cell: CellReferenceIndex) -> CalcResult {
+        if args.is_empty() {
+            return CalcResult::new_args_number_error(cell);
+        }
         let mut result = false;
         for arg in args {
             match self.evaluate_node_in_context(arg, cell) {
--- a/base/src/test/mod.rs
+++ b/base/src/test/mod.rs
@@ -46,6 +46,7 @@ pub(crate) mod util;
 
 mod engineering;
 mod test_fn_offset;
+mod test_fn_or;
 mod test_number_format;
 
 mod test_escape_quotes;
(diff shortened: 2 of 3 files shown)


nhatcher (approved):
LGTM!

Many thanks

PR #186, opened 2024-12-13
FIX: Makes styling in Rename Sheet Dialog consistent with rest of dia…

dg-ac:
Hi there,

I think this the last dialog that was still using the old style. See below the before/after:

Before:
image

After:
image

Additionally, I've added something that is not present in other dialogs, as is the ability of applying changes by hitting 'Enter'. Let me know if that also makes sense.

Best,
D

### DIFF
--- a/webapp/src/components/navigation/menus.tsx
+++ b/webapp/src/components/navigation/menus.tsx
@@ -1,17 +1,10 @@
-import {
-  Button,
-  Dialog,
-  DialogActions,
-  DialogContent,
-  DialogTitle,
-  TextField,
-  styled,
-} from "@mui/material";
+import { Dialog, TextField, styled } from "@mui/material";
 import Menu from "@mui/material/Menu";
 import MenuItem from "@mui/material/MenuItem";
 import { Check } from "lucide-react";
 import { useState } from "react";
 import { useTranslation } from "react-i18next";
+import { theme } from "../../theme";
 import type { SheetOptions } from "./types";
 
 function isWhiteColor(color: string): boolean {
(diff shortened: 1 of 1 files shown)


webapp/src/components/navigation/menus.tsx
```diff
@@ -147,4 +172,78 @@ const ItemName = styled("div")`
...
+    color: ${theme.palette.common.black};
+    background-color: ${theme.palette.common.white};
+  }
+  &:hover .MuiInputBase-input {
+    border: 1px solid ${theme.palette.grey["500"]};
+  }
+`;
+
+const DialogFooter = styled("div")`
+  color: #757575;
```
nhatcher, line 223: Theme?

webapp/src/components/navigation/menus.tsx
```diff
@@ -147,4 +172,78 @@ const ItemName = styled("div")`
...
+  border-top: 1px solid ${theme.palette.grey["300"]};
+  font-family: Inter;
+  justify-content: flex-end;
+  padding: 12px;
+`;
+
+const StyledButton = styled("div")`
+  cursor: pointer;
+  color: #ffffff;
+  background: #f2994a;
```
nhatcher, line 235: Theme?

webapp/src/components/navigation/menus.tsx
```diff
@@ -147,4 +172,78 @@ const ItemName = styled("div")`
...
+  padding: 0px 10px;
+  height: 36px;
+  line-height: 36px;
+  border-radius: 4px;
+  display: flex;
+  align-items: center;
+  font-family: "Inter";
+  font-size: 14px;
+  &:hover {
+    background: #d68742;
```
nhatcher, line 245: Theme?

nhatcher (approved):
LGTM!

PR #189, opened 2024-12-15
FIX: Made menu on SheetTab consistent with Figmas

dg-ac:
Hi there,

Changed another menu so all look the same. Find below the change:

Before:
image

After:
image

Best,
D


nhatcher (approved):
LGTM!

PR #195, opened 2024-12-16
FIX: When cut and paste to a different sheet origin is removed

### LINKED ISSUE #193: Webapp: cut across sheets does not remove the original data
ironcalc-cut_across_sheets

Is this expected behavior? I noticed that the code path for this actually does not include a way to communicate the sheet, so maybe it was intended? I don't have a copy of Excel so I can't test. @nhatcher

nhatcher:
Fixes #193

FYI @enjoythecode 

Feel free to review :)


enjoythecode:
I trust you :) LGTM as far as I can tell :)

PR #196, opened 2024-12-16
FIX: Border picker cosmetics

dg-ac:
Hi!

Super late PR 😅 but had to fix one more menu.

I've made the usual padding/margin/spacing/etc changes plus cleaned things up a bit (use theme styles instead of hex colors, replaced an outdated icon and reordered somethings so we don't have buttons inside buttons). See the difference below, as usual:

Before:
image

After:
image

Cheers,
D


nhatcher (approved):
LGTM!

PR #199, opened 2024-12-17
Make XOR, OR, AND functions more consistent with Excel

gian-hancock:
# Description

The way the `XOR`, `OR`, and `AND` functions interpret their arguments is inconsistent with Excel in a few ways:

- **EmptyCell & Text:** Excel ignores arguments evaluating to these types of values, treating them as if they didn't exist.

- **Error if no args:** Excel returns a #VALUE! error for these functions if no arguments are provided, or if all arguments are ignored (see above).

- **EmptyArg:** Bizarrely, Unlike EmptyCell, EmptyArg is not ignored and is treated as if it were FALSE by Excel.

- **ErrorPropagation:** Excel propagates errors in the arguments and in cells belonging to any Range arguments.

Additionally, these functions are not consistent with each other, XOR, OR, AND vary in how they handle the cases mentioned above.

Rectify these consistency issues by re-implementing them all in terms of a single base function which is more consistent with Excel behavior.

# Tests
I wasn't sure if unit tests or the .xslx based tests would be preferred, let me know what you think.

# Further Work
I think there are still some differences when it comes to nested ranges, for example:
```
A1: =OR(B1:B10)
B1: =C1:C10
```
However I think this relates to differences between ranges in Excel and IronCalc in general so I haven't tackled that yet.


base/src/test/test_fn_or_xor.rs
```diff
@@ -0,0 +1,204 @@
...
+        // range with error
+        model._set("B20", "=1/0");
+        model._set("A20", &format!(r#"={func}(B20:C20)"#));
+
+        model.evaluate();
+
+        assert_eq!(model._get_text("A1"), *"#VALUE!");
+        assert_eq!(model._get_text("A2"), *"#VALUE!");
+        assert_eq!(model._get_text("A3"), *"TRUE");
+        assert_eq!(model._get_text("A4"), *"FALSE");
```
gian-hancock, line 135: A1 through A4 used to return TRUE. See "EmptyCell & Text" in PR description.

base/src/test/test_fn_or_xor.rs
```diff
@@ -0,0 +1,204 @@
...
+        assert_eq!(model._get_text("A5"), *"TRUE");
+        assert_eq!(model._get_text("A6"), *"FALSE");
+        assert_eq!(model._get_text("A7"), *"TRUE");
+
+        assert_eq!(model._get_text("A8"), *"TRUE");
+
+        assert_eq!(model._get_text("A9"), *"#VALUE!");
+        assert_eq!(model._get_text("A10"), *"#VALUE!");
+
+        assert_eq!(model._get_text("A11"), *"#VALUE!");
```
gian-hancock, line 146: A9 through A11 used to return FALSE. See "Error if no args" in PR description

base/src/test/test_fn_or_xor.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        // TODO: This one depends on spill behaviour which isn't implemented yet
+        // assert_eq!(model._get_text("A12"), *"TRUE");
+
+        assert_eq!(model._get_text("A13"), *"#VALUE!");
+        assert_eq!(model._get_text("A14"), *"FALSE");
```
gian-hancock, line 152: A13 through A14 used to return TRUE. See "EmptyCell & Text" in PR description.

base/src/test/test_fn_or_xor.rs
```diff
@@ -0,0 +1,204 @@
...
+        assert_eq!(model._get_text("A15"), *"TRUE");
+
+        // TODO: This one depends on @ implicit intersection behaviour which isn't implemented yet
+        // assert_eq!(model._get_text("A16"), *"#VALUE!");
+
+        assert_eq!(model._get_text("A17"), *"TRUE");
+
+        assert_eq!(model._get_text("A18"), *"#VALUE!");
```
gian-hancock, line 160: A18 used to return TRUE. See "EmptyCell & Text" in PR description.

base/src/test/test_fn_or_xor.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        assert_eq!(model._get_text("A19"), *"TRUE");
+
+        assert_eq!(model._get_text("A20"), *"#DIV/0!");
```
gian-hancock, line 164: A20 used to return #VALUE. See "ErrorPropagation" in PR descriptin

base/src/test/test_fn_or_xor.rs
```diff
@@ -0,0 +1,204 @@
...
+    inner("or");
+    inner("xor");
+
+    fn inner(func: &str) {
+        println!("Testing function: {func}");
+
+        let mut model = new_empty_model();
+        model._set("A1", &format!("={}()", func));
+        model.evaluate();
+        assert_eq!(model._get_text("A1"), *"#ERROR!");
```
gian-hancock, line 179: Used to return #VALUE. Excel disallows creating this formula (`=OR()`), IronCalc should return #ERROR! in these cases.

base/src/test/test_fn_or_xor.rs
```diff
@@ -0,0 +1,204 @@
...
+        model._set("A2", &format!("={func}(,1)"));
+        model._set("A3", &format!("={func}(1,)"));
+        model._set("A4", &format!("={func}(,B1)"));
+        model._set("A5", &format!("={func}(,B1:B4)"));
+        model.evaluate();
+        assert_eq!(model._get_text("A1"), *"FALSE");
+        assert_eq!(model._get_text("A2"), *"TRUE");
+        assert_eq!(model._get_text("A3"), *"TRUE");
+        assert_eq!(model._get_text("A4"), *"FALSE");
+        assert_eq!(model._get_text("A5"), *"FALSE");
```
gian-hancock, line 202: A1, A4, A5 used to return #VALUE!. See "EmptyArg" in PR Description

gian-hancock (commented):
Added some comments in the tests indicating the changes in behaviour

base/src/test/test_fn_or_xor.rs
```diff
@@ -0,0 +1,204 @@
...
+use crate::test::util::new_empty_model;
+
+// These tessts are grouped because in many cases XOR and OR have similar behaviour.
+
+// Test specific to xor
+#[test]
+fn fn_xor() {
+    let mut model = new_empty_model();
+
+    model._set("A1", r#"=XOR(1, 1, 1, 0, 0)"#);
```
nhatcher: It is ok to have those here, but you need to have them also in an xlsx file created by Excel. That way we are sure that is the desired behaviour.

You don't need the `r#"` that is used in rust when you have a `"` in the string and don't want to escape it.

nhatcher (changes requested):
Hi @gian-hancock , I haven't finished my review but one thing that would be nice is to add some Excel tests workbooks. Just create files with all the examples you have in here and place that file in:

The testing system will pick it up and ensure we return the same value as Excel does.

base/src/test/test_fn_or_xor.rs
gian-hancock: Will make those changes when I get a chance

gian-hancock:
> Hi @gian-hancock , I haven't finished my review but one thing that would be nice is to add some Excel tests workbooks. Just create files with all the examples you have in here and place that file in:

I've added Excel workbook tests, I added an additional sheet to logical.xlsx.


> You don't need the r#" that is used in rust when you have a " in the string and don't want to escape it.

I removed the unnecessary raw string literalls.

---

I also added some test cases I hadn't thought of and fixed another edge case.

nhatcher:
Hi @gian-hancock , thanks for your patience. I will try to get this review today.

base/src/functions/logical.rs
```diff
@@ -66,91 +66,62 @@ impl Model {
...
+            acc.unwrap_or(true) && value
         }
-        CalcResult::Boolean(true)
+        self.fn_logical_nary(args, cell, fold_fn, Some(false))
     }
 
     pub(crate) fn fn_or(&mut self, args: &[Node], cell: CellReferenceIndex) -> CalcResult {
+        fn fold_fn(acc: Option<bool>, value: bool) -> bool {
+            let r = acc.unwrap_or(false) || value;
+            println!("acc: {:?}, value: {:?}, r: {:?}", acc, value, r);
```
nhatcher: Leftover debug statement? (I'll add a lint rule so this doesn't happen)

base/src/functions/logical.rs
```diff
@@ -166,94 +137,58 @@ impl Model {
...
-                                    }
-                                }
-                                CalcResult::String(_value) => {
-                                    return CalcResult::Boolean(true);
+                                    result = Some(fold_fn(result, value != 0.0))
                                 }
                                 error @ CalcResult::Error { .. } => return error,
-                                CalcResult::Range { .. } => {}
-                                CalcResult::EmptyCell | CalcResult::EmptyArg => {}
+                                CalcResult::EmptyArg => debug_assert!(false, "unreachable"),
```
nhatcher: This is not a bad idea at all, but this wold be the only place where we have a debug_assert! in the code.
Could you remove it and add a comment with the word "unreachable"?

base/src/functions/logical.rs
nhatcher: Done, could you rebase the main branch? (the are no conflicts)

xlsx/tests/calc_tests/logical.xlsx
nhatcher, line 1: Sorry to do this, but could you create a new file instead of modifying this one?

It will be easier to review and to look back for tests in the future

base/src/functions/logical.rs
```diff
@@ -66,91 +66,62 @@ impl Model {
...
+    /// - Empty cell & String: Ignored, behaves exactly like the argument wasn't passed in at all
+    /// - Error: Propagated
+    ///
+    /// If no arguments are provided, or all arguments are ignored, the function returns a #VALUE! error
+    ///
+    /// **`fold_fn`:** The function that combines the running result with the next value boolean value. The running result
+    /// starts as `None`.
+    ///
+    /// **`short_circuit_value`:** If the running result reaches `short_circuit_value`, the function returns early.
+    fn fn_logical_nary(
```
nhatcher: Could you call this method `logical_nary` instead? Function names like `fn_or` means there is an Excel function called `OR`. Nothing relies on this but I'd like to keep that semantics if possible.

base/src/functions/logical.rs
```diff
@@ -66,91 +66,62 @@ impl Model {
...
+        }
+        self.fn_logical_nary(args, cell, fold_fn, Some(true))
+    }
+
+    pub(crate) fn fn_xor(&mut self, args: &[Node], cell: CellReferenceIndex) -> CalcResult {
+        fn fold_fn(acc: Option<bool>, value: bool) -> bool {
+            acc.unwrap_or(false) ^ value
+        }
+        self.fn_logical_nary(args, cell, fold_fn, None)
+    }
```
nhatcher, line 88: You could do something like:

```rust
pub(crate) fn fn_and(&mut self, args: &[Node], cell: CellReferenceIndex) -> CalcResult {
    self.fn_logical_nary(args, cell,  |acc, r| acc.unwrap_or(true) && r, Some(false))
}

pub(crate) fn fn_or(&mut self, args: &[Node], cell: CellReferenceIndex) -> CalcResult {
    self.fn_logical_nary(args, cell, |acc, r| acc.unwrap_or(false) || r , Some(true))
}

pub(crate) fn fn_xor(&mut self, args: &[Node], cell: CellReferenceIndex) -> CalcResult {
    self.fn_logical_nary(args, cell, |acc, r| acc.unwrap_or(false) ^ r, None)
}
```

nhatcher (changes requested):
Actually was very easy to review. I only have a couple of minor comments.

Thanks for taking the time to do this, and sorry for the delayed review

base/src/functions/logical.rs
gian-hancock: I can do that, another option is to use the `unreachable!` macro. I didn't use that because it panics. What are your thoughts on `unreachable!()` vs just a comment?

nhatcher: Yeah, no we can't panic. If there are more `unreachable!` in the codebase is a bug.
Let's just have a comment for the time being. I will most likely convert all those do debug_assert or something similar one day :)

nhatcher:
Ah! you can just add `#[allow(clippy::print_stdout)]` in the test file. Otherwise LGTM!

gian-hancock:
Alright, think that's everything now!

nhatcher (approved):
LGTM!

nhatcher:
(merged)

Thanks!

PR #200, opened 2024-12-17
FIX[WebApp]: Disable delete for sheet if it is the last one

enjoythecode:
Fixes !192

### DIFF
--- a/webapp/src/components/SheetTabBar/SheetTab.tsx
+++ b/webapp/src/components/SheetTabBar/SheetTab.tsx
@@ -1,4 +1,4 @@
-import { Button, Menu, MenuItem, styled } from "@mui/material";
+import { Button, Menu, MenuItem, MenuItemProps, styled } from "@mui/material";
 import { ChevronDown } from "lucide-react";
 import { useRef, useState } from "react";
 import { theme } from "../../theme";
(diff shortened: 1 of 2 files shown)


enjoythecode:
Read for your review @nhatcher

webapp/src/components/SheetTabBar/SheetTab.tsx
```diff
@@ -14,6 +15,7 @@ interface SheetTabProps {
   onSelected: () => void;
   onColorChanged: (hex: string) => void;
   onRenamed: (name: string) => void;
+  canDelete: () => boolean;
```
nhatcher, line 18: I _think_ this doesn't need to be a function, can just be a boolean.

webapp/src/components/SheetTabBar/SheetTabBar.tsx
```diff
@@ -67,6 +67,9 @@ function SheetTabBar(props: SheetTabBarProps) {
               onRenamed={(name: string): void => {
                 props.onSheetRenamed(name);
               }}
+              canDelete={(): boolean => {
```
nhatcher, line 70: If you just pass a boolean this would be: `onDelete={sheets.length > 1}`

nhatcher (changes requested):
Hi @enjoythecode , thanks for taking care of this.

One tiny detail if you agree.

nhatcher:
Let's do the properties thing on a separate PR.

PR #201, opened 2024-12-18
Make DATE function more compliant with excel

yiblet:
### Motivation

While running IronCalc on a complicated Excel model, I noticed that some users rely on Excel’s `DATE` function to handle out-of-range day and month values as a feature rather than an error. For instance, they use `=DATE(2002,1,31+30)` to calculate the date 30 days after January 31. Initially, this seemed like an edge case, and IronCalc didn’t support it because it felt unnecessary. However, it appears this behavior is widely utilized in practice, and improving compatibility here will make IronCalc more versatile and reliable for real-world use cases.

---

### Changes

This PR refactors the date handling logic to improve compatibility with Excel’s `DATE` function, particularly around its permissive handling of months and days that go beyond normal bounds. For example, Excel allows formulas like `=DATE(2002,1,31+30)` to calculate the date 30 days after January 31, 2002. Previously, IronCalc didn’t support this behavior, as it was deemed unnecessary, but it appears to be widely used in practice.

To address this, I introduced a new function, `permissive_date_to_serial_number`, that mimics Excel’s handling of out-of-range months and days. This allows for more flexible date parsing while leaving the existing `date_to_serial_number` function intact, as it might still be needed in less permissive contexts.

The net result is that IronCalc now behaves more like Excel’s `DATE` function in many cases. However, this PR does not address all known compatibility issues. See below for details.

---

### Known Incompatibilities

While this PR improves compatibility, there are some long-standing discrepancies between IronCalc and Excel when it comes to date handling. These issues predate the changes in this PR and remain unresolved:

1. **1900 Leap Year Bug:**  
   Excel incorrectly treats 1900 as a leap year, which introduces a phantom date, February 29, 1900, into its date system. In Excel, serial numbers start from 1 on 1900-01-01. However, IronCalc starts from 1 on 1900-01-01 as well but does not include the leap year bug. This results in a mismatch: in Excel, 1900-01-01 is serial number 1, but in IronCalc, it becomes serial number 2. This offset continues until 1900-02-29, after which the serial numbers align between IronCalc and Excel.

2. **Handling of Dates Before 1900:**  
   Excel supports dates before 1900 using a custom scheme that does not involve negative serial numbers. IronCalc does not handle dates prior to 1900, as its date system is limited to positive serial numbers. The exact mechanism Excel uses for pre-1900 dates is unclear but differs significantly from IronCalc’s current implementation.

These issues are not regressions caused by this PR but are included here for transparency and to highlight potential areas for future improvement.


nhatcher:
Hi @yiblet, thanks for the fix! This is indeed a difficult subject. Will try to get this reviewed tomorrow.

yiblet:
Thank you, I have a few more PRs that I wanted to add but wanted to get started with this one to get a sense of how things work. 

Would you be interested in a PR that changes the excel `TEXT` formula to be case-insensitive? 

It turns out that excel is case insensitive in practice so people have been using that in their excel formulas as well. For context, I have a bunch of spreadsheets that someone else made that I want to try to run on IronCalc, I'm not able to alter them since there's tons of them. So I've identified places where IronCalc differs from excel when evaluating and I've been wanting to fix each of them.

base/src/test/test_date_and_time.rs
```diff
@@ -37,12 +37,12 @@ fn test_fn_date_arguments() {
     assert_eq!(model._get_text("A3"), *"#ERROR!");
     assert_eq!(model._get_text("A4"), *"#ERROR!");
 
-    assert_eq!(model._get_text("A5"), *"#NUM!");
-    assert_eq!(model._get_text("A6"), *"#NUM!");
-    assert_eq!(model._get_text("A7"), *"#NUM!");
-    assert_eq!(model._get_text("A8"), *"#NUM!");
+    assert_eq!(model._get_text("A5"), *"10/10/1974");
```
nhatcher, line 40: Excel-ent!. Can you update the comment on lines 21-22 too?

base/src/formatter/dates.rs
```diff
@@ -12,7 +26,117 @@ pub fn from_excel_date(days: i64) -> NaiveDate {
...
 }
+
+pub fn permissive_date_to_serial_number(day: i32, month: i32, year: i32) -> Result<i32, String> {
+    // Excel parses `DATE` very permissively. It allows not just for valid date values, but it
+    // allows for invalid dates as well. If you for example enter `DATE(1900, 1, 32)` it will
+    // return the date `1900-02-01`. Despite giving a day that is out of range it will just
+    // wrap the month and year around.
+    //
+    // This function applies that same logic to dates. And does it in the most compatible way as
+    // possible.
```
nhatcher, line 41: Could you add this comment on the top of the function with three `///` instead? Those are doc comments.
Eventually all public functions need to be documented that way.

base/src/constants.rs
```diff
@@ -16,3 +16,12 @@ pub(crate) const LAST_ROW: i32 = 1_048_576;
 // NaiveDate::from_ymd(1900, 1, 1).num_days_from_ce() - 2
 // The 2 days offset is because of Excel 1900 bug
 pub(crate) const EXCEL_DATE_BASE: i32 = 693_594;
+
+// Excel can handle dates until the year 0000-01-01
+// However, it uses a different numbering scheme for dates
+// that are before 1900-01-01.
+// So for now we will simply not support dates before 1900-01-01.
+pub(crate) const EXCEL_DATE_MIN: i32 = 2;
```
nhatcher, line 24: Could you rename it to `DATE_SERIAL_NUMBER_MINIMUM`? and also the other one.
Note that they have different units from the (also a bit ill-named) `EXCEL_DATE_BASE`

nhatcher (changes requested):
Looks great! Just a couple of minor issues and we are good to go.

nhatcher:
@yiblet 
> Excel supports dates before 1900 using a custom scheme that does not involve negative serial numbers

does it? I mean it accepts the arguments in the DATE function but DATE(1845, 2, 3) and DATE(3745, 2, 3) return the same serial number

nhatcher:
> Would you be interested in a PR that changes the excel TEXT formula to be case-insensitive?

Out of the scope of this PR, but could you give me an example

> I have a bunch of spreadsheets that someone else made that I want to try to run on IronCalc

That is a mine of gold! But I have two caveats. At the moment IronCalc is way behind functions implemented and features in general, implementing Excel's obscure corners shouldn't be our priority right now.
The second is more general. We want to be compatible with Excel, but do we ant to be compatible with Excel when Excel is wrong? A function like DATE(2025, 13, 3) could be wrong because they swapped day and month. Excel doesn't document that and it is debatable whether is the right thing to do or not.

One thing I was thinking is that we could have two different builds, one with strict compatibility and the "IronCalc" build that essentially compatible but more sensible and strict and doesn't follow Excel in corner cases.
Anyhow, all that is tomorrow's Nico's problem

nhatcher (approved):
Hi @yiblet,

Since there is nothing wrong here and my emergency fix #223 will make mergeing this more difficult, I will merge this now.

yiblet:
@nhatcher 

Thank you! 

Re: the case insensitivity on the formatter here's what I noticed in excel:
Screenshot 2024-12-30 at 4 55 39 PM


As you can see excel parses `YYYY-MM-DD` as the same thing as `yyyy-mm-dd` It's case insensitive when it comes to TEXT formatting. So I was thinking of creating a PR for IronCalc to have the same behavior. From what I gather it seems like excel is case insensitive in all strings except string literals. I.E. `"DATE:" YYYY-MM-DD` will output something differently than `"date:" YYYY-MM-DD`. I'll run more tests to see if it applies to the other formatting rules but a quick test shows that it's the case. 


Before that though I'll send a separate patch in for a little unescaped xml bug in the xlsx serialization code.

PR #202, opened 2024-12-18
UPDATE: Replace one preset color from color picker

dg-ac:
Hi there,

Tiny change here. I wanted to add IronCalc's orange (F2994A) to the color presets. The main motivation is to use it in the examples we will be creating for our Docs soon. Besides,

1. we were lacking an orange, and had too many shades of red,
2. and the textfield can't be directly edited at the moment (I think we should fix this soon) so adding the preset seems the quickest way of doing this

Thanks!
D


nhatcher:
> he main motivation is to use it in the examples we will be creating for our Docs soon

Remember that the docs need to be created in Excel

nhatcher (approved):
LGTM!

PR #204, opened 2024-12-19
FIX: make grid header colors consistent

dg-ac:
### **Description**
Column headers in IronCalc were using the wrong color (a shade of blue instead of a shade of grey).

Before:
image

After:
image

### **Motivation**
I know this is super hard to notice for some people and in some displays but as it's a small change I thought it'd be good to get it done now 🙂.

Best,
D


nhatcher (approved):
LGTM!

PR #205, opened 2024-12-19
FIX[WebApp]: Some fixes to the DeleteWorkbook and Import dialogs

nhatcher:
Hi @dg-ac,

Some minor changes hopefully they are self-evident.

### DIFF
--- a/webapp/src/AppComponents/DeleteWorkbookDialog.tsx
+++ b/webapp/src/AppComponents/DeleteWorkbookDialog.tsx
@@ -1,20 +1,23 @@
 import styled from "@emotion/styled";
 import { Trash2 } from "lucide-react";
-import { forwardRef, useEffect } from "react";
-import { theme } from "../theme";
+import { useEffect, useRef } from "react";
 
-export const DeleteWorkbookDialog = forwardRef<
-  HTMLDivElement,
-  {
-    onClose: () => void;
-    onConfirm: () => void;
-    workbookName: string;
-  }
->((properties, ref) => {
+interface DeleteWorkbookDialogProperties {
+  onClose: () => void;
+  onConfirm: () => void;
+  workbookName: string;
+}
+
+function DeleteWorkbookDialog(properties: DeleteWorkbookDialogProperties) {
+  const deleteButtonRef = useRef<HTMLButtonElement>(null);
+
   useEffect(() => {
     const root = document.getElementById("root");
     if (root) {
       root.style.filter = "blur(2px)";
+    }
+    if (deleteButtonRef.current) {
+      deleteButtonRef.current.focus();
     }
     return () => {
       const root = document.getElementById("root");
@@ -26,8 +29,12 @@ export const DeleteWorkbookDialog = forwardRef<
 
   return (
     <DialogWrapper
-      ref={ref}
       tabIndex={-1}
+      onKeyDown={(event) => {
+        if (event.code === "Escape") {
+          properties.onClose();
+        }
+      }}
       role="dialog"
       aria-labelledby="delete-dialog-title"
       aria-describedby="delete-dialog-description"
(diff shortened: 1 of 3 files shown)


webapp/src/AppComponents/DeleteWorkbookDialog.tsx
```diff
@@ -1,21 +1,24 @@
...
-import { theme } from "../theme";
-
-export const DeleteWorkbookDialog = forwardRef<
-  HTMLDivElement,
-  {
-    onClose: () => void;
-    onConfirm: () => void;
-    workbookName: string;
-  }
->((properties, ref) => {
```
nhatcher, line 13: We are using functions in all these things, is there any particular reason you went on a different route?
I might be missing something though.

webapp/src/AppComponents/DeleteWorkbookDialog.tsx
```diff
@@ -1,21 +1,24 @@
...
+  const deleteButtonRef = useRef<HTMLButtonElement>(null);
+
   useEffect(() => {
     const root = document.getElementById("root");
     if (root) {
       root.style.filter = "blur(2px)";
     }
+    if (deleteButtonRef.current) {
+      deleteButtonRef.current.focus();
+    }
```
nhatcher, line 21: This is so that the delete button is focused initially. We need to have something in the Dialog focused so ESC can work.

webapp/src/AppComponents/DeleteWorkbookDialog.tsx
```diff
@@ -55,10 +63,25 @@ export const DeleteWorkbookDialog = forwardRef<
       </ContentWrapper>
     </DialogWrapper>
   );
-});
+}
 
 DeleteWorkbookDialog.displayName = "DeleteWorkbookDialog";
 
+// some colors taken from the IronCalc palette
```
nhatcher, line 70: We can't use the palette. This is part of the app we are building that is built on the top of the IronCalc widget.

Maybe we could export those colors somehow.

webapp/src/AppComponents/DeleteWorkbookDialog.tsx
```diff
@@ -139,22 +162,24 @@ const Button = styled.button`
...
-  color: ${theme.palette.grey["700"]};
+  background-color: ${GREY_200};
+  color: ${GREY_700};
   &:hover {
-    background-color: ${theme.palette.grey["300"]};
+    background-color: ${GREY_300};
   }
 `;
+
+export default DeleteWorkbookDialog;
```
nhatcher, line 185: For things like that, there is only one export, we want to use the default export (so we use `import myDialog from ..` instead of `import { myDialog } from ..`

webapp/src/AppComponents/UploadFileDialog.tsx
```diff
@@ -1,17 +1,34 @@
 import styled from "@emotion/styled";
 import { BookOpen, FileUp } from "lucide-react";
-import { type DragEvent, useRef, useState } from "react";
+import { type DragEvent, useEffect, useRef, useState } from "react";
```
nhatcher, line 3: Since I am here I did the changes for the import Dialog too.

webapp/src/AppComponents/DeleteWorkbookDialog.tsx
dg-ac: Gotcha!

webapp/src/AppComponents/UploadFileDialog.tsx
dg-ac: Makes sense.

dg-ac (approved):
LGTM! Sorry for the extra review and thanks for all the explanations, it helps a lot 🙏

PR #206, opened 2024-12-19
FIX[UI]: Use `canDelete` as a property not a function

nhatcher:
FYI: @enjoythecode


enjoythecode:
Yup, makes sense and LGTM. I didn't have much time this week to code.

nhatcher:
I merged this in #207  :D

PR #207, opened 2024-12-19
UPDATE: Hide/Unhide sheets

nhatcher:
This implements hide/unhide sheets in the UI and bindings

There are some corner cases that are not treated correctly:

1. Hide sheet
2. Unhide sheet and select it
3. Undo

The you end up in a selected sheet that is hidden.

You probably will want to change some stuff later:

image
image


dg-ac (approved):
LGTM!

PR #209, opened 2024-12-20
FIX: Use the right icon for paste format action

dg-ac:
Hi @nhatcher,

We recently updated the 'Copy/Paste Format' icon in the toolbar but we still had the old one as a cursor, after clicking the button. So here's my attempt to fix that. 

In the video below you can see the before/after:

https://github.com/user-attachments/assets/c4656abc-0112-4095-8e98-d068921a78a0

Let me know if this is the right way of doing it, of if there's maybe a 'cleaner' way.

Thanks!

D


nhatcher (approved):
LGTM!

PR #210, opened 2024-12-22
FIX: Replace SVG "X" icons with Lucide icons for consistency in dialogs

dg-ac:
Hi there,

This PR replaces the custom 'X' svg icons with the Lucide icon library in all dialogs.

Best,
D

### DIFF
--- a/webapp/src/AppComponents/UploadFileDialog.tsx
+++ b/webapp/src/AppComponents/UploadFileDialog.tsx
@@ -114,6 +114,7 @@ function UploadFileDialog(properties: {
           style={{ marginRight: 12 }}
           onClick={handleClose}
           title="Close Dialog"
+          ref={crossRef}
           tabIndex={0}
         >
           <X />
--- a/webapp/src/AppComponents/UploadFileDialog.tsx
+++ b/webapp/src/AppComponents/UploadFileDialog.tsx
@@ -1,5 +1,5 @@
 import styled from "@emotion/styled";
-import { BookOpen, FileUp } from "lucide-react";
+import { BookOpen, FileUp, X } from "lucide-react";
 import { type DragEvent, useEffect, useRef, useState } from "react";
 
 function UploadFileDialog(properties: {
@@ -113,30 +113,11 @@ function UploadFileDialog(properties: {
         <Cross
           style={{ marginRight: 12 }}
           onClick={handleClose}
+          title="Close Dialog"
           ref={crossRef}
           tabIndex={0}
         >
-          <svg
-            width="16"
-            height="16"
-            viewBox="0 0 16 16"
-            fill="none"
-            xmlns="http://www.w3.org/2000/svg"
-          >
-            <title>Close</title>
-            <path
-              d="M12 4.5L4 12.5"
-              stroke="#333333"
-              strokeLinecap="round"
-              strokeLinejoin="round"
-            />
-            <path
-              d="M4 4.5L12 12.5"
-              stroke="#333333"
-              strokeLinecap="round"
-              strokeLinejoin="round"
-            />
-          </svg>
+          <X />
         </Cross>
       </UploadTitle>
       {message === "" ? (
@@ -237,6 +218,11 @@ const Cross = styled("div")`
   cursor: pointer;
   align-items: center;
   justify-content: center;
+  svg {
+    width: 16px;
+    height: 16px;
+    stroke-width: 1.5;
+  }
 `;
 
 const DocLink = styled("span")`
--- a/webapp/src/AppComponents/UploadFileDialog.tsx
+++ b/webapp/src/AppComponents/UploadFileDialog.tsx
@@ -114,7 +114,6 @@ function UploadFileDialog(properties: {
           style={{ marginRight: 12 }}
           onClick={handleClose}
           title="Close Dialog"
-          ref={crossRef}
           tabIndex={0}
         >
           <X />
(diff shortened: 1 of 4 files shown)


webapp/src/components/formatPicker.tsx
```diff
@@ -36,28 +37,14 @@ const FormatPicker = (properties: FormatPickerProps) => {
...
-              d="M4 4.5L12 12.5"
-              stroke="#333333"
-              strokeLinecap="round"
-              strokeLinejoin="round"
-            />
-          </svg>
+        <Cross
+          onClick={handleClose}
+          title="Close Dialog"
+          ref={crossRef}
```
nhatcher: Why do you need the ref?

nhatcher (changes requested):
Just one question!

Thanks for taking care of this

webapp/src/components/SheetTabBar/SheetRenameDialog.tsx
```diff
@@ -13,35 +14,22 @@ interface SheetRenameDialogProps {
...
-              strokeLinecap="round"
-              strokeLinejoin="round"
-            />
-          </svg>
+        <Cross
+          onClick={handleClose}
+          title="Close Dialog"
+          ref={crossRef}
+          tabIndex={0}
+          onKeyDown={() => {}}
```
nhatcher, line 28: The `onKeyDown` event is currently not doing anything.

It's extra code I added some time ago. Also, `ref` and `tabIndex` aren’t doing anything right now:  
[MDN Documentation on `tabIndex`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)

In that documentation, you can read about the `-1` value. It’s used so the element can receive keyboard focus. This solves two issues:

1. **Focus in a widget with no other focusable tags**  
   When the dialog opens, it grabs the focus. Without a focusable element, the focus would remain on the workbook. This is useful for capturing the `ESC` keypress event, for example.

2. **Capturing the `Enter` keypress when the X is focused**  
   We can use the `onKeyDown` event to catch the `Enter` keypress when the close button (the X) has focus.

The update dialog behaves incorrectly. If you open it and start pressing the tab key, you’ll see that the focus eventually moves to the workbook. That doesn’t happen with the other dialogs, such as the rename or delete workbook dialogs.

webapp/src/components/formatPicker.tsx
dg-ac: Removed, it was not really needed.

webapp/src/components/SheetTabBar/SheetRenameDialog.tsx
dg-ac: Oh I see. Thanks for the explanation. So it would make sense to also add the tabIndex to the X and 'Save' buttons, in all dialogs, as well?

> The update dialog behaves incorrectly. 

You mean the "Import' dialog?

nhatcher: > You mean the "Import' dialog?

Yes, sorry.

> So it would make sense to also add the tabIndex to the X and 'Save' buttons, in all dialogs, as well?

So elements like inputs and texareas are focusable (I'm not sure if that is a word) and you don't need a tabIndex on them

webapp/src/AppComponents/UploadFileDialog.tsx
```diff
@@ -113,30 +113,10 @@ function UploadFileDialog(properties: {
         <Cross
           style={{ marginRight: 12 }}
           onClick={handleClose}
-          ref={crossRef}
```
nhatcher, line 116: This is an error you *need* ref in this case, because in this case something needs to have the focus and there is no input in the dialog.

We could also have the focus on the UploadTitle, but something needs to have the focus

webapp/src/components/SheetTabBar/SheetRenameDialog.tsx
nhatcher: You don't need tabIndex and onKeydown here unless you are planning on using them.

One neat use of it would be to exit when pressing enter while the X is focused

webapp/src/components/formatPicker.tsx
```diff
@@ -36,28 +36,13 @@ const FormatPicker = (properties: FormatPickerProps) => {
...
-              stroke="#333333"
-              strokeLinecap="round"
-              strokeLinejoin="round"
-            />
-          </svg>
+        <Cross
+          onClick={handleClose}
+          title="Close Dialog"
+          tabIndex={0}
+          onKeyDown={() => {}}
```
nhatcher: * tabIndex and onKeyDown are not used here.
* Inside the widget we shouldn't use text as in "Close Dialog", you should add the string here:
https://github.com/ironcalc/IronCalc/blob/main/webapp/src/locale/en_us.json
And then use the translation tool. See how it is dine for the "save" button.

dg-ac: I've removed it for now. I'll work on this next.

webapp/src/AppComponents/UploadFileDialog.tsx
dg-ac: Added back, sorry!

webapp/src/components/SheetTabBar/SheetRenameDialog.tsx
dg-ac: Should be fixed now.

nhatcher: You still don't need the "onKeyDown"

Let's tackle this and other issues in a new PR.

nhatcher (approved):
LGTM!

PR #211, opened 2024-12-23
Add pv description and examples

stevethesleeve:
Includes minor typo fix typo for one of the links in the FV function description.

PR #213, opened 2024-12-27
Include PV function description

nhatcher:
Hi @stevethesleeve , this is PR #211 without the extra stuff from FV


nhatcher:
Hi @stevethesleeve, the problem was that IronCalc shows the values given by Excel on loading. You need to do one computation to realize there is a bug.

For instanace introduce the number 1 in any empty cell.

PR #215, opened 2024-12-28
FIX: Increments the dum counter in the number of documented functions

nhatcher:


nhatcher:
> dum counter
dumb...

PR #220, opened 2024-12-29
Name Manager frontend

aloifran:
#### This PR adds the Name Manager Modal with essential functionality such as 'create, edit, delete' named ranges.

There some todo notes for missing things, such as:
- validations and error messages
- logic that should be leveraged to the modal instead of its child components
- minor style adjustments


nhatcher:
Hi @aloifran , could you add a description? You can link to the Figma and include a couple of piuctures.
Also mention the things that are not yet done.

webapp/src/components/util.ts
```diff
@@ -57,7 +61,24 @@ export function rangeToStr(
...
-    columnStart,
-  )}${rowStart}:${columnNameFromNumber(columnEnd)}${rowEnd}`;
+  return `${sheetName}${columnNameFromNumber(columnStart)}${rowStart}:${columnNameFromNumber(
+    columnEnd,
+  )}${rowEnd}`;
+}
+
+export function getFullRangeToString(
+  selectedView: SelectedView,
+  worksheets: WorksheetProperties[], // solo pasar names
```
nhatcher: Left over comment?

webapp/src/components/util.ts
```diff
@@ -57,7 +61,24 @@ export function rangeToStr(
...
+): string {
+  // order of values is confusing compared to rangeToStr range type, needs refactoring for consistency
```
nhatcher, line 69: Could we do this in this PR?

I realize the rest of the code doesn't have them yet, but could you add a unit test for this function?

webapp/src/locale/en_us.json
```diff
@@ -58,14 +59,12 @@
   "num_fmt": {
     "title": "Custom number format",
     "label": "Number format",
-    "close": "Close dialog",
```
nhatcher: Why this change?

webapp/src/locale/en_us.json
```diff
@@ -58,14 +59,12 @@
...
     "save": "Save"
   },
   "sheet_rename": {
     "rename": "Save",
     "label": "New name",
-    "title": "Rename Sheet",
-    "close": "Close dialog"
```
nhatcher: Those two are needed? If they are not could you place them in a separate PR?

webapp/src/components/toolbar.tsx
```diff
@@ -34,6 +36,7 @@ import {
   DecimalPlacesIncreaseIcon,
 } from "../icons";
 import { theme } from "../theme";
+import NameManagerDialog from "./NameManager/NameManagerDialog";
```
nhatcher: I am not sure but I _think_ the path should be `import NameManagerDialog from "./NameManagerDialog/NameManagerDialog";`

If you add an index.tsx with `export { default } from "./NameManagerDialog";`
Then you can just:

`import NameManagerDialog from "./NameManagerDialog`

webapp/src/components/toolbar.tsx
```diff
@@ -60,6 +63,7 @@ type ToolbarProperties = {
   onFillColorPicked: (hex: string) => void;
   onNumberFormatPicked: (numberFmt: string) => void;
   onBorderChanged: (border: BorderOptions) => void;
+  onNamedRangesUpdate: () => void;
```
nhatcher: This is not used, should it be passed on from the workbook?

nhatcher:
I can't add a name if the list needs scrolling:

image

nhatcher:
When you update a name, add a name or delete one the changes should be applied immediately to the workbook.
You just need to "emit" a "onChange" event and run a redraw request in the workbook. You can see how the borderPicker does it.

webapp/src/components/NameManager/NameManagerDialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+    <StyledDialog open={open} onClose={onClose} maxWidth={false} scroll="paper">
+      <StyledDialogTitle>
+        {t("name_manager_dialog.title")}
+        <IconButton onClick={() => onClose()}>
+          <X size={16} />
+        </IconButton>
+      </StyledDialogTitle>
+      <StyledDialogContent dividers>
+        <StyledRangesHeader>
+          <Box width="171px">{t("name_manager_dialog.name")}</Box>
```
nhatcher: This should be a StyledBox then

webapp/src/components/NameManager/NameManagerDialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+          <Box width="171px">{t("name_manager_dialog.range")}</Box>
+          <Box width="171px">{t("name_manager_dialog.scope")}</Box>
+        </StyledRangesHeader>
+        {definedNamesLocal?.map((definedName) => (
```
nhatcher: `definedBamesLocal` is always defined. You don't need the question mark

nhatcher (changes requested):
Changes so far

webapp/src/components/toolbar.tsx
aloifran: Thanks, done

webapp/src/components/NameManager/NameManagerDialog.tsx
aloifran: Done

webapp/src/components/NameManager/NameManagerDialog.tsx
aloifran: Done

webapp/src/components/toolbar.tsx
aloifran: Done

webapp/src/components/util.ts
aloifran: I've added a unit test and refactored params to use only worksheetNames instead of worksheets.
Regarding my comment in line 73, this should be done later on as many files could be affected

webapp/src/locale/en_us.json
aloifran: To remove text that is not needed after design changes.

webapp/src/locale/en_us.json
aloifran: I am removing them since they're not needed after design changes.

nhatcher: Could we put them in a separate PR? I rather have this PR focused on the name manager

webapp/src/locale/en_us.json
nhatcher: Could we put it on a separate PR? I rather have this PR focused on the name manager

webapp/src/components/NameManager/NamedRange.tsx
```diff
@@ -0,0 +1,219 @@
...
+              <Check size={12} />
+            </IconButton>
+            <IconButton
+              onClick={handleCancel}
+              sx={{ color: (theme) => theme.palette.error.main }}
+            >
+              <X size={12} />
+            </IconButton>
+          </>
+        ) : (
```
nhatcher: This is too much logic in a single function. You can split this two methods for active and non active ranges

webapp/src/locale/en_us.json
aloifran: Actually, I noticed this changes are not related to name manager. I'll revert them

webapp/src/locale/en_us.json
aloifran: Actually, I noticed this changes are not related to name manager. I'll revert them

webapp/src/components/NameManager/NamedRangeActive.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const [newFormula, setNewFormula] = useState(formula);
+
+  // todo: add error messages for validations
+  const [nameError, setNameError] = useState(false);
+  const [formulaError, setFormulaError] = useState(false);
+
+  const handleSaveUpdate = () => {
+    const definedNamesModel = model.getDefinedNameList();
+
+    if (definedNamesModel.find((n) => n.name === name)) {
```
nhatcher: You need name and scope

webapp/src/components/NameManager/NamedRange.tsx
aloifran: Splitted it!

webapp/src/components/NameManager/NamedRangeActive.tsx
aloifran: Applied, thanks

nhatcher (approved):
Ok @aloifran , extraordinary work at your first complex dialog.
Many thanks for the hard work!

LGTM!

PR #225, opened 2024-12-31
fix: escape string formulas in worksheet xml export

yiblet:
Right now, if IronCalc tries to serialize the value of the cell `="x > y"` it would generate invalid xml since the `>` isn't escaped. Here's a tiny one-line fix for that along with an accompanying test in `calc_test`. 

I'm not sure if the accompanying test in in the right place or if there's a different way you'd like me to add a test for this. I'm happy to change that as needed.


nhatcher (approved):
LGTM!

PR #233, opened 2025-01-04
UPDATE[docs]: Compatibility and other engines

nhatcher:
Hi @stevethesleeve,  this draft PR contains some proposals.

First I think we need an entry in numbers and floats as people will find surprising that 0.2+0.1 is not 0.3

I think we should add section in many (all?) functions called "Compatibility Notes" or something along those lines saying where we diverge from Excel.

Maybe we can also talk about other spreadsheet engines. Maybe not in every function, but in those places where there is something interesting to say.

Note that the pages and comments in this PR can only be consider placeholders.

### DIFF
--- a/docs/src/.vitepress/config.mts
+++ b/docs/src/.vitepress/config.mts
@@ -75,8 +75,12 @@ export default defineConfig({
             link: "/features/units",
           },		  
           {
-            text: "Serial Numbers",
+            text: "Dates and serial numbers",
             link: "/features/serial-numbers",
+          },
+          {
+            text: "Numbers in IronCalc",
+            link: "/features/numbers-in-ironcalc",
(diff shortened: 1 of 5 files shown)


stevethesleeve:
Hi @nhatcher 

With regards to your 0.2+0.1 $\ne$ 3 point, we've already made reference to this behaviour in https://docs.ironcalc.com/features/value-types.html#numbers. Where else do you think would be appropriate?

I'm happy with the idea of adding a "Compatibility Notes" section which we can fill in with information that we have to hand. Probably best to keep the same content headings on each page, and write "None" in this section when we have nothing to say. It makes good sense to include Google Sheets compatibility issues in these sections. Whether we want to extend to less-used engines is debatable.

PR #234, opened 2025-01-06
UPDATE: Dialog fixes

dg-ac:
Hey there,

This PR intends to fix some issues and inconsistencies in our dialogs. Before, hitting the key Tab on dialogs would also select the content behind. That should be fixed now. 

Best,
D

### DIFF
--- a/webapp/src/AppComponents/UploadFileDialog.tsx
+++ b/webapp/src/AppComponents/UploadFileDialog.tsx
@@ -1,4 +1,4 @@
-import styled from "@emotion/styled";
+import { Dialog, styled } from "@mui/material";
 import { BookOpen, FileUp, X } from "lucide-react";
 import { type DragEvent, useEffect, useRef, useState } from "react";
 
@@ -16,7 +16,7 @@ function UploadFileDialog(properties: {
   useEffect(() => {
     const root = document.getElementById("root");
     if (root) {
-      root.style.filter = "blur(2px)";
+      root.style.filter = "none";
     }
     if (crossRef.current) {
       crossRef.current.focus();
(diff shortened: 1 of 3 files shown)


webapp/src/components/formatPicker.tsx
```diff
@@ -36,7 +36,12 @@ const FormatPicker = (properties: FormatPickerProps) => {
     >
       <StyledDialogTitle>
         {t("num_fmt.title")}
-        <Cross onClick={handleClose} title={t("num_fmt.close")}>
+        <Cross
+          onClick={handleClose}
+          title={t("num_fmt.close")}
+          tabIndex={0}
+          onKeyDown={(event) => event.key === "Enter" && properties.onClose()}
```
nhatcher: Please use the longer form:

```js
if (event.key === "Enter") {
   properties.onClose();
}
``

File: webapp/src/AppComponents/UploadFileDialog.tsx [hunk 2]
```diff
@@ -16,7 +16,7 @@ function UploadFileDialog(properties: {
   useEffect(() => {
     const root = document.getElementById("root");
     if (root) {
-      root.style.filter = "blur(2px)";
+      root.style.filter = "none";
```
nhatcher: Do we even need this now? we can probably completely remove it

nhatcher (changes requested):
Two small questions

webapp/src/AppComponents/UploadFileDialog.tsx
dg-ac: You're right, we don't need it. Removed in last commit.

webapp/src/components/formatPicker.tsx
dg-ac: Changed.

nhatcher (approved):
LGTM!

PR #236, opened 2025-01-09
Make SUM and PRODUCT functions more consistent with Excel

gian-hancock:
In Excel the SUM and PRODUCT functions convert non numeric arguments. This PR updates IronCalc to do the same.

Notably, Excel **does not** convert values which are part a range or reference argument, this PR adds tests for this oddity (among other things).


nhatcher:
WHi @gian-hancock , thanks for taking care of this. Will review soon.

nhatcher:
Merged! Thank you @gian-hancock !

PR #242, opened 2025-01-12
FIX: Name Manager cosmetics and responsiveness

dg-ac:
Hi there,

I've been making a few cosmetic adjustments to the new Name Manager. What we have is great as a first version and works quite well, however I thought some design adjustments were needed. Besides, now it can be used on mobile devices. 

So, in summary, I have:

1. Changed a bunch of borders, margins, paddings, etc to improve the layout
2. Improved accessibility on action buttons by increasing icon sizes, adding hover style and tooltips
3. Adjusted header and footer to make it consistent with the other dialogs
4. Adjusted the dialog so it can be used on mobile devices

Before:
image

After:
image

Regarding the mobile view, it still needs to improve a lot, but at least is usable. Next step here would be to get back to Figma and design a proper mobile version of this dialog.

I would also like to work on focus styles, and maybe add new features.

Thanks,
D

### DIFF
--- a/webapp/src/components/NameManagerDialog/NameManagerDialog.tsx
+++ b/webapp/src/components/NameManagerDialog/NameManagerDialog.tsx
@@ -196,27 +196,31 @@ const StyledDialogTitle = styled(DialogTitle)`
 
 const NameListWrapper = styled(Stack)`
   overflow-y: auto;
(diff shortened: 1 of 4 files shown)


nhatcher (approved):
LGTM!

PR #243, opened 2025-01-15
FIX: Focus and button syling in name manager

dg-ac:
Hey there,

I made a couple of extra small adjustments to the name manager:

- Keyboard focus works now as expected,
- and buttons styling (added the state `focusVisible` and changed `disabled` a bit)

Thanks,
D

### DIFF
--- a/webapp/src/components/NameManagerDialog/NameManagerDialog.tsx
+++ b/webapp/src/components/NameManagerDialog/NameManagerDialog.tsx
@@ -84,7 +84,7 @@ function NameManagerDialog(properties: NameManagerDialogProperties) {
           <StyledBox>{t("name_manager_dialog.range")}</StyledBox>
           <StyledBox>{t("name_manager_dialog.scope")}</StyledBox>
         </StyledRangesHeader>
-        <NameListWrapper>
+        <NameListWrapper tabIndex={-1}>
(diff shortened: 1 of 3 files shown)


webapp/src/components/NameManagerDialog/NameManagerDialog.tsx
```diff
@@ -84,7 +84,7 @@ function NameManagerDialog(properties: NameManagerDialogProperties) {
           <StyledBox>{t("name_manager_dialog.range")}</StyledBox>
           <StyledBox>{t("name_manager_dialog.scope")}</StyledBox>
         </StyledRangesHeader>
-        <NameListWrapper>
+        <NameListWrapper tabIndex={-1}>
```
nhatcher, line 87: Could you explain why are you doing this?

webapp/src/components/NameManagerDialog/NamedRangeActive.tsx
```diff
@@ -92,12 +92,14 @@ function NamedRangeActive(properties: NamedRangeProperties) {
               }
             }}
             title={t("name_manager_dialog.apply")}
+            disableFocusRipple={true}
```
nhatcher, line 95: What does this do?

PR #244, opened 2025-01-18
Add ERFC, ERF.PRECISE and ERFC.PRECISE functions

stevethesleeve:


nhatcher:
Hi Steve, I am not big on the "WolframAlpha Value" Could we call it "Exact value"? Or even remove it completely?
I am not sure if that adds value. Maybe it does

docs/src/functions/engineering/erf-precise.md
```diff
@@ -3,9 +3,50 @@ layout: doc
 outline: deep
 lang: en-US
 ---
+# ERF.PRECISE function
```
nhatcher: I think ERF.PRECISE should point to the same page as ERF and add a note that ERF. PRECISE is there only for comptibility reasons.

docs/src/functions/engineering/erf-precise.md
```diff
@@ -3,9 +3,50 @@ layout: doc
...
+<center><img src="/functions/images/error-function-curve.png" width="350" alt="Graph showing erf(x) for x between -3 and +3."></center>
+
+* This figure illustrates some of the key characteristics of the error function:
+
+  * $\text{erf}(0) = 0$
+  * $\text{erf}(x) = -\text{erf}(x)$
+  * As $x \rightarrow \infty$, $\text{erf}(x) \rightarrow 1$
+  * As $x \rightarrow -\infty$, $\text{erf}(x) \rightarrow -1$
+
+* The error function is a [transcendental](https://en.wikipedia.org/wiki/Transcendental_function), non-algebraic mathematical function. IronCalc implements the ERF.PRECISE function by numerical approximation using a power series.
```
nhatcher: The precise way in which a transcendental function is implemented on IronCalc might change in the future, I would remove ~using power series~.

docs/src/functions/engineering/erfc-precise.md
```diff
@@ -3,9 +3,50 @@ layout: doc
...
+## Overview
+ERFC.PRECISE (<u>ER</u>ror <u>F</u>unction <u>C</u>omplementary) is a function of the Engineering category that calculates a value for the _complementary error function_, defined by $\text{erfc}(x) = 1 - \text{erf}(x)$. Also known as the _complementary Gauss error function_, the complementary error function represents the probability of a random variable falling outside a certain range, given that it follows a specified normal distribution.
 
-# ERFC.PRECISE
+ERFC.PRECISE is provided for compatibility with other spreadsheets. For all real values of $x$, $\text{ERFC.PRECISE}(x)=\text{ERFC}(x)$.
+## Usage
+### Syntax
+**ERFC.PRECISE(<span title="Number" style="color:#1E88E5">X</span>) => <span title="Number" style="color:#1E88E5">erfc.precise</span>**
+### Argument descriptions
+* *X* ([number](/features/value-types#numbers), required). The lower integration limit to be used to calculate the complementary error function. ERFC.PRECISE integrates over the range [X, $\infty$).
```
nhatcher: Do we need to specify what function is integrated over the range? also I think we can use Latex for the whole range:

```
[X, \infty)
```
Rather than just the $\infty$ symbol
$$[X, \infty)$$

nhatcher (changes requested):
Hi @stevethesleeve, I left a couple of questions

stevethesleeve:
Hi @nhatcher 

I will update .md and .xlsx files for the four ERF/ERFC functions this evening to address your comments - I don't want to leave something half done.

Regards,

@stevethesleeve

stevethesleeve:
Hi @nhatcher 

Files updated. Please note the following:
(1) Running the development server, select the Functions > Engineering option. Note that the ERF.PRECISE and ERFC.PRECISE menu options currently reference non-existent HTML files. Maybe you could fix?
(2) Presumably related to the previous point, but at the bottom of the ERF page the Previous page link again refers to a non-existent HTML file.
(3) On the ERFC page, I don't think we need to specify which function is integrated over the range within the argument description. The equation is included lower down the page.

Regards,

@stevethesleeve

nhatcher:
Thank you @stevethesleeve for all your invaluable work here!
I will continue this PR in #251

PR #248, opened 2025-01-19
FIX: Uses python3 if python is not available

nhatcher:


nhatcher:
FYI: @dg-ac this fixes the python/python3

PR #252, opened 2025-01-21
UPDATE: Deleting a sheet prompts a confirmation dialog

dg-ac:
Hi,

This PR introduces a confirmation dialog to prevent accidental deletion of sheets and potential data loss. To see it, click on the chevron icon and then on 'Delete':

image

After that, a the dialog like this one that should appear.

image

Thanks,
D

### DIFF
--- a/webapp/IronCalc/src/locale/en_us.json
+++ b/webapp/IronCalc/src/locale/en_us.json
@@ -68,6 +68,13 @@
     "title": "Rename Sheet",
     "close": "Close dialog"
   },
+  "sheet_delete": {
+    "title": "Are you sure?",
+    "message1": "The sheet",
+    "message2": "will be permanently deleted. This action cannot be undone.",
+    "confirm": "Yes, delete sheet",
+    "cancel": "Cancel"
+  },
   "formula_input": {
     "update": "Update",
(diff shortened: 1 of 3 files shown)


webapp/IronCalc/src/locale/en_us.json
```diff
@@ -68,6 +68,13 @@
     "title": "Rename Sheet",
     "close": "Close dialog"
   },
+  "sheet_delete": {
+    "title": "Are you sure?",
+    "message1": "The sheet",
+    "message2": "will be permanently deleted. This action cannot be undone.",
```
nhatcher: You can do this with a single message and a substitution template. Do you want to research how to do this?

nhatcher (changes requested):
Sorry for the late review. One comment, let me know if you have questions

webapp/IronCalc/src/locale/en_us.json
dg-ac: Yep, let me work on that.

dg-ac: Changed, please take a look 🙂

nhatcher (approved):
LGTM!

PR #278, opened 2025-02-15
FIX: Forces calculation after insert/delete columns/rows

nhatcher:


nhatcher:
FYI: @dg-ac , regresion tests added

PR #279, opened 2025-02-15
FIX: Shows borders correctly in case of frozen rows.

nhatcher:
Also draws the lines at the correct position


nhatcher:
FYI: @dg-ac

PR #282, opened 2025-02-18
FIX: It it possible to have DF scoped to the first sheet

nhatcher:


nhatcher:
FYI: @dg-ac , that's the bug we saw yesterday on the call.

PR #289, opened 2025-02-23
FIX: Add "Apply" button to color picker

nhatcher:
image

Hey Dani, this adds a much needed button to the color picker


dg-ac:
Thanks @nhatcher. I may want to change some cosmetics here but I think it works for now!

PR #310, opened 2025-02-23
UPDATE: Update to Rust 2024 edition

nhatcher:
:rocket: :rocket: :rocket:


nhatcher:
The test are failing because GitHub doesn't have 1.85 yet

nhatcher:
Will do later

Issue #304, opened 2025-02-23
Update: Share button dialog


dg-ac:
Having a dialog as a 2nd step here should avoid users clicking ‘Share’ multiple times. We could start with a simple menu like this one below:

<img width="575" alt="Image" src="https://github.com/user-attachments/assets/d056a053-2387-42e0-8855-5d221462638a" />

Icons:

- `globe-lock` from Lucide
- `copy` from Lucide

[Link to Figma](https://www.figma.com/design/grMy3fOW69WhRKtsb8geZk/%E2%9A%99%EF%B8%8F-IronCalc?node-id=1694-24924&t=WcFiF4xjH22IxJyK-1)

nhatcher:
This is done (merged and deployed) See #311 and #318

PR #311, opened 2025-02-23
UPDATE: Prompt a dialog when clicking 'Share'

dg-ac:
> Note: this should address #304 

This PR intends to prompt a dialog when users click on 'Share'. The reason behind this is to avoid them clicking 'Share' multiple times when sharing workbooks. It also includes a QR code with the idea of getting the mobile and desktop versions closers.

Should look like this:

image

---

To test it, please

- [x] Ensure the URL is created properly
- [x] Ensure the 'copy URL' button works properly
- [x] Ensure it's visible on mobile (not the QR code, as it's hidden on mobile devices)

### DIFF
--- a/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
@@ -1,7 +1,6 @@
 import styled from "@emotion/styled";
 import type { Model } from "@ironcalc/workbook";
 import { IronCalcIcon, IronCalcLogo } from "@ironcalc/workbook";
-import { CircleCheck } from "lucide-react";
 import { useRef, useState } from "react";
 import { FileMenu } from "./FileMenu";
 import { ShareButton } from "./ShareButton";
@@ -18,7 +17,6 @@ export function FileBar(properties: {
   onDelete: () => void;
 }) {
   const hiddenInputRef = useRef<HTMLInputElement>(null);
-  const [toast, setToast] = useState(false);
   const [isDialogOpen, setIsDialogOpen] = useState(false);
 
   return (
@@ -55,20 +53,7 @@ export function FileBar(properties: {
         type="text"
         style={{ position: "absolute", left: -9999, top: -9999 }}
       />
-      <div style={{ marginLeft: "auto" }}>
-        {toast ? (
-          <Toast>
-            <CircleCheck style={{ width: 12 }} />
-            <span
-              style={{ marginLeft: 8, marginRight: 12, fontFamily: "Inter" }}
-            >
-              URL copied to clipboard
-            </span>
-          </Toast>
-        ) : (
-          ""
-        )}
-      </div>
+      <div style={{ marginLeft: "auto" }}></div>
       <DialogContainer>
         <ShareButton onClick={() => setIsDialogOpen(true)} />
         {isDialogOpen && (
--- a/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
@@ -9,6 +9,8 @@ import { ShareButton } from "./ShareButton";
 import { WorkbookTitle } from "./WorkbookTitle";
 import { downloadModel, shareModel } from "./rpc";
 import { updateNameSelectedWorkbook } from "./storage";
+import { Button } from "@mui/material";
+import ShareWorkbookDialog from "./ShareWorkbookDialog";
 
 export function FileBar(properties: {
   model: Model;
@@ -19,6 +21,8 @@ export function FileBar(properties: {
 }) {
   const hiddenInputRef = useRef<HTMLInputElement>(null);
   const [toast, setToast] = useState(false);
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
+
   return (
     <FileBarWrapper>
       <StyledDesktopLogo />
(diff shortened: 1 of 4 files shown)


webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
```diff
@@ -19,6 +19,8 @@ export function FileBar(properties: {
 }) {
   const hiddenInputRef = useRef<HTMLInputElement>(null);
   const [toast, setToast] = useState(false);
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
```
nhatcher, line 20: I think you need to remove the previous code, the "toast". So line 21 should go. Also line 4 with the lucide svg that we don't use anymore. Finally the whole toast in lines 51 to 71 (approx)

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
```diff
@@ -0,0 +1,202 @@
...
+  const handleClose = () => {
+    properties.onClose();
+  };
+
+  const handleCopy = async () => {
+    try {
+      await navigator.clipboard.writeText(url);
+      setCopied(true);
+    } catch (err) {
+      console.error("Failed to copy text: ", err);
```
nhatcher, line 51: We should have a modal dialog alerting the user. Maybe a good time to create one?

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
```diff
@@ -0,0 +1,202 @@
...
+        <URLWrapper>
+          <StyledTextField
+            hiddenLabel
+            disabled
+            value={url}
+            variant="outlined"
+            fullWidth
+            margin="normal"
+            size="small"
+            style={{
```
nhatcher: You are using `StyledTextField` already why not put all properties there?

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
```diff
@@ -0,0 +1,202 @@
...
+              fontSize: "12px",
+              paddingTop: "0px",
+            }}
+          />
+          <StyledButton
+            variant="contained"
+            color="primary"
+            size="small"
+            onClick={handleCopy}
+            style={{ textTransform: "capitalize", fontSize: "14px" }}
```
nhatcher: Same comment, why not put this style in the `StyledButton` definition. It's normally one or the other. The reason for doing both is that you have several of them that are slightly so different.

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
```diff
@@ -0,0 +1,202 @@
...
+}) {
+  const [url, setUrl] = useState<string>("");
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    const generateUrl = async () => {
+      if (properties.model) {
+        const bytes = properties.model.toBytes();
+        const fileName = properties.model.getName();
+        const hash = await shareModel(bytes, fileName);
```
nhatcher, line 21: This `await` means the hash comes from the server, so the svg can.
At the time of writing this dependency adds 77kB to the uncompressed build. Certainly not a lot, but enough to cosider

nhatcher: NB: no action needed on your side.

nhatcher (changes requested):
Awesome!
I deployed a version of this to try it out.
You will need to squash the commits and remove the WIP commit.
Let me know if you need help with that

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
dg-ac: You mean, some kind of error modal?

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
dg-ac: You're right, I can change that. But are you only talking about `style`, or also about `fullWidth`, `margin` and `size`?

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
dg-ac: I'll change that, thanks!

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
nhatcher: Yeah, but perhaps not a great idea. This is part of the application, not the widget. I guess it is ok.

webapp/app.ironcalc.com/frontend/src/components/ShareWorkbookDialog.tsx
nhatcher: Only the `style` part

dg-ac:
Hi @nhatcher, thanks for the comments. Just pushed 2 commits, please take a look and resolve the pending conversations (unless there are open topics).

nhatcher:
Hi @dg-ac , there is a WIP commit (the first one)

nhatcher:
Superseded by #318

PR #327, opened 2025-03-06
FIX[app.ironcalc.com]: Clean up code for the title

nhatcher:
Hi Dani, in this PR:

- Add `useWindowWidth` hook to track and update window width on resize.
- Introduce `maxWidth` prop to `WorkbookTitle` for responsive truncation.
- Replace hidden text element with an editable `input` controlled by a mirror `div` for measuring width.
- Position the title absolutely in `FileBar` and use a spacer to keep it centered.
- Allow pressing “Enter” to commit name changes and “Escape” to revert them.


dg-ac:
Hi @nhatcher,

Looks good! I'd only adjust 2 things, if possible:

- I'd reduce the max-width so it doesn't overlap the 'Help' button
image

- Pressing 'Esc' reverts the changes but I'd also unfocus the field

PR #329, opened 2025-03-17
UPDATE: New color picker

dg-ac:
Hi there,

This PR intends to update our current color pickers in order to improve usability. The new color picker brings some new characteristics:

- Preselected color palette
- Immediate feedback – no need to confirm color selection
- Custom and recent colors shown together

image

This change affects 4 places:

1. Text color
2. Cell background
3. Border color
4. Sheet button highlight color

Also, note there are still a couple of things that are not implemented yet and are needed:

1. Default color for background fill should be white or transparent, and its label should say 'No fill'. The square container should be white.
2. Same for to Sheet buttons
3. For border colors, the label should say "No border"
4. 'Cancel' button should open back the _palette menu_ when clicked
5. On border colors, the picker should be placed on the right of the button that opens it


nhatcher:
Edit: I think I understood. Have a look at the last implementation

nhatcher:
> For border colors, the label should say "No border"
No border is not achieved with a color. The default border color is indeed black, right?

nhatcher:
Hi @dg-ac , this is ready for a new review:

* Removed all optional parameters
NB: A good rule of thumb is to NOT use default parameters and not use parameters that can be null
* Position Dialogs correctly
NB: you can study the playground in https://mui.com/material-ui/react-popover/#anchor-playground
* Split the ColorPicker in two different components
Splitting things this way allows to see more relevant code at one glance and helps maintainability.
* Added a border to the default color swatch

Let me know what you think. We can go over the changes in a call if you like

elsaminsut:
Hi @nhatcher,

Just checked everything and looks great. Thanks for your help and for taking the time to explain everything 🙏

Best,
D

dg-ac:
Hi @nhatcher,

Just checked everything and looks great. Thanks for your help and for taking the time to explain everything 🙏

Best,
D

nhatcher (approved):
LGTM!

webapp/IronCalc/src/components/Toolbar/Toolbar.tsx
```diff
@@ -410,28 +410,41 @@ function Toolbar(properties: ToolbarProperties) {
...
+        type="button"
+        $pressed={false}
+        onClick={() => {
+          // Add your onClick handler logic here
+        }}
+        disabled={!canEdit}
+        title={t("toolbar.new_button")}
+      >
+        {/* Add your button icon or text here */}
+      </StyledButton>
```
nhatcher, line 442: Ah! This shouldn't be here. @dg-ac , do you think you can do a PR removing this code?

webapp/IronCalc/src/components/Toolbar/Toolbar.tsx
```diff
@@ -67,7 +67,7 @@ type ToolbarProperties = {
   onToggleVerticalAlign: (v: string) => void;
   onToggleWrapText: (v: boolean) => void;
   onCopyStyles: () => void;
-  onTextColorPicked: (hex: string) => void;
+  onTextColorPicked: (hex: string | null) => void;
```
nhatcher, line 70: Also this cannot be null

nhatcher: FYI: @dg-ac

webapp/IronCalc/src/components/Toolbar/Toolbar.tsx
dg-ac: Ah, sorry about that! Removed in #333

webapp/IronCalc/src/components/Toolbar/Toolbar.tsx
dg-ac: Removed in #333

PR #332, opened 2025-03-24
UPDATE: Add a drawer to store user's workbooks

dg-ac:
Hi there,

This PR aims to refresh the interface in the webapp by adding a collapsable drawer on the left side. Some reasons for this change:
- This will allow to list more workbooks
- Maybe will also allow search, and bulk actions (like multiple deletion)
- Declutter the current File Menu
- Use to list examples/templates (in the future)
- Use it to place sign in/sign out buttons (in the future)

There are still some missing things until this PR is ready to be merged:
- [ ] Ellipsis buttons should only be shown when hovering the list's items
- [ ] The 'Download workbook' button inside the ellipsis menu doesn't work for now
- [ ] Same for 'Delete workbook'

@nhatcher we can discuss this week.

### DIFF
--- a/webapp/app.ironcalc.com/frontend/src/App.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/App.tsx
@@ -136,13 +136,14 @@ function App() {
             const newModel = Model.from_bytes(bytes);
             saveModelToStorage(newModel);
             setModel(newModel);
-            refreshModelsData(); // Refresh after model change
+            refreshModelsData();
           }}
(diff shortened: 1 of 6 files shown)


nhatcher:
Hi @dg-ac , as a first step could you add the drawer on its own folder?

dg-ac:
@nhatcher Long delay on this but just pushed a new version with a couple of small changes, including the one you suggested. Let me know what you think.

Thanks,
D

PR #369, opened 2025-06-07
UPDATE: Add "Export Area to markdown"

nhatcher:
Hi @dg-ac , I am hesitant to merge this.

On the one hand it is cool, on the other I am not not sure how useful it is. Maybe we can have a common:

Export to:
* PNG
* Markup/down
* ...

### DIFF
--- a/base/src/model.rs
+++ b/base/src/model.rs
@@ -1931,16 +1931,32 @@ impl Model {
     }
 
     /// Returns markup representation of the given `sheet`.
-    pub fn get_sheet_markup(&self, sheet: u32) -> Result<String, String> {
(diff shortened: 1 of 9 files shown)


dg-ac:
@nhatcher I'll take a look this weekend.

BrianHung:
I wouldn't do a sheet to markdown method. I would do a `getSheetValues` or `getSheetContent` that takes `rowStart,rowEnd,colStart,colEnd` and returns a 2D array of string values.

Then the developer can decide whether to convert that into a CSV, TSV, HTML table, or a Markdown table.

nhatcher:
Archiving this for now.
FYI: @dg-ac

PR #371, opened 2025-06-08
feat: add tracking of changed cells

BrianHung:
Exposes an array of changed "evaluated" cells as an API. This is useful for syncing values for UI and for optimistic backend updates; for authoritative backend updates, developer should sync only the updated user value or action.


nhatcher:
Hi @BrianHung, sorry for the tremendous delay in responding. First of all, thank you very much for your contributions, there are so many good ideas in here!

I wanted to find a nice way to incorporate your PRs in the systems. And I have found one but it is a bit of work. But let's go PR by PR.

In this case this is a bit of a hack. The `self.cells` is a internal structure and it is there to support the evaluation algorithm. Exposing it like this will likely break when the internals of the evaluation strategy change.

Also, and this is in common with the other PRs, we can't have APIs that are not justified somehow.  The problem is that if we have an API like `model.get_me_this_and_that` and it is not used in the frontend or anywhere else we are likely to remove it or change it in the future. Specially since we are before version 1.0. I thought about adding a comment "This is used by Blah and Blah, please do not change" But that is not going to scale, also we have no way of properly testing. Also what in the call is the wrong thing to do for the integration? We have no way of knowing unless we call someone up

There is a solution though, the solution is to have another integration supported by IronCalc. If you want to have a proper integration, say with the Glide Data Grid, we just go ahead and do that. This is a big-ish project and we will need to sync before embarking in such and adventure.

Let me know if this makes sense to you

nhatcher (changes requested):
Please let me know what you think

BrianHung:
> In this case this is a bit of a hack.... it is not used in the frontend or anywhere... There is a solution though, the solution is to have another integration supported by IronCalc. If you want to have a proper integration, say with the Glide Data Grid, we just go ahead and do that.

I would actually refactor the iron calc frontend in this case to use this `onCellsEvaluated` API: right now, it fetches and redraws _all_ cells, regardless of whether their values have actually changed.

This API isn't specific to glide-data-grid, and would help re-draw the minimal number of cells.

nhatcher:
Hi @BrianHung, I moved this into a discussion and closing it for the tiem being. But this is a good idea and one we will likely implement in the comming years.

PR #372, opened 2025-06-09
feat: add getRecentDiffs API with TypeScript interfaces

BrianHung:
feat: add getRecentDiffs API with TypeScript interfaces

- Add getRecentDiffs() method to expose recent changes without queue modification
- Include WASM and Node.js bindings with proper serialization
- Add comprehensive TypeScript interfaces for all diff types
- Include tests for various diff scenarios and queue integrity


nhatcher (changes requested):
Interesting. The problem I see is that this introduces code but it is not clear the purpose. Please see comments in:

https://github.com/ironcalc/IronCalc/pull/371#issuecomment-3015008734

BrianHung:
The purpose of this PR is to let UI performantly listen to changes in the calculation engine, and only re-draw what is minimally necessary.

Re-drawing the entire grid every time we get a change is not a scalable pattern that works with other data grid UIs.

nhatcher:
Not a bad idea but parked for the time being

PR #373, opened 2025-06-09
feat: return sheet name and index from newSheet

BrianHung:
Makes `newSheet` return the name and index of the sheet it created, so callers can address the new sheet right away, for example with `renameSheet` or `setUserInput`, instead of doing a separate lookup.

### What changes

`newSheet` now returns `{ name, index }` across the wasm, nodejs and python bindings.

### Usage

```TypeScript
const { name, index } = model.newSheet();
model.setUserInput(index, 1, 1, "Hello");
```


bindings/nodejs/src/user_model.rs
```diff
@@ -92,7 +92,7 @@ impl UserModel {
 
   #[napi(js_name = "newSheet")]
   pub fn new_sheet(&mut self) -> Result<()> {
-    self.model.new_sheet().map_err(to_js_error)
+    self.model.new_sheet().map(|_| ()).map_err(to_js_error)
```
nhatcher: We would want to return the name also here to be consistent.

bindings/wasm/src/lib.rs
```diff
@@ -105,8 +111,12 @@ impl Model {
...
 
     #[wasm_bindgen(js_name = "newSheet")]
-    pub fn new_sheet(&mut self) -> Result<(), JsError> {
-        self.model.new_sheet().map_err(to_js_error)
+    pub fn new_sheet(&mut self) -> Result<JsValue, JsError> {
+        let (name, sheet_index) = self.model.new_sheet().map_err(to_js_error)?;
+
+        let result = NewSheetResult { name, sheet_index };
+
+        serde_wasm_bindgen::to_value(&result).map_err(|e| to_js_error(e.to_string()))
```
nhatcher, line 247: Can you just return `(name, index)`? Maybe not I haven't tried.

nhatcher (changes requested):
This is actually a nice change. Can you:

* Rebase main
* Add tests in the Rust  base module also

bindings/wasm/src/lib.rs
BrianHung: LMK if changes work for you!

BrianHung:
Rebased against main and added a test for core and user_model.

PR #374, opened 2025-06-09
feat: add getSheetDimensions method to WASM bindings

BrianHung:
Useful for letting UI know how many rows and columns to possibly render. For example, I'm using glide-data-grid and it needs to know how many rows and columns to load ahead of time.


nhatcher:
This is a fairly big change. We will need to revisit it when we restudy the main algorithm

PR #375, opened 2025-06-10
feat: add event emitter for diffs

BrianHung:
Allows wasm-binding to listen to changes in diffs.


bindings/wasm/README.pkg.md
```diff
@@ -31,3 +31,20 @@ function compute() {
...
+```TypeScript
+import init, { Model } from "@ironcalc/wasm";
+
+await init();
+
+const model = new Model("Sheet1", "en", "UTC");
+
+model.onDiffs(() => {
+    // React to diff list updates
+    redraw();
```
nhatcher, line 46: This is amazing!

BrianHung: https://github.com/BrianHung/IronCalc/pull/13/commits/4da4bb7360e7a557d62d31e9bd486c0a11468d8c
https://github.com/BrianHung/IronCalc/pull/13/commits/2f5346aa87446996c4884564fc9f7478b0e8a6da

Two downstream commits I would probably merge into this PR! 

I realized diffs should be emitted as a list, and we should output a diff type to allow typescript to do discriminate unions better.

nhatcher:
Hi @BrianHung is it ok if we close this in favour of #380 ?

PR #379, opened 2025-06-12
feat: add xlsx and icalc import export methods

BrianHung:
This does increase the package size from 2.4 mb to 3.7 mb. Alternatives would be to create wasm bindings for xlsx specifically (although that might run into js -> wasm conversion issues now), or to create a separate ironcalc-xlsx.wasm build and keep the normal ironcalc.wasm.


BrianHung:
Would probably close this PR in favor of #381 which separates the calc engine from xlsx handling in a separate wasm file.

That does require two separate calls to init now.

nhatcher:
Hi @BrianHung, closing this in favour of #381

PR #381, opened 2025-06-12
feat: publish xlsx import export as separate wasm

BrianHung:
Publishes XLSX import and export as a separate WebAssembly module, so the core `@ironcalc/wasm` bundle stays the same size for projects that do not use it. This addresses the bundle size concern from #379.

### What the package provides

`@ironcalc/wasm` gives you the core engine (`wasm_bg.wasm`).

`@ironcalc/wasm/xlsx` gives you XLSX import and export (`xlsx_bg.wasm`), as a separate module you can load on demand.

The XLSX helpers live in their own `xlsx-wasm` crate and build the same way as the core engine, with no bin target or empty `main()`. The XLSX module also runs on its own for straight file conversion, without loading the calc engine. This is relevant to #1113 (converting xlsx in the browser).

### Usage

```TypeScript
import init, { Model } from "@ironcalc/wasm";
import initXLSX, { toXLSXBytes, fromXLSXBytes } from "@ironcalc/wasm/xlsx";

await init();
await initXLSX();

const model = new Model("Workbook1", "en", "UTC", "en");
const xlsxBytes = toXLSXBytes(model.toBytes(), "en");
const modelBytes = fromXLSXBytes(xlsxBytes, "Workbook1", "en", "UTC", "en");
```


nhatcher:
Hi @BrianHung,

This is definitely a good idea! And people have asked for this. Lets do it.

But I am not sure if we can use the xlsx crate as is, as that crate might grow in considerable size.

That being said my first problem is that building the wasm as per your instructions I actually get a smaller build! What am I doing wrong?

```
➜  wasm git:(xlsx) ✗ ls -all pkg/*.wasm
-rw-rw-r-- 1 nicolas nicolas 2395869 Jun 28 08:52 pkg/ironcalc_bg.wasm
-rw-rw-r-- 1 nicolas nicolas 2367828 Jun 28 08:52 pkg/xlsx_bg.wasm
```

xlsx/Cargo.toml
```diff
@@ -12,7 +12,9 @@ readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-zip = "0.6"
+zip = { version = "0.6.6", default-features = false, features = ["deflate", "time"] }
+flate2 = { version = "1.0", default-features = false, features = ["rust_backend"] }
+time = { version = "0.3", features = ["wasm-bindgen"] }
```
nhatcher, line 22: Could you explain why we need all this?

nhatcher:
Or is it that you need _both_ wasm files? That is an awesome feat! I was imagining you would have two builds:

* One of 2.4 Mb that is the old wasm
* A new one of 3.7 Mb that contains the model and the xlsx import/export

I have done that in the past and discarded for now as I didn't see an immediate benefit for IronCalc. But having two different packages is interesting as you could load the second one on demand only when importing/exporting.

bindings/wasm/src/xlsx.rs
```diff
@@ -0,0 +1,39 @@
...
+    writer.flush().map_err(|e| to_js_error(e.to_string()))?;
+    Ok(writer
+        .into_inner()
+        .map_err(|e| to_js_error(e.to_string()))?
+        .into_inner())
+}
+
+fn main() {
+    // This is required for cargo to compile this as a binary
+}
```
nhatcher: I'm a bit curious about this, why didn't you go the same way as the other build? Note in Cargo you can have different build targets. See for instance in IronCalc:

https://github.com/ironcalc/IronCalc/blob/main/base/Cargo.toml#L24-L27

https://github.com/ironcalc/IronCalc/blob/main/base/src/model.rs#L38-L64

Maybe that would make things easier? We just have two builds with different features?

nhatcher (changes requested):
Hi @BrianHung, I left a few questions, mostly for my own understanding. Once I understand what is going on I would like to streamline the process a bit. But I think this is awesome!

xlsx/Cargo.toml
BrianHung: xlsx are actually zip files, so we need to bundle the utility functions `flate2` (for deflating a zip file) and `time` (timestamps in headers) along with it.

BrianHung:
> But having two different packages is interesting as you could load the second one on demand only when importing/exporting.

Yep, that's right! We keep ironcalc and xlsx import/export as separate wasm files so we don't cause client to have an unnecessary larger bundle size if they don't use xlsx utilities.

bindings/wasm/src/xlsx.rs
BrianHung: This is so we can have one crate, but two wasm files outputted.

If there isn't a main declaration, cargo would just treat it as a library and won't turn it into an artifact.

I'm pretty new to rust, so if you have a preferred method that achieves the same goal, I would defer to you here!

jorgedanisc:
Bumping this to ask if there is anything blocking progress here, or is this on hold for a specific reason? Just trying to understand the status.

For context, I came across this PR while looking for a way to handle ironcalc's xlsx import/export in WASM, which is something I'd find useful.

sylvinus:
We'd also be interested in a making it possible to convert xlsx from the browser (see #1113). Having a separate wasm to load for that feature is fine.

jorgedanisc:
Hi @nhatcher, just following up on this. Have you had a chance to make any progress on the separate WASM build approach since the discussion in #1113?

Still very interested in having XLSX import/export available client-side without increasing the main IronCalc bundle.

BrianHung:
Updated the PR and merged in the latest main, with conflicts resolved. XLSX now builds as its own `cdylib` crate instead of a bin target, and the README and usage examples are updated to match. Should be ready for another look now, @nhatcher. Thanks!

bindings/wasm/src/xlsx.rs
BrianHung: Addressed this one: the xlsx bindings now live in their own `xlsx-wasm` cdylib crate that builds like the core engine, so the empty `main()` is gone. Kept it as a separate module so the core bundle does not grow.

nhatcher:
Hi all, unless there is some showstopper I can't think of right now I will make sure this gets merged within a week and it's ready for v0.9.0

Thanks for your patience. This is a great feature!

Makefile
```diff
@@ -1,7 +1,8 @@
 .PHONY: lint
 lint:
 	cargo fmt -- --check
-	cargo clippy --all-targets --all-features -- -W clippy::unwrap_used -W clippy::expect_used -W clippy::panic -D warnings
+	cargo clippy --workspace --exclude wasm --all-targets --all-features -- -W clippy::unwrap_used -W clippy::expect_used -W clippy::panic -D warnings
```
nhatcher, line 4: Why do you need to exclude wasm?

xlsx/Cargo.toml
nhatcher: But it did work initially. It doesn't examplin why you need them now. I _think_ this could be done a bit nicer.

bindings/wasm/fix_package.py
```diff
@@ -14,6 +14,19 @@
...
 
+# The XLSX helpers are built as a second wasm bundle and copied into pkg/ after
+# wasm-pack runs, so wasm-pack never lists them. Ship them and expose the core
+# engine at the package root and the helpers at `/xlsx`, so consumers can
+# `import ... from "@ironcalc/wasm/xlsx"`.
+for xlsx_file in ("xlsx.js", "xlsx.d.ts", "xlsx_bg.wasm"):
+    if xlsx_file not in package["files"]:
+        package["files"].append(xlsx_file)
+
+package["exports"] = {
```
nhatcher, line 25: This will break things, I think. There didn't used to be an exports section. This means that every path not listed is sealed.

bindings/wasm/src/lib.rs
```diff
@@ -136,6 +136,7 @@ impl Model {
         Ok(Model { model })
     }
 
+    #[wasm_bindgen(js_name = "fromBytes")]
```
nhatcher, line 139: This will break a lot of things. You can't just rename a method in the library bindings without renaming all the calling sistes.

But to be fair, this is my bad. The typescript should be part of the linter. I will add that

.github/workflows/build-test.yaml
```diff
@@ -56,6 +60,8 @@ jobs:
         run: |
           curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
           npm install -g pnpm@9.15.4
+      - name: Install wasm-bindgen-cli
+        run: cargo install wasm-bindgen-cli
```
nhatcher, line 64: I need to think about this. Note that npm.yaml does not install it.
I think we should just use wasm-pack instead of cargo build + wasm-bindgen CLI. For instance you forgot to run wasm-opt on the build

xlsx/src/import/mod.rs
```diff
@@ -208,3 +208,13 @@ pub fn load_from_icalc<'a>(file_name: &str, language_id: &'a str) -> Result<Mode
         .map_err(|e| XlsxError::IO(format!("Failed to decode file: {e}")))?;
     Model::from_workbook(workbook, language_id).map_err(XlsxError::Workbook)
 }
+
+/// Loads a [`Model`] from the bytes of an `ic` file.
+pub fn load_from_icalc_bytes<'a>(
```
nhatcher, line 213: You don't need this function. You can just do `Model::from_bytes`

nhatcher (changes requested):
Hi @BrianHung,  I left some questions and extracted a couple of PRs out of this one.

It might be easier for you to close this PR and open a new one.

I think the big question is do we want a package that only contains the import/export or can we have two packages:

* The one we have now
* A larger one that includes what we have now +import/export

I _think_ the second option is better.

Also you have done a great job pushing for this. If you preffer I can close this PR and do it myself.

BrianHung:
@nhatcher Opened a new PR that follows the two-package split you preferred: @ironcalc/wasm stays as-is, @ironcalc/wasm-xlsx is the same engine plus Model.fromXlsx / toXlsx. https://github.com/ironcalc/IronCalc/pull/1363

nhatcher:
Superseeded by #1363

PR #382, opened 2025-06-12
feat: use wasm-bindgen unchecked_return_type

BrianHung:
https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-rust-exports/unchecked_type.html


nhatcher (approved):
Awesome job! Many thanks!

PR #408, opened 2025-07-11
feat: insert delete rows cols

BrianHung:
bulk insert delete rows cols

### DIFF
--- a/base/src/test/user_model/test_batch_row_column_diff.rs
+++ b/base/src/test/user_model/test_batch_row_column_diff.rs
@@ -1,0 +1,154 @@
+#![allow(clippy::unwrap_used)]
+
+use bitcode::decode;
+
+use crate::{
+    test::util::new_empty_model,
+    user_model::history::{Diff, QueueDiffs},
+    UserModel,
+};
+
+fn last_diff_list(model: &mut UserModel) -> Vec<Diff> {
+    let bytes = model.flush_send_queue();
+    let queue: Vec<QueueDiffs> = decode(&bytes).unwrap();
+    // Get the last operation's diff list
+    queue.last().unwrap().list.clone()
+}
+
+#[test]
+fn diff_invariant_insert_rows() {
+    let base = new_empty_model();
+    let mut model = UserModel::from_model(base);
+
+    assert!(model.insert_rows(0, 5, 3).is_ok());
+
+    let list = last_diff_list(&mut model);
+    assert_eq!(list.len(), 3);
+    for diff in list {
+        match diff {
+            Diff::InsertRow { sheet, row } => {
+                assert_eq!(sheet, 0);
+                assert_eq!(row, 5);
+            }
+            _ => panic!("Unexpected diff variant"),
+        }
+    }
+}
+
+#[test]
+fn diff_invariant_insert_columns() {
+    let base = new_empty_model();
+    let mut model = UserModel::from_model(base);
+
+    assert!(model.insert_columns(0, 2, 4).is_ok());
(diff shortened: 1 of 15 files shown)


nhatcher (changes requested):
This is great!

Some comments:

1. we shouldn't keep the old API for insert 1 row or column
2. We need tests in the Rust code (its also fine to have them in JavaScript. Test should cover corner cases, like too many columns, negative values, etc

Point 1. will imply changing the UI a little bit. You don't need to add UI to insert bulk columns/rows as @dg-ac will need to do some design, juts adapt the API call.

It was a mistake to do it like this for me in the first place. Can't wait for this to be in place.

BrianHung:
I didn't know whether to remove the singular case to keep things backwards compatible or not. Might require a version bump if we do? If you're fine with it, I'll remove it!

nhatcher:
> I didn't know whether to remove the singular case to keep things backwards compatible or not. Might require a version bump if we do? If you're fine with it, I'll remove it!

Yes, please remove them. I'll handle releases and version bumps. You will need to update bindings and the js code of course, but should be pretty straightforward. you will need to change places like this:
https://github.com/ironcalc/IronCalc/blob/main/webapp/IronCalc/src/components/Worksheet/Worksheet.tsx#L343
But you don't need to update the UI or anything

BrianHung:
Do you think we should also support `InsertRows` `InsertCols` `DeleteRows` `DeleteCols` as diff types? Would be easier to multiplayer support if it builds on top of diffs 🤔 Can do in follow-up PR.

base/src/user_model/history.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(deprecated)]
```
nhatcher: Please remove all deprecated functions. We really don't need them. At this point it is ok to break this.

base/src/user_model/common.rs
```diff
@@ -2478,4 +2614,6 @@ mod tests {
             assert_eq!(horizontal(&format!("{a}")), Ok(a));
         }
     }
+
+    // (batch diff tests moved to separate file)
```
nhatcher: You can remove this two lines

base/src/test/user_model/test_batch_row_column_diff.rs
```diff
@@ -0,0 +1,681 @@
...
+    let list = last_diff_list(&mut model);
+    assert_eq!(list.len(), 1);
+    match &list[0] {
+        Diff::InsertRows { sheet, row, count } => {
+            assert_eq!(*sheet, 0);
+            assert_eq!(*row, 5);
+            assert_eq!(*count, 3);
+        }
+        _ => panic!("Unexpected diff variant"),
+    }
```
nhatcher: A slightly nicer way to do this is:
```
assert!(matches!(list[0], Diff::InsertRows { sheet: 0, row: 5, count: 3 }));
```

nhatcher (changes requested):
Almost there!
Please remove the deprecated methods. Once taht's done I'll approve.

base/src/user_model/history.rs
BrianHung: Will do!

BrianHung:
wasm-pack / CI CD broken upstream but tests + code works locally for me.

nhatcher (approved):
LGTM!

nhatcher:
Could you rebase main? There is a small conflict, but easy to resolve

Many thanks for this feature!

BrianHung:
pushed force with lease; would just squash and merge all the commits!

PR #409, opened 2025-07-12
update: add a Help menu to App

dg-ac:
This PR introduces a small change in the file bar of the app by adding a new menu for the Help button, which until now it was just opening the documentation.

The update avoids confusion caused by accidental clicking (as it was just opening a new tab which was very sudden). In the future we should add other items to this menu and maybe place the shortcuts in its own dialog, instead of just relying in the Docs.

image

### DIFF
--- a/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
@@ -3,6 +3,7 @@ import type { Model } from "@ironcalc/workbook";
 import { IronCalcIcon, IronCalcLogo } from "@ironcalc/workbook";
 import { useLayoutEffect, useRef, useState } from "react";
 import { FileMenu } from "./FileMenu";
+import { HelpMenu } from "./HelpMenu";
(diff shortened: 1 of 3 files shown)


nhatcher (changes requested):
Hi @dg-ac I think we should link to the general documentation of IronCalc as you do and the documentation about the Web application:

https://docs.ironcalc.com/web-application/about.html

I don't think the keyboard shortcuts are any more important than the name manager, for instance. Maybe if we add some keyboard shortcuts for the Web Application.

I think we should document a bit better the keyboard shortcuts. I will show you how to do that reading the code. There is no real difference between MaC and Linux. Except MaC uses Command where Linux/Windows uses Control key. To navigate to the end of the page you don't really use "Fn + Right Arrow" you use the "End" key, some keyboards (including the Macbook Pro and my Lenovo) map Fn + Right Arrow to End,for instance. there are many mare keyboard shortcuts that we shoud document a as Enter, F2, or block navigation.
I would suggest not linking to the keyboard shortcuts until we document it.

PR #421, opened 2025-07-18
feat: add customization for the base api url

Nastaliss:


nhatcher:
This is great! I'll try and have a look over the weekend

Nastaliss:
@nhatcher I do not think [this](https://github.com/ironcalc/IronCalc/pull/421/commits/42eef14b899f4a0e254b8ffbbfe26ce97f4ee6a9) will be necessary, using Caddy inside docker at first will work.

PR #427, opened 2025-07-30
financial functions

BrianHung:


BrianHung:
Will prioritize this over the weekend.

nhatcher:
See https://github.com/ironcalc/IronCalc/pull/518

nhatcher:
Hi @BrianHung, those are finally done. These are really difficult functions and took a lot of testing, reasearch, understanding,... 
We still only have a "beta" release of those but they are in the system finally.

PR #431, opened 2025-08-09
Fix nodejs binding import

mattlehrer:
use nodejs package instead of wasm here


mattlehrer:
replaced with https://github.com/ironcalc/IronCalc/pull/432

PR #441, opened 2025-09-20
FIX: Shift+Letter starts editing cell

nhatcher:


nhatcher:
FYI: @dg-ac , @elsaminsut thanks for reporting.

PR #449, opened 2025-10-08
Fix example in README to reference Model correctly.

dsagal:
Since https://github.com/ironcalc/IronCalc/pull/27, using `base::model::Model` complains about `model` being a private module, and `base::Model` should be used instead.


nhatcher (approved):
Many thanks!

Issue #390, opened 2025-06-29
Please design new "Name Manager"


nhatcher:
Hi @dg-ac, The present defined name manager has served us good, but it is very difficult to use. I think we need a new "Name Manager".

Most likely it neds to open to the right of the spreadsheet like other modern spreadsheet engines do. This leaves us space to see the spreadsheet itself and also use "browse mode", that is set a formula using user selected cells and ranges.

Challenges:

* I think the drawer in the right must me resizable
* You need to be able to filter names
* I think we want a preview of the value (do we?) or know if they are in an error state
* Most importantly this needs to be in sync with other drawers tha open to the right (conditional formatting, cell styles and formats also maybe?)

The last point makes this difficult to design in isolation. Also usability is an issue here. We need to assume that there might be hundreds of defined names some of them in error states and some of them with fairly long formulas.

Also note that formulas for defined names are not as generic as in cells. For instance "=A1" is not a valid formula you will need "=Sheet1!A1" and although that might work what you really want is "=Sheet1!$A$1"

We will probably need a bit on back and forth on this one but I think nailing this "right drawers" thing down is an important step.

Note that your idea of having a left drawer in the application to manage workbooks and users is unrelated to this.

FYI: @aloifran

PR #526, opened 2025-11-09
update: name manager v2

dg-ac:
This PR introduces a new version for the **Name Manager**. We've redesigned the way named ranges are displayed and managed in IronCalc. This PR address the issue #390 

image

---

### Changes

1. **Moved Name Manager to a drawer**: the drawer opens from the right side, allowing users to view and edit names without blocking the view

2. **Split the view between Named Ranges list and Add/Edit range**: this supports longer range names and references. It also leaves some space for future features like range-specific comments

image

3. Enabled **Drawer resizing**: clicking on the left edge of the drawer allows making it wider or narrower

image
image

4. Improved **error handling** and user feedback: every time users input invalid names we warn them and tell them why

image
image
image

5. If the referred cell/range is selected, we now we mark their named range

image

6. Added **mobile support**

7. **Additional improvements**
   - Updated translation files to include new strings
   - Toolbar now opens the drawer instead of the old dialog
   - Usage of proper MUI tooltips

---

### Testing

- [ ] Verify the drawer opens correctly from the toolbar.
- [ ] Check that all Name Manager actions (add, edit, delete, rename) work as expected.  
- [ ] Confirm drawer resizing behaves correctly (only desktop)
- [ ] Validate that the main spreadsheet remains fully visible and interactive when the drawer is open
- [ ] Test responsiveness and layout on both desktop and mobile devices

---

### Notes

This is just a first iteration of the new Name Manager. Soon we will add more capabilities, such as:

- Search and filtering
- List ranges from the container at the left of the formula bar
- Select ranges directly from the grid, without typing
- Find the referred cells by clicking on the Named range
- Range-specific comments/notes
- Handling dynamic referrals (what happens if we don't use `$` to lock it and insert columns/rows?)
- Previewing if the Range is in an error state
- Extensive documentation of the feature

### DIFF
--- a/webapp/IronCalc/src/components/RightDrawer/NamedRanges/EditNamedRange.tsx
+++ b/webapp/IronCalc/src/components/RightDrawer/NamedRanges/EditNamedRange.tsx
@@ -46,7 +46,7 @@ function EditNamedRange({
     let counter = 1;
     let defaultName = `Range${counter}`;
     const scopeIndex = worksheets.findIndex((s) => s.name === initialScope);
-    const newScope = scopeIndex >= 0 ? scopeIndex : undefined;
+    const newScope = scopeIndex >= 0 ? scopeIndex : null;
 
     while (
       definedNameList.some(
@@ -62,17 +62,15 @@ function EditNamedRange({
   const [name, setName] = useState(getDefaultName());
   const [scope, setScope] = useState(initialScope);
   const [formula, setFormula] = useState(initialFormula);
-  const [nameError, setNameError] = useState<string | undefined>(undefined);
-  const [formulaError, setFormulaError] = useState<string | undefined>(
-    undefined,
-  );
+  const [nameError, setNameError] = useState<string>("");
+  const [formulaError, setFormulaError] = useState<string>("");
 
   const isSelected = (value: string) => scope === value;
 
   // Validate name (format and duplicates)
   useEffect(() => {
     const trimmed = name.trim();
-    let error: string | undefined;
+    let error = "";
 
     if (!trimmed) {
       error = t("name_manager_dialog.errors.range_name_required");
@@ -85,7 +83,7 @@ function EditNamedRange({
     } else {
       // Check for duplicates only if format is valid
       const scopeIndex = worksheets.findIndex((s) => s.name === scope);
-      const newScope = scopeIndex >= 0 ? scopeIndex : undefined;
+      const newScope = scopeIndex >= 0 ? scopeIndex : null;
       const existing = definedNameList.find(
         (dn) =>
           dn.name === trimmed &&
@@ -103,7 +101,7 @@ function EditNamedRange({
     setNameError(error);
   }, [name, scope, definedNameList, editingDefinedName, worksheets]);
 
-  const hasAnyError = nameError !== undefined || formulaError !== undefined;
+  const hasAnyError = nameError !== "" || formulaError !== "";
 
   return (
     <Container>
@@ -219,7 +217,7 @@ function EditNamedRange({
                 value={formula}
                 onChange={(e) => {
                   setFormula(e.target.value);
-                  setFormulaError(undefined);
+                  setFormulaError("");
                 }}
                 onKeyDown={(e) => e.stopPropagation()}
                 onClick={(e) => e.stopPropagation()}
--- a/webapp/IronCalc/src/components/RightDrawer/NamedRanges/NamedRanges.tsx
+++ b/webapp/IronCalc/src/components/RightDrawer/NamedRanges/NamedRanges.tsx
@@ -94,9 +94,7 @@ function NamedRanges({
       try {
         updateDefinedName(
           editingDefinedName.name,
-          editingDefinedName.scope !== undefined
-            ? editingDefinedName.scope
-            : null,
+          editingDefinedName.scope ?? null,
           name,
           newScope,
           formula,
@@ -119,7 +117,7 @@ function NamedRanges({
     if (editingDefinedName) {
       name = editingDefinedName.name;
       scopeName =
-        editingDefinedName.scope !== undefined
+        editingDefinedName.scope != null
           ? worksheets[editingDefinedName.scope]?.name || "[unknown]"
           : "[Global]";
       formula = editingDefinedName.formula;
@@ -134,7 +132,7 @@ function NamedRanges({
     return (
       <Container>
         <EditHeader>
-          <Tooltip title={t("name_manager_dialog.back")}>
+          <Tooltip title={t("name_manager_dialog.back_to_list")}>
             <IconButtonWrapper
               onClick={handleCancel}
               onKeyDown={(e) => {
@@ -142,7 +140,7 @@ function NamedRanges({
                   handleCancel();
                 }
               }}
-              aria-label={t("name_manager_dialog.back")}
+              aria-label={t("name_manager_dialog.back_to_list")}
               tabIndex={0}
             >
               <ArrowLeft />
@@ -247,7 +245,7 @@ function NamedRanges({
           <ListContainer>
             {definedNameList.map((definedName) => {
               const scopeName =
-                definedName.scope !== undefined
+                definedName.scope != null
                   ? worksheets[definedName.scope]?.name || "[Unknown]"
                   : "[Global]";
               const isSelected =
@@ -300,9 +298,7 @@ function NamedRanges({
                           if (deleteDefinedName) {
                             deleteDefinedName(
                               definedName.name,
-                              definedName.scope !== undefined
-                                ? definedName.scope
-                                : null,
+                              definedName.scope ?? null,
                             );
                           }
                         }}
@@ -313,9 +309,7 @@ function NamedRanges({
                             if (deleteDefinedName) {
                               deleteDefinedName(
                                 definedName.name,
-                                definedName.scope !== undefined
-                                  ? definedName.scope
-                                  : null,
+                                definedName.scope ?? null,
                               );
                             }
                           }
--- a/webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
+++ b/webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
@@ -49,11 +49,6 @@ const RightDrawer = ({
   const [isResizing, setIsResizing] = useState(false);
   const resizeHandleRef = useRef<HTMLDivElement>(null);
 
-  // Update local width when prop changes
-  useEffect(() => {
-    setDrawerWidth(width);
-  }, [width]);
-
   const handleMouseDown = useCallback((e: ReactMouseEvent) => {
     e.preventDefault();
     setIsResizing(true);
--- a/webapp/IronCalc/src/components/Workbook/Workbook.tsx
+++ b/webapp/IronCalc/src/components/Workbook/Workbook.tsx
@@ -753,13 +753,7 @@ const Workbook = (props: { model: Model; workbookState: WorkbookState }) => {
           newScope: number | null,
           newFormula: string,
         ) => {
-          model.updateDefinedName(
-            name,
-            scope ?? undefined,
-            newName,
-            newScope ?? undefined,
-            newFormula,
-          );
+          model.updateDefinedName(name, scope, newName, newScope, newFormula);
           setRedrawId((id) => id + 1);
         }}
         newDefinedName={(
@@ -767,11 +761,11 @@ const Workbook = (props: { model: Model; workbookState: WorkbookState }) => {
           scope: number | null,
           formula: string,
         ) => {
-          model.newDefinedName(name, scope ?? undefined, formula);
+          model.newDefinedName(name, scope, formula);
           setRedrawId((id) => id + 1);
         }}
         deleteDefinedName={(name: string, scope: number | null) => {
-          model.deleteDefinedName(name, scope ?? undefined);
+          model.deleteDefinedName(name, scope);
           setRedrawId((id) => id + 1);
         }}
         selectedArea={() => {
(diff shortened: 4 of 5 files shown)


webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+    formula: string,
+  ) => void;
+  deleteDefinedName?: (name: string, scope: number | undefined) => void;
+  selectedArea?: () => string;
+}
+
+const RightDrawer = ({
+  isOpen,
+  onClose,
+  width = DEFAULT_DRAWER_WIDTH,
```
nhatcher: Please don't use default arguments like this. Let the function pass the argument

Why this is bad?

The calling function has no idea of what the default parameters are. Some programming languages like python allow for default parameters, the consensus today seems to be that that is more harmful than useful.
Just let the calling function pass all the parameters.

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+import NamedRanges from "./NamedRanges/NamedRanges";
+
+const DEFAULT_DRAWER_WIDTH = 360;
+const MIN_DRAWER_WIDTH = 300;
+const MAX_DRAWER_WIDTH = 500;
+
+interface RightDrawerProps {
+  isOpen: boolean;
+  onClose: () => void;
+  width?: number;
```
nhatcher: None of those properties need to be optional, please remove the '?' for the type in all of them.

Why this is bad?

Because we effectively loose all type information. Now the parameter might or might not be there so every time we want to use it we need to check first if it exists. Also the calling function might make a mistake by not providing a parameter.

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+    e.preventDefault();
+    setIsResizing(true);
+  }, []);
+
+  useEffect(() => {
+    if (!isResizing) return;
+
+    // Prevent text selection during resize
+    document.body.style.userSelect = "none";
+    document.body.style.cursor = "col-resize";
```
nhatcher, line 62: This set css properties for the whole webpage. We should do this at all.

You can also leave that to me.

Why this is bad?

This applies to the whole page, if we make a mistake it will remain like that. Also in CSS there are specificity rules, meaning that won't necessarily apply f some other CSS rule is more specific.

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    const handleMouseMove = (e: MouseEvent) => {
+      const newWidth = window.innerWidth - e.clientX;
+      const clampedWidth = Math.max(
+        MIN_DRAWER_WIDTH,
+        Math.min(MAX_DRAWER_WIDTH, newWidth),
+      );
+      setDrawerWidth(clampedWidth);
+      onWidthChange?.(clampedWidth);
```
nhatcher: If onWidthChange is always defined you don't need the question mark

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+    };
+
+    const handleMouseUp = () => {
+      setIsResizing(false);
+      document.body.style.userSelect = "";
+      document.body.style.cursor = "";
+    };
+
+    document.addEventListener("mousemove", handleMouseMove);
+    document.addEventListener("mouseup", handleMouseUp);
```
nhatcher, line 81: This might be ok, but it would be much better if we can avoid attaching events to the document.

you can also leave that to me.

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+}
+
+const RightDrawer = ({
+  isOpen,
+  onClose,
+  width = DEFAULT_DRAWER_WIDTH,
+  onWidthChange,
+  children,
+  showCloseButton = true,
+  title = "Named Ranges",
```
nhatcher: Aside from not being a default value the title needs to be i18n.

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+const MIN_DRAWER_WIDTH = 300;
+const MAX_DRAWER_WIDTH = 500;
+
+interface RightDrawerProps {
+  isOpen: boolean;
+  onClose: () => void;
+  width?: number;
+  onWidthChange?: (width: number) => void;
+  children?: ReactNode;
+  showCloseButton?: boolean;
```
nhatcher: You can remove this property, as you always show the close button. If you think we will need it eventually then we can add it then

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+  backgroundColor?: string;
```
nhatcher: You are not using this property at all.

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+  deleteDefinedName?: (name: string, scope: number | undefined) => void;
+  selectedArea?: () => string;
+}
+
+const RightDrawer = ({
+  isOpen,
+  onClose,
+  width = DEFAULT_DRAWER_WIDTH,
+  onWidthChange,
+  children,
```
nhatcher: You are not using this property at all, you can remove it

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  if (!isOpen) return null;
+
+  return (
+    <DrawerContainer $drawerWidth={drawerWidth}>
+      <ResizeHandle
+        ref={resizeHandleRef}
+        onMouseDown={handleMouseDown}
+        $isResizing={isResizing}
+        aria-label="Resize drawer"
```
nhatcher: This needs to be i18n.

webapp/IronCalc/src/components/RightDrawer/NamedRanges/EditNamedRange.tsx
```diff
@@ -0,0 +1,440 @@
...
+    name: string,
+    scope: string,
+    formula: string,
+  ) => SaveError | undefined;
+  onCancel: () => void;
+  definedNameList?: DefinedName[];
+  editingDefinedName?: DefinedName | null;
+}
+
+const EditNamedRange: React.FC<EditNamedRangeProps> = ({
```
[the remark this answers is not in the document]
nhatcher: They are right.

webapp/IronCalc/src/components/RightDrawer/NamedRanges/NamedRanges.tsx
```diff
@@ -0,0 +1,554 @@
...
+import { theme } from "../../../theme";
+import EditNamedRange, { type SaveError } from "./EditNamedRange";
+
+const normalizeRangeString = (range: string): string => {
+  return range.trim().replace(/['"]/g, "");
+};
+
+interface NamedRangesProps {
+  title?: string;
+  onClose?: () => void;
```
nhatcher, line 23: Please don't use optional arguments

webapp/IronCalc/src/components/RightDrawer/NamedRanges/NamedRanges.tsx
```diff
@@ -0,0 +1,554 @@
...
+  newDefinedName?: (
+    name: string,
+    scope: number | undefined,
+    formula: string,
+  ) => void;
+  deleteDefinedName?: (name: string, scope: number | undefined) => void;
+  selectedArea?: () => string;
+}
+
+const NamedRanges: React.FC<NamedRangesProps> = ({
```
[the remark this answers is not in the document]
nhatcher: They are correct

webapp/IronCalc/src/components/RightDrawer/NamedRanges/NamedRanges.tsx
```diff
@@ -0,0 +1,554 @@
...
+    newName: string,
+    newScope: number | undefined,
+    newFormula: string,
+  ) => void;
+  newDefinedName?: (
+    name: string,
+    scope: number | undefined,
+    formula: string,
+  ) => void;
+  deleteDefinedName?: (name: string, scope: number | undefined) => void;
```
nhatcher: Please don'u use undefined. Use the empty string, 0 or null

webapp/IronCalc/src/components/RightDrawer/NamedRanges/NamedRanges.tsx
```diff
@@ -0,0 +1,554 @@
...
+  const handleCancel = () => {
+    setEditingDefinedName(null);
+    setIsCreatingNew(false);
+  };
+
+  const handleSave = (
+    name: string,
+    scope: string,
+    formula: string,
+  ): SaveError | undefined => {
```
nhatcher: Please never return `undefined`.

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
dg-ac: Not sure about how to fix this one.

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
dg-ac: Not sure about how to fix this one either...

webapp/IronCalc/src/components/RightDrawer/RightDrawer.tsx
nhatcher: Let's do this for now

webapp/IronCalc/src/locale/en_us.json
```diff
@@ -104,15 +104,33 @@
...
     "scope": "Range",
-    "help": "Learn more about Named Ranges",
+    "help": "About Named Ranges",
     "new": "Add new",
     "workbook": "Workbook",
     "global": "(Global)",
     "close": "Close dialog",
     "delete": "Delete Range",
     "edit": "Edit Range",
+    "back": "Back to list",
```
nhatcher: Maybe call it `back_to_list`?

webapp/IronCalc/src/components/Workbook/Workbook.tsx
```diff
@@ -660,109 +666,134 @@ const Workbook = (props: { model: Model; workbookState: WorkbookState }) => {
...
+        updateDefinedName={(
+          name: string,
+          scope: number | null,
+          newName: string,
+          newScope: number | null,
+          newFormula: string,
+        ) => {
+          model.updateDefinedName(
+            name,
+            scope ?? undefined,
```
nhatcher: You don't need the `?? undefined`

webapp/IronCalc/src/components/Workbook/Workbook.tsx
```diff
@@ -660,109 +666,134 @@ const Workbook = (props: { model: Model; workbookState: WorkbookState }) => {
...
           setRedrawId((id) => id + 1);
-          focusWorkbook();
         }}
-        onTextUpdated={() => {
+        newDefinedName={(
+          name: string,
+          scope: number | null,
+          formula: string,
+        ) => {
+          model.newDefinedName(name, scope ?? undefined, formula);
```
nhatcher: You don't need the `?? undefined`

webapp/IronCalc/src/components/Workbook/Workbook.tsx
```diff
@@ -660,109 +666,134 @@ const Workbook = (props: { model: Model; workbookState: WorkbookState }) => {
...
         }}
-        model={model}
-        workbookState={workbookState}
-      />
-      <Worksheet
-        model={model}
-        workbookState={workbookState}
-        refresh={(): void => {
+        deleteDefinedName={(name: string, scope: number | null) => {
+          model.deleteDefinedName(name, scope ?? undefined);
```
nhatcher: You don't need the `?? undefined`

webapp/IronCalc/src/components/Workbook/Workbook.tsx
```diff
@@ -660,109 +666,134 @@ const Workbook = (props: { model: Model; workbookState: WorkbookState }) => {
...
+              alert(`${e}`);
+            }
+          }}
+          onSheetRenamed={(name: string): void => {
+            try {
+              model.renameSheet(model.getSelectedSheet(), name);
+              setRedrawId((value) => value + 1);
+            } catch (e) {
+              // TODO: Show a proper modal dialog
+              alert(`${e}`);
```
nhatcher, line 726: Can you either add a `console.error` or add a proper modal dialog. We have them

webapp/IronCalc/src/components/Workbook/Workbook.tsx
```diff
@@ -41,6 +44,9 @@ const Workbook = (props: { model: Model; workbookState: WorkbookState }) => {
   // This is needed because `model` or `workbookState` can change without React being aware of it
   const setRedrawId = useState(0)[1];
 
+  const [isDrawerOpen, setDrawerOpen] = useState(false);
+  const [drawerWidth, setDrawerWidth] = useState(DEFAULT_DRAWER_WIDTH);
```
nhatcher, line 48: I don't understand why we need this

nhatcher (changes requested):
Some more nits

Issue #533, opened 2025-11-11
Edge cases failures in N, SHEET, SHEETS, and CELL Functions


elsaminsut:
## Summary
The test file `N_SHEET_SHEETS_CELL.xlsx` contains edge cases that reveal several bugs in the newly implemented information functions: `N`, `SHEET`, `SHEETS`, and `CELL`. Although `SHEET` is not part of the newly released functions, it's also included in the test.

#### 1. CELL!B6 - Error Type Mismatch
- **Expected**: `#NAME?` error
- **Actual**: `#VALUE!` error with message "Argument must be a reference"
- **Issue**: When CELL receives an invalid reference that Excel would treat as `#NAME?`, IronCalc incorrectly returns `#VALUE!`. 
- **Root Cause**: The CELL function (lines 381-394) calls `evaluate_node_with_reference` and only checks if the result is a Range. If it's not a Range, it immediately returns `#VALUE!` without checking if `evaluate_node_with_reference` returned an error. The function should:
  - Check if `evaluate_node_with_reference` returned an error
  - If it's a `#NAME?` error (from `WrongVariableKind`), propagate it
  - If it's a `#REF!` error (from `WrongReferenceKind`), propagate it
  - Only return `#VALUE!` if the result is not a Range and not an error (which shouldn't happen)

#### 2. CELL!B15 - Incorrect Value
- **Expected**: `4.0`
- **Actual**: `2.0`
- **Issue**: The CELL function is returning an incorrect value. This might be related to confusion between COL (column) and ROW (row) values, or incorrect reference handling. Needs investigation of what info_type and reference are being used.

#### 3. CELL!B32 - Error Type Mismatch for Invalid References
- **Expected**: `#REF!` error
- **Actual**: `#VALUE!` error with message "Argument must be a reference"
- **Issue**: When CELL receives an invalid or deleted reference that Excel would treat as `#REF!`, IronCalc incorrectly returns `#VALUE!`.
- **Root Cause**: Same as B6 - the function doesn't check for errors returned by `evaluate_node_with_reference`. When the parser encounters a `WrongReferenceKind` node, it should return `#REF!`, but the CELL function is ignoring this and returning `#VALUE!` instead.

#### 4. SHEET_SHEETS!B4 - Parsing Issue with _xlfn Prefix
- **Expected**: `#N/A` error
- **Actual**: `#NAME?` error with message "Variable name \"_xleta.CELL\" not found."
[...]

PR #540, opened 2025-11-13
update: adds testing for CELL, INFO, N, SHEET and SHEETS

### LINKED ISSUE #515: Please test and create docs for N, SHEETS, INFO and CELL
Note that:

* INFO is not implemented at all. You could add Rust tests, cheking argument number and that it returns N/IMPL!.
You could also think about what we should implement
* SHEETS is only implemented when it has no arguments
* CELL has a bunch of things that we do not impement. I will implement some of them like COLOR later on, but for now we are compatible with Excl for the Web. you could add the tests manually. (in this case it return #VALUE!)

Good luck!

elsaminsut:
This PR adds testing for the CELL, INFO, N, SHEET and SHEETS functions. These include both xlsx tests with edge cases and unit tests to check the number of arguments taken by the function. It also includes an update to the main Information documentation page and their individual pages to show them as now available.

This PR closes #515

(See failing edge cases in #533)


base/src/test/test_cell_info_n_sheets
```diff
@@ -0,0 +1,35 @@
...
+    assert_eq!(model._get_text("A3"), *"#N/IMPL!");
+    assert_eq!(model._get_text("A4"), *"#ERROR!");
+
+    assert_eq!(model._get_text("A5"), *"1");
+    assert_eq!(model._get_text("A6"), *"#ERROR!");
+    assert_eq!(model._get_text("A7"), *"#ERROR!");
+
+    assert_eq!(model._get_text("A8"), *"1");
+    assert_eq!(model._get_text("A9"), *"#N/IMPL!");
+}
```
nhatcher: Missing new line

elsaminsut: fixed!

nhatcher:
> You can see the failing tests split per function in https://github.com/ironcalc/IronCalc/issues/533

Thanks. In this PR can you split the test case in different files also?
Ideally we would have 1 file per functions.

elsaminsut:
> > You can see the failing tests split per function in #533
> 
> Thanks. In this PR can you split the test case in different files also? Ideally we would have 1 file per functions.

Fixed!

nhatcher (approved):

PR #568, opened 2025-11-20
docs: documentation style guide edits

elsaminsut:
This PR adds a few edits to the recently published documentation style guide for contributors. The aim is to improve overall consistency, readability and clarity.

### DIFF
--- a/docs/src/.vitepress/config.mts
+++ b/docs/src/.vitepress/config.mts
@@ -2036,6 +2036,10 @@ export default defineConfig({
             text: "How to contribute",
             link: "/contributing/how-to-contribute",
           },
+          {
+            text: "Function documentation guide",
+            link: "/contributing/function-documentation-guide",
+          },
         ],
       },
     ],


docs/src/contributing/function-documentation-guide.md
```diff
@@ -40,7 +40,7 @@ The title should be the function name followed by the word "function":
 # FV function
 ```
 
-Use the exact function name as it appears in IronCalc (case-sensitive).
+Use the exact function name as it appears in IronCalc (typically in uppercase).
```
nhatcher: Functions are case insensitive.
=Sin(3) should work just fine

elsaminsut: At least in the docs we refer to functions using their name in uppercase like SIN, COS and TAN functions.

nhatcher: Yes, but that is just for style. IronCalc function names (and Excel, libreoffice, etc) are case insensitive.

elsaminsut: What about this? "The function name should be written in uppercase when mentioned in the documentation."

nhatcher: I missunderstood this. all good

PR #570, opened 2025-11-20
update: add data-storage warnings to the app

dg-ac:
## Summary

This PR introduces a couple of alerts to the app to better communicate potential data-loss risks.  

---

## Changes

1. Added a **warning tooltip** next to the app title to notify users about data-storage limitations.  

image image

2. Added a **dismissible alert** in the left drawer to reinforce the warning in a more persistent, visible way.

image

---

## Testing

- [ ] Hover the warning icon next to the title and confirm the tooltip appears.  
- [ ] Open the left drawer and verify the dismissable alert is displayed.  
- [ ] Dismiss the alert and ensure it properly disappears.  
- [ ] Reload the app and confirm the expected alert behavior (the alert is hidden).  
- [ ] Check responsiveness and layout for both elements.

### DIFF
--- a/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
@@ -41,8 +41,8 @@ export function FileBar(properties: {
   const [maxTitleWidth, setMaxTitleWidth] = useState(0);
   const width = useWindowWidth();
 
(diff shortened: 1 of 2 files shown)


nhatcher:
Nice! I also think it is important to say that IronCalc will release incompatible versions and that workbooks that work today might not work tomorrow. Wether they are shared or in the browser

PR #584, opened 2025-11-25
FIX: Uses lookup tables and lngamma for large numbers

### LINKED ISSUE #574: FACT and FACTDOUBLE testing take too long for large numbers
## Summary
When running `cargo test` the functions FACT and FACTDOUBLE take excessively long to compute for large input values.
This makes the test suite hang or run significantly slower than expected when evaluating factorial operations on big integers.

## Performance Problem
Both FACT(n) and FACTDOUBLE(n) exhibit exponential slowdowns for large n.
Excel returns results instantly for these same values, while IronCalc becomes unresponsive or requires several seconds to finish evaluation.

During execution of the function test files, any cell that calls FACT(n) or FACTDOUBLE(n) with large values leads to long computation times. Excel returns these results instantly (or returns #NUM! when the input exceeds supported limits). 

Even when removing obvious large values (such as 170 or 300) from the test file, the test keeps running for a long time.

## Test file
[FACT_FACTDOUBLE.xlsx](https://github.com/user-attachments/files/23697923/FACT_FACTDOUBLE.xlsx) (With obvious large values removed)

nhatcher:
Fixes #574

### DIFF
--- a/base/src/functions/mathematical.rs
+++ b/base/src/functions/mathematical.rs
@@ -8,7 +8,73 @@ use crate::single_number_fn;
 use crate::{
     calc_result::CalcResult, expressions::parser::Node, expressions::token::Error, model::Model,
 };
+use statrs::function::gamma::ln_gamma;
+use std::f64::consts::LN_2;
 use std::f64::consts::PI;
+
+const FACT_TABLE: [f64; 26] = [
+    1.0,                          // 0!
+    1.0,                          // 1!
+    2.0,                          // 2!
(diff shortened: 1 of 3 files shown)


rezat-c:
what's the status of this?

nhatcher:
It's a long story. But long story short we need to do all of our transcendent function implementations and I am sure Gamma is going to be one of them. At the moment I am not working on this. If you have worked with things like this before feel free to pick this up but if not, it doesn't make sense. It's a minefield.

PR #591, opened 2025-11-25
fix: mobile issues in app

dg-ac:
This PR fixes a couple of annoying behaviours that were affecting the mobile view of the app.

---

## Changes

1. Allow triggering the title tooltip on tap, not only on hover.

image

2. Increase breakpoint to trigger the mobile view on wider screens: should make easier using the app in wider mobile displays and tablets.

image

---

## Testing

- [ ] Make sure the fixes work both on desktop and mobile

### DIFF
--- a/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
@@ -3,6 +3,7 @@ import type { Model } from "@ironcalc/workbook";
 import { IconButton, Tooltip } from "@mui/material";
 import { CloudOff, PanelLeftClose, PanelLeftOpen } from "lucide-react";
 import { useLayoutEffect, useRef, useState } from "react";
+import { MIN_MAIN_CONTENT_WIDTH_FOR_MOBILE } from "../App";
 import { FileMenu } from "./FileMenu";
 import { HelpMenu } from "./HelpMenu";
 import { downloadModel } from "./rpc";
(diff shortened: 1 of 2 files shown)


webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
```diff
@@ -77,7 +77,7 @@ export function FileBar(properties: {
           {properties.isDrawerOpen ? <PanelLeftClose /> : <PanelLeftOpen />}
         </DrawerButton>
       </Tooltip>
-      {width > 440 && (
+      {width > 768 && (
```
nhatcher: Can you use the above defined constant instead of repeating the number?

webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
```diff
@@ -92,7 +92,7 @@ export function FileBar(properties: {
           onDelete={properties.onDelete}
         />
       )}
-      {width > 440 && <HelpMenu />}
+      {width > 768 && <HelpMenu />}
```
nhatcher: Same thing here. You can export `MIN_MAIN_CONTENT_WIDTH_FOR_MOBILE` and use it here

webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
dg-ac: Good point, I've changed that. Thanks!

webapp/app.ironcalc.com/frontend/src/components/FileBar.tsx
dg-ac: Done!

PR #594, opened 2025-11-26
update: adds testing for FACT and FACTDOUBLE functions

### LINKED ISSUE #490: Please test and add documentation for new functions
Hi @elsaminsut in #489 we added some new functions please:

* Create xlsx tests for them
* Add the missing documenttaion (skeleton only)

NB: FACT and FACTDOUBLE are transcendental functions and we have implemented only an approximation to those functions.
I'll do a better approximation in the future. Here you only need to check that the approximation is not too bad.

elsaminsut:
This PR adds testing for the FACT and FACTDOUBLE functions. These include both xlsx tests with edge cases and unit tests to check the number of arguments taken by the function. It also includes an update to the main Math and Trigonometry documentation page and their individual pages to show them as now available.

This PR closes #490


nhatcher (changes requested):
Why are you removing the xlsx file?

elsaminsut:
I removed the existing xlsx test while I was testing the new one. I'll do that next.

elsaminsut:
This PR is now outdated. Will close it for now and test again when #584 (or a similar fix) is released.

PR #598, opened 2025-11-26
fix: remove duplicate xlsx tests

elsaminsut:
This PR removes some files in the test suite that were duplicates of other existing tests. Due to workflow inconsistencies, some tests were added even though equivalent ones already existed. This fix improves overall consistency. 

Bonus: this PR also fixes a few rust unit tests with missing `.rs` file formats and missing references in the `mod.rs` test file.


nhatcher:
> @nhatcher Can we merge this?

Can you check that indeed we are not removing any tests?

PR #607, opened 2025-11-28
fix/inverted-range-evaluation

iamanaws:
Fixed range evaluation with mixed absolute/relative references

Ranges like `=SUM(B$2:B4)` were being inverted to `=SUM(B4:B$2)` during 
formula re-parsing, causing them to evaluate to 0.

**Fixes:**
- Make inverted ranges work by normalizing coordinates in evaluation using min/max
- Prevent ranges from being inverted during re-parsing by only swapping 
  coordinates in A1 mode (R1C1 values can be negative, causing false swaps)

**Tests:** Added regression tests for inverted and mixed absolute/relative ranges

### DIFF
--- a/base/src/expressions/parser/mod.rs
+++ b/base/src/expressions/parser/mod.rs
@@ -649,13 +649,15 @@ impl Parser {
                 let mut absolute_row1 = left.absolute_row;
                 let mut absolute_row2 = right.absolute_row;
 
-                if row1 > row2 {
-                    (row2, row1) = (row1, row2);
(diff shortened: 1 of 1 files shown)


iamanaws:
This PR fixes both the range swap, and the range selection when inverted.

https://github.com/user-attachments/assets/38c05933-b99d-4b90-b678-1a1927f90772

nhatcher (approved):
LGTM!

Thanks. Sorry for the late review :/

Issue #629, opened 2025-12-10
WORKDAY.INTL - Wrong Weekend types


elsaminsut:
Part of  #621

**Test file:** 
[WORKDAY_WORKDAYINTL.xlsx](https://github.com/user-attachments/files/24086095/WORKDAY_WORKDAYINTL.xlsx)
[NETWORKDAYS_NETWORKDAYSINTL.xlsx](https://github.com/user-attachments/files/24146418/NETWORKDAYS_NETWORKDAYSINTL.xlsx)

When using unconventional types in the _weekend_ argument, Excel passes them in as a number and returns the calculation result but IronCalc returns a `#VALUE!` error.

This issue also affects the NETWORKDAYS.INTL function - see xlsx test.

| Cell | Formula | Excel | IronCalc | Notes | 
|---|---|---|---|---|
| I26 | `=WORKDAY.INTL(DATE(2025,1,10), 10, TRUE)` | 24/01/2025 | `#VALUE!` | Weekend is boolean | 
| I27 | `=WORKDAY.INTL(DATE(2025,1,10), 10, 1.01)` | 24/01/2025 | `#VALUE!` | Weekend is decimal number | 
| I28 | `=WORKDAY.INTL(DATE(2025,1,10), 10, 1.9999)` | 24/01/2025 | `#VALUE!` | Weekend is decimal number |

Issue #625, opened 2025-12-10
WORKDAY and WORKDAY.INTL - Incorrect boolean handling


elsaminsut:
Part of  #621

**Test file:** 
- [WORKDAY_WORKDAYINTL.xlsx](https://github.com/user-attachments/files/24086095/WORKDAY_WORKDAYINTL.xlsx)
- [NETWORKDAYS_NETWORKDAYSINTL.xlsx](https://github.com/user-attachments/files/24146432/NETWORKDAYS_NETWORKDAYSINTL.xlsx)
[DAYS_DAYS360.xlsx](https://github.com/user-attachments/files/24178811/DAYS_DAYS360.xlsx)

When the argument is boolean, Excel returns a `#VALUE!` error. IronCalc treats the input as 0 or 1 and returns a result.
This issue also affects the NETWORKDAYS.INTL function - see xlsx test.
This issue also affects the DAYS and DAYS360 functions - see xlsx test.

| Cell | Argument | Excel | IronCalc | Notes |
|---|---|---|---|---|
| C27 | `=WORKDAY(TRUE, FALSE)` |  `#VALUE!` | 1 | Boolean | 
| D27 | `=WORKDAY.INTL(TRUE, FALSE)` |  `#VALUE!` | 1 | Boolean |

elsaminsut:
Boolean handling in NETWORKDAYS and NETWORKDAYS.INTL fixed in #1340

Issue #624, opened 2025-12-10
WORKDAY and WORKDAY.INTL - Wrong error types when 0


elsaminsut:
Part of  #621

**Test file:** 
- [WORKDAY_WORKDAYINTL.xlsx](https://github.com/user-attachments/files/24086095/WORKDAY_WORKDAYINTL.xlsx)
- [NETWORKDAYS_NETWORKDAYSINTL.xlsx](https://github.com/user-attachments/files/24146428/NETWORKDAYS_NETWORKDAYSINTL.xlsx)
[DAYS_DAYS360.xlsx](https://github.com/user-attachments/files/24178811/DAYS_DAYS360.xlsx)

When using 0 as any of the arguments for WORKDAY and WORKDAY.INTL, Excel returns 0 (or the date serial number 0) whereas IronCalc returns `#NUM!`. This also applies to a blank argument, which is treated as a zero. Also incorrectly applied to error propagation and to fractional numbers between 0 and 1.

This issue also occurs in WEEKDAY, WEEKNUM and ISOWEEKNUM functions, as noted in #618 
This issue also affects the NETWORKDAYS and NETWORKDAYS.INTL functions - see xlsx test.
This issue also affects the DAYS and DAYS360 functions - see xlsx test.

| Cell | Argument | Excel | IronCalc | Notes |
|---|---|---|---|---|
| C15 | `=WORKDAY(0, 0)` | 0 | `#NUM!` | Both arguments are zero | 
| D15 | `=WORKDAY.INTL(0, 0)` | 0 | `#NUM!` | Both arguments are zero | 
| C17 | `=WORKDAY(0, 10)` |  13/01/1900 (13) | `#NUM!` | Date argument is zero | 
| D17 | `=WORKDAY.INTL(0, 10)` |  13/01/1900 (13) | `#NUM!` | Date argument is zero | 
| C22 | `=WORKDAY("", 1/0)` |  `#DIV/0!` | `#NUM!` | Zero & division by zero error | 
| D22 | `=WORKDAY.INTL("", 1/0)` |  `#DIV/0!` | `#NUM!` | Zero & division by zero error | 
| C34 | `=WORKDAY(0.00625, 1)` | 2 | `#NUM!` | Decimal between 0 and 1 | 
| D34 | `=WORKDAY.INTL(0.00625, 1)` | 2 | `#NUM!` | Decimal between 0 and 1 | 
[...]

Issue #601, opened 2025-11-26
Add testing for DAYS, DAYS360, NETWORKDAYS and NETWORKDAYS.INTL


elsaminsut:
Include: 
- [ ] xlsx test file
- [x] rust unit test
- [x] mark as implemented in docs

PR #631, opened 2025-12-13
update: adds unit test for NETWORKDAYS, NETWORKDAYS.INTL

elsaminsut:
This PR adds testing for WORKDAY and WORKDAY.INTL functions. These include unit tests to check the number of arguments taken by the function.

The xlsx tests have issues that need to be fixed, as described in #624, #625 and #629

This PR closes part of #601


nhatcher (approved):
LGTM!

PR #640, opened 2025-12-17
feat: Add extensive tests for cell behavior and styles

rezat-c:
This PR expands the test suite for the base crate, focusing on cell properties and styling.

### DIFF
--- a/base/src/test/test_cell.rs
+++ b/base/src/test/test_cell.rs
@@ -22,6 +22,8 @@ fn test_cell_get_type() {
     model._set("A12", "=1>0");
     model.evaluate();
 
+    model._set("A13", "=42"); // a CellFormula
+
     assert_eq!(model._get_cell("A1").get_type(), CellType::Text);
     assert_eq!(model._get_cell("A2").get_type(), CellType::Number);
     assert_eq!(model._get_cell("A3").get_type(), CellType::Number);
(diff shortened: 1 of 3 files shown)


base/src/test/test_cell.rs
```diff
@@ -13,13 +14,16 @@ fn test_cell_get_type() {
     model._set("A5", "1+2");
     model._set("A6", "TRUE");
     model._set("A7", "#VALUE!");
-    model._set("A8", "=Z100"); // an empty cell, considered to be a CellType::Number
```
nhatcher, line 16: Why are you removing this test?

base/src/test/test_cell.rs
```diff
@@ -32,4 +36,78 @@ fn test_cell_get_type() {
...
+
+    assert_eq!(model.get_cell_content(0, 1, 1).unwrap(), "");
+}
+
+#[test]
+fn test_from_f64_for_cell_value() {
+    // Arrange
+    let float = 42.42;
+    // Act
+    let result: CellValue = float.into();
```
nhatcher, line 92: Wait you can do this????
TIL! Thanks.

That's really useful.

nhatcher (commented):
The failing tests might not be related to your changes. I will check soon.
Many thanks!

base/src/test/test_cell.rs
rezat-c: At the moment, this line does not check the behavior of `Cell::EmptyCell`, so I added line 24, which converts cell A8 to the `Cell::EmptyCell` type, and replaced Z100 with a number so it wouldn’t be confusing.

nhatcher: @rezat-c ping

nhatcher (changes requested):
Maybe a valid test is being removed?

nhatcher (approved):
LGTM! (It's nit hurting anyone)

Issue #642, opened 2025-12-18
FIX: condition don't work correct in custom format field


rezat-c:
### How to reproduce

1. Set the value of cell **A1** to `150`.
2. Go to **Format number → Custom** and enter:
```

[>100]"High";"Low"

```

<img width="801" height="480" alt="Image" src="https://github.com/user-attachments/assets/c01ab2fe-797e-452d-8c68-e796d37849c1" />

### Expected behavior

`High`

### Actual behavior

`#VALUE!`

### Notes

This happens because the closing bracket (`]`) is not consumed in `consume_condition`.

If you enter the format without the closing bracket, for example:
```

[>100"High";"Low"

```
it works as expected.

<img width="799" height="478" alt="Image" src="https://github.com/user-attachments/assets/aba7257b-9b4d-45ba-be23-132b50ae5833" />

rezat-c:
Also, if you enter a value ≤ 100 in A1, it still displays `High`. `Low` is shown only when the value is less than zero.

PR #643, opened 2025-12-18
Fix: Prevent closing bracket being swallowed in consume condition

rezat-c:
This PR corrects a bug in the consume_condition function that would incorrectly swallow a closing bracket.
Additional tests have been added to cover both successful and failed parsing conditions.

#642.


nhatcher (approved):
Well done. Thank you!

LGTM!

Issue #26, opened 2024-03-26
Create "Fill Data Automatically" feature


nhatcher:
Most spreadsheets applications let's you fill data automatically:

Number series, weekdays and short name of weekdays and same for months

* 1,2, 3, 4 => 5, 6, 7, 8, ...
* Monday, Tuesday => Wednesday, Thursday, ...
* Jan, Feb => Mar, ..
* 1, 3, 5 => 7, 9, 11

Note that this will depend on localization features.

nhatcher:
Not really a fan of putting lots of code in IronCalc but this might be a good opportunity to try some TinyML

jaycarlton:
I'm not following how we need machine leanring for this, unless you want to pick up on patterns that aren't linear/discrete.

pyshx:
Consider implementing sequence detection (like Excel's auto-fill) with a path to grow into program synthesis. Starting simple but with room to evolve.

## Stages

### Stage 1: Basic Detection
Just handle common patterns first:
```rust
// Simple numbers and days
"1,2,3" -> next is 4
"Mon,Tue" -> next is Wed
"Project1,Project2" -> next is Project3
```

### Stage 2: Pattern Matching
Start recognizing pattern structure:
```rust
// Detect steps and formats
"1,3,5" -> Pattern(start=1, step=2)
"Jan,Feb" -> Pattern(start=Jan, step=1, format=Month)
```

### Stage 3: Simple Programs  
Move towards program-like structure:
```rust
// Q1,Q2,Q3 becomes something like:
Program {
    start: 1,
    step: 1,
    format: "Q{}"
}
```

## References
- Based on approaches from [Flash Fill paper](https://dl.acm.org/doi/pdf/10.1145/3571226)
- [Sequence-Specific Work](https://people.csail.mit.edu/rishabh/papers/cacm12.pdf)

Thoughts?

PR #666, opened 2026-01-01
Feat: Enhance autofill functionality

rezat-c:
This pull request enhances the autofill functionality by introducing arithmetic progression detection; support for floating-point numbers, suffixed number, weekdays and short name of weekdays and same for months 

Depend on localization features.
Covered by tests. 

* 1,2, 3, 4 => 5, 6, 7, 8, ...
* 1, 3, 5 => 7, 9, 11
* Project1, Project2 -> Project3
* 1.5, 2.5, 3.5 => 4.5, 5.5
* Monday, Tuesday => Wednesday, Thursday, ...
* Jan, Feb => Mar, ..

#26

### DIFF
--- a/base/src/user_model/sequence_detector.rs
+++ b/base/src/user_model/sequence_detector.rs
@@ -41,7 +41,19 @@ impl DateProgression {
 
 fn round_sig(value: f64) -> f64 {
     // rounding up to 15 significant figures
-    format!("{:.14e}", value).parse().unwrap_or(value)
+    if value == 0.0 {
+        return 0.0;
+    }
+    let rounded = value.round();
+    if (value - rounded).abs() <= 1e-11 * value.abs().max(1.0) {
+        return rounded;
+    }
+    let sign = value.signum();
+    let abs_value = value.abs();
+    let exponent = abs_value.log10().floor();
+    let normalized = abs_value / 10.0_f64.powf(exponent);
+    let rounded_normalized = (normalized * 1e14).round() / 1e14;
+    sign * rounded_normalized * 10.0_f64.powf(exponent)
 }
 
 pub(crate) enum Progression {


base/src/user_model/sequence_detector.rs
```diff
@@ -0,0 +1,213 @@
...
+        if value.len() == potential_numeric_suffixes.len() {
+            0
+        } else {
+            potential_numeric_suffixes.len()
+        }
+    }
+}
+
+impl SequenceDetector for SuffixedNumberDetector {
+    fn detect(&self, values: &[String]) -> Option<Progression> {
```
nhatcher, line 235: This might get a bit tricky with floating point numbers:

image

base/src/user_model/sequence_detector.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn detect(&self, values: &[String]) -> Option<Progression>;
+}
+
+struct NumericProgressionDetector;
+
+impl SequenceDetector for NumericProgressionDetector {
+    fn detect(&self, values: &[String]) -> Option<Progression> {
+        values
+            .iter()
+            .map(|s| s.parse::<f64>())
```
nhatcher: Also you will have to have the locale into account

base/src/user_model/sequence_detector.rs
```diff
@@ -0,0 +1,213 @@
...
+impl SuffixedNumberDetector {
+    fn suffix_index(value: &str) -> usize {
+        let mut rev = String::new();
+
+        let potential_numeric_suffixes = value
+            .chars()
+            .rev()
+            .map_while(|x| {
+                rev.push(x);
+                rev.parse::<f64>().ok()
```
nhatcher: Also you will have to have the locale into account here?

base/src/user_model/sequence_detector.rs
```diff
@@ -0,0 +1,213 @@
+use crate::locale::Locale;
```
nhatcher, line 1: I think this file lends itself to a bunch of unit tests for each of the detectors and the final one.

base/src/user_model/sequence_detector.rs
nhatcher: I _think_ in this "sequence detector" we should only consider integers:

image

base/src/user_model/sequence_detector.rs
rezat-c: Every detector is covered by tests; only some regions are not

base/src/user_model/sequence_detector.rs
rezat-c: Excel actually behaves strangely here; I think it shouldn’t work like this
I would expect 2.8, 2.9 ⇒ 3

image

base/src/user_model/sequence_detector.rs
nhatcher: This file is very easily testable by unit tests:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn numeric_progression_basic() {
        let p = NumericProgression { last: 3.0, step: 2.0 };
        assert_eq!(p.next(0), 5.0);
        assert_eq!(p.next(1), 7.0);
    }

    #[test]
    fn suffixed_progression_basic() {
        let values = vec!["A1", "A2", "A3"]
            .into_iter().map(String::from).collect::<Vec<_>>();

        let prog = SuffixedNumberDetector.detect(&values).unwrap();
        assert_eq!(prog.next(0), "A4");
        assert_eq!(prog.next(1), "A5");
    }

    #[test]
    fn numeric_detector_rejects_non_progression() {
        let values = vec!["1", "2", "4"]
            .into_iter().map(String::from).collect::<Vec<_>>();
        assert!(NumericProgressionDetector.detect(&values).is_none());
    }
}
```

What you have (and we need, those are good) are more generic tests.
Tests in this file would serve also for documentation and are easily hackable.

base/src/user_model/sequence_detector.rs
nhatcher: In `Project2.8` the stem is `Project2.` and not `Project`. Excel only deals with integers as suffix and we should do the same.
In column C you have strings also, the stems being `2,`

base/src/user_model/sequence_detector.rs
```diff
@@ -0,0 +1,693 @@
...
+        let progression = detector.detect(&values).unwrap();
+        assert_eq!(progression.next(0), "3000,5");
+
+        let values = vec!["1,7".to_string(), "1,8".to_string(), "1,9".to_string()];
+        let progression = detector.detect(&values).unwrap();
+        assert_eq!(progression.next(0), "2");
+        assert_eq!(progression.next(1), "2,1");
+    }
+
+    // need to add the Indian locale
```
nhatcher: Please remove this test, there is no point in having tests for non existing code

base/src/user_model/sequence_detector.rs
```diff
@@ -0,0 +1,693 @@
...
+        assert_eq!(p.next(0), "May");
+
+        let values = vec!["1".to_string(), "A".to_string(), "foo".to_string()];
+        assert!(detect_progression(&values, locale).is_none());
+
+        let values = vec!["1".to_string(), "2".to_string()];
+        let p = detect_progression(&values, locale).unwrap();
+        match p {
+            Progression::Numeric(_) => {}
+            _ => unreachable!(),
```
nhatcher: You can use `assert!(matches!` but please don't use `unreachable!`

base/src/test/user_model/test_autofill_rows.rs
```diff
@@ -2,15 +2,16 @@
...
 use crate::test::util::new_empty_model;
 use crate::UserModel;
 
 #[test]
 fn basic_tests() {
     let model = new_empty_model();
     let mut model = UserModel::from_model(model);
-    // This is cell A3
-    model.set_user_input(0, 3, 1, "alpha").unwrap();
+    model.set_user_input(0, 3, 1, "alpha").unwrap(); // A3
```
nhatcher, line 12: why did you change this test? If you need to do this kind of changes, can you do them in a separate PR?
It makes reading the diffs in this PR much more difficult

base/src/user_model/sequence_detector.rs
rezat-c: There are about seven other tests that are ignored, and I decided it was better to leave this one as is, because at the moment the code responsible for handling non-standard grouping is not tested, the field in `Locale` cannot be changed, and creating it manually would be too cumbersome.

base/src/user_model/sequence_detector.rs
rezat-c: I didn’t know that, thanks!

base/src/test/user_model/test_autofill_rows.rs
rezat-c: I modified the test because there was too much duplicated code, and I decided it would be better to reduce it.

base/src/user_model/sequence_detector.rs
rezat-c: So, can I use `panic!()` in tests?

base/src/test/user_model/test_autofill_rows.rs
nhatcher: In tests it is completely ok to have duplicated code. But even if you would want to reduce that it would be nicer to do in a separate PR where I can see easily which tests are being deleted and if we are loosing some infomration or not

base/src/user_model/sequence_detector.rs
nhatcher: I wouldn't use `panic!` or `unreachable!` in tests although you absolutely can, the test will fail.

base/src/user_model/sequence_detector.rs
nhatcher: Having untested code is always problematic and in the long run it hurst more than it helps.

PR #667, opened 2026-01-03
update: adds testing for inverted ranges

### LINKED ISSUE #657: Please test "inverted" ranges
Please test that in every circumstance "inverted" ranges `=SUM(C5:A:1)` behave as expected.

* Don't break anything
* They behave consistently
* Are we compatible with Excel?

elsaminsut:
This PR adds testing to make sure inverted ranges behave exactly as their counterparts, so that `SUM(A3:A1)` returns the same value as `SUM(A1:A3)`. The tests include:
- Different sized ranges: vertical, horizontal and rectangular
- Absolute references, like `SUM(A$3:A$1)`
- Ranges with blank and error cells
- Various functions: AVERAGE, COUNT, COUNTA, COUNTIF, INDEX, MATCH, SUMIF, CONCATENATE, ROWS, COLUMNS
- Mixed data types

It also includes a small fix to add a missing file extension for a test and its missing reference to the module tree.

This PR closes #657

### DIFF
--- a/base/src/test/test_inverted_ranges.rs
+++ b/base/src/test/test_inverted_ranges.rs
@@ -13,6 +13,13 @@ fn sum_inverted_range_vertical() {
     model._set("B2", "=SUM(A3:A1)");
     model.evaluate();
 
+    // Formula should not be inverted
+    assert_eq!(model._get_formula("B2"), *"=SUM(A1:A3)");
+
+    // Result check
+    assert_eq!(model._get_text("B1"), *"6");
+
+    // Result should be the same
     assert_eq!(model._get_text("B1"), model._get_text("B2"));
 }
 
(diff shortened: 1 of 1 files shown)


base/src/test/test_inverted_ranges.rs
```diff
@@ -0,0 +1,182 @@
...
+    let mut model = new_empty_model();
+    model._set("A1", "1");
+    model._set("A2", "2");
+    model._set("A3", "3");
+
+    model._set("B1", "=SUM(A1:A3)");
+    model._set("B2", "=SUM(A3:A1)");
+    model.evaluate();
+
+    assert_eq!(model._get_text("B1"), model._get_text("B2"));
```
nhatcher, line 23: I _think_ it would be nicer:

* check that the formula is correctly rewritten
* check that the formula evaluates correctly

elsaminsut:
Done! Please review

nhatcher (approved):

PR #668, opened 2026-01-04
update: add specific context menus to column and row headers

dg-ac:
This PR introduces **dedicated context menus** for column and row headers, allowing header-specific actions to be surfaced more clearly and consistently.  

Previously, header interactions relied on the generic cell context menu, which included options that were not always relevant.

### Changes

1. Optimized `CellContextMenu`: cleaned up duplicated code, improved anchor positioning, moved all menus to their own folder.

image

2. Added a **specific context menu for column headers** with column-related actions. Row actions are not present in this menu. 'Unfreeze' action only appears when there are frozen columns.

image

3. Added a **specific context menu for row headers** with row-related actions.  Column actions are not present in this menu. 'Unfreeze' action only appears when there are frozen rows.

image

---

### Testing

- [ ] Right-click on a **column header** and verify the column-specific menu appears.  
- [ ] Right-click on a **row header** and verify the row-specific menu appears.  
- [ ] Ensure the generic cell context menu still works correctly on regular cells.  
- [ ] Confirm that menu actions apply to the correct row or column.

### DIFF
--- a/webapp/IronCalc/src/components/Worksheet/ContextMenus/RowHeaderContextMenu.tsx
+++ b/webapp/IronCalc/src/components/Worksheet/ContextMenus/RowHeaderContextMenu.tsx
@@ -1,0 +1,132 @@
+import { ArrowDown, ArrowUp, Plus, Snowflake, Trash2 } from "lucide-react";
+import { useTranslation } from "react-i18next";
+import { theme } from "../../../theme";
+import {
+  DeleteButton,
+  ItemNameStyled,
+  MenuDivider,
+  StyledMenu,
+  StyledMenuItem,
+} from "./CellContextMenu";
+
+const red_color = theme.palette.error.main;
+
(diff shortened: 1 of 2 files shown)


webapp/IronCalc/src/components/Worksheet/ContextMenus/CellContextMenu.tsx
```diff
@@ -10,14 +10,14 @@ import {
...
-import { theme } from "../../theme";
+import { theme } from "../../../theme";
 
 const red_color = theme.palette.error.main;
 
 interface CellContextMenuProps {
   open: boolean;
   onClose: () => void;
-  anchorEl: HTMLDivElement | null;
+  anchorPosition: { top: number; left: number } | null;
```
nhatcher, line 20: Do we need `null`?

Issue #623, opened 2025-12-10
Failure in WORKDAY.INTL causes chrono overflow


elsaminsut:
## Summary
When weekend_mask has all 7 values set to true (meaning all days are weekends), the function enters an infinite loop because it can never find a workday to count. 

## Test
Write `'=WORKDAY.INTL(DATE(2025,1,10),10,"1111111")` into any cell. The first two arguments can be anything as long as the weekend pattern is a string with 7 ones.

## Solution
Need to add validation before the loop to check if at least one day in the week is a workday, and return #VALUE! if not.

PR #670, opened 2026-01-04
fix: WORKDAY.INTL won't overflow if all days are weekends

elsaminsut:
This PR addresses issue #623, where an edge case with a weekend mask of `1111111` would result in an overflow of the chrono crate. It adds an initial check for that weekend mask and results in a `#VALUE!` error if found. It also adds a small test to check the changes work as they should.

### DIFF
--- a/base/src/functions/date_and_time.rs
+++ b/base/src/functions/date_and_time.rs
@@ -1561,7 +1561,7 @@ impl Model {
         };
 
         // Checks if all days are weekends
-        if weekend_mask == [true, true, true, true, true, true, true] {
+        if weekend_mask == [true; 7] {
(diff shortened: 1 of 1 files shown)


base/src/functions/date_and_time.rs
```diff
@@ -1559,6 +1559,16 @@ impl Model {
             Ok(h) => h,
             Err(e) => return e,
         };
+
+        // Checks if all days are weekends
+        if weekend_mask == [true, true, true, true, true, true, true] {
```
nhatcher: What about `[true; 7]`?

nhatcher (changes requested):

PR #683, opened 2026-01-12
Workbook: don't bundle dependencies, only build ES module

jvdp:
This drastically reduces bundle size (which is always nice) but also allows Emotion's `CacheProvider` and `ThemeProvider` to affect the workbook component structure. (The context variable name is probably renamed to something shorter otherwise by Rollup.)

This seems to be the only thing getting in the way of rendering the workbook in a shadow DOM.


nhatcher:
Hmmm, that's a tough sell. Let me think about this.

Issue #735, opened 2026-02-02
CHISQ functions should return `#NUM!` for DOF > 10E10


elsaminsut:
## Summary
According to the [Excel documentation](https://support.microsoft.com/en-us/office/chisq-dist-function-8486b05e-5c05-4942-a9ea-f6b341518732) for the CHISQ.DIST functions, the argument _deg_freedom_ has an upper bound of 10^10. Anything above that value for that argument should return a `#NUM!` error but IronCalc returns a calculated result instead.

## Test File
[CHISQ.xlsx](https://github.com/user-attachments/files/25026521/CHISQ.xlsx)

## Reproduction
The following examples show the failing cases:

| Cell | Formula | Excel | IronCalc | Notes | 
|---|---|---|---|---|
| D39 | `=CHISQ.DIST(10, 1E12, TRUE)` | `#NUM!` | 0 | DOF > 10E10 |
| E39 | `=CHISQ.INV(D39, 1E12)` | `#NUM!` | 0 | DOF > 10E10 |
| C34 | `=CHISQ.DIST.RT(10, 1E12)` | `#NUM!` | 1 | DOF > 10E10 |
| D34 | `=CHISQ.INV.RT(C34, 1E12)` | `#NUM!` | 0 | DOF > 10E10 |

PR #737, opened 2026-02-03
fix: CHISQ functions now handle booleans and deg_freedom has an upper bound

### LINKED ISSUE #734: CHISQ functions should handle boolean values
Similarly to #672, CHISQ functions should treat booleans as their respective numeric value. IronCalc instead treats them as just booleans, returning a #VALUE! error with a message "Expecting number".

## Test File
[CHISQ.xlsx](https://github.com/user-attachments/files/25026521/CHISQ.xlsx)

## Reproduction
The following examples show the failing cases:

| Cell | Formula | Excel | IronCalc | Notes | 
|---|---|---|---|---|
| D69 | `=CHISQ.DIST(TRUE, 7, TRUE)` | 0.99954704... | `#VALUE!` | Values as booleans |
| E69 | `=CHISQ.INV(D69, TRUE)` | 1 | `#VALUE!` | Values as booleans |
| C52 | `=CHISQ.DIST.RT(12.3, TRUE)` |  0.000452958... | `#VALUE!` | Values as booleans |
| D52 | `=CHISQ.DIST.RT(C52, TRUE)` |  12.2999999... | `#VALUE!` | Values as booleans |

elsaminsut:
This PR fixes #734 and #735. Now the set of CHISQ functions (CHISQ.DIST, CHISQ.INV, CHISQ.DIST.RT and CHISQ.INV.RT) can handle boolean values, following the behavior in Excel. It also adds an upper domain limit (10^10) for the _degrees of freedom_ argument. 

Some unit tests have been added as well to check the fix works as intended.

### DIFF
--- a/base/src/functions/statistical/chisq.rs
+++ b/base/src/functions/statistical/chisq.rs
@@ -5,6 +5,7 @@ use crate::expressions::types::CellReferenceIndex;
 use crate::{
     calc_result::CalcResult, expressions::parser::Node, expressions::token::Error, model::Model,
 };
+const MAX_DEGREES_OF_FREEDOM: f64 = 10_000_000_000.0;
 
 impl<'a> Model<'a> {
     // CHISQ.DIST(x, deg_freedom, cumulative)
@@ -36,7 +37,7 @@ impl<'a> Model<'a> {
             );
         }
         // if degrees of freedom < 1 or > 10^10 → #NUM!
-        if !(1.0..=10000000000.0).contains(&df) {
+        if !(1.0..=MAX_DEGREES_OF_FREEDOM).contains(&df) {
             return CalcResult::new_error(
                 Error::NUM,
                 cell,
@@ -99,7 +100,7 @@ impl<'a> Model<'a> {
         }
 
         // if degrees of freedom < 1 or > 10^10 → #NUM!
-        if !(1.0..=10000000000.0).contains(&df) {
+        if !(1.0..=MAX_DEGREES_OF_FREEDOM).contains(&df) {
             return CalcResult::new_error(
                 Error::NUM,
                 cell,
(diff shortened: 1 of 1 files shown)


base/src/test/statistical/test_fn_chisq.rs
```diff
@@ -121,8 +133,10 @@ fn test_fn_chisq_inv_rt_smoke() {
     // probability < 0 or > 1 -> #NUM!
     model._set("A5", "=CHISQ.INV.RT(-0.1, 4)");
     model._set("A6", "=CHISQ.INV.RT(1.1, 4)");
-    // deg_freedom < 1 -> #NUM!
+    // deg_freedom < 1 or > 10^10 -> #NUM!
     model._set("A7", "=CHISQ.INV.RT(0.5, 0)");
+    model._set("A8", "=CHISQ.INV.RT(1, 10000000000)");
```
nhatcher, line 138: You should use the group separator also in tests

base/src/test/statistical/test_fn_chisq.rs
```diff
@@ -54,8 +58,10 @@ fn test_fn_chisq_dist_rt_smoke() {
     // Domain errors
     // x < 0 -> #NUM!
     model._set("A5", "=CHISQ.DIST.RT(-1, 4)");
-    // deg_freedom < 1 -> #NUM!
+    // deg_freedom < 1 or > 10^10 -> #NUM!
     model._set("A6", "=CHISQ.DIST.RT(0.5, 0)");
+    model._set("A7", "=CHISQ.DIST.RT(0, 10000000000)");
+    model._set("A8", "=CHISQ.DIST.RT(0, 10000000001)");
```
nhatcher, line 64: This can be `MAX_FREEDOM_NUMBER+1`

nhatcher: Also I think this needs to be tested in Excel. Is there any reason why we do this? (just curious)

elsaminsut: The xlsx test failed, that's how I found the bug. In the documentation, this upper bound is explicitly mentioned and I wanted to test whether IronCalc had it or not (turns out it didn't!).

Will fix readability in the tests in my next PR.

Issue #295, opened 2025-02-23
Feature: Commenting


dg-ac:
TBD, waiting for spec and designs.

PR #740, opened 2026-02-04
initial_draft_threaded_comments

deanm0000:
working on https://github.com/ironcalc/IronCalc/issues/295

I did a drive-by refactor on the `load_sheet_rels` function. It was both returning a value and mutating an input. To keep from iterating on the same rel files twice, I needed to add my bit to it which meant another thing it'd be mutating along the way. Rather than having it mutate two variables while returning values for the third, I just made it mutate 3 variables and have no value returned.

Progress:
Made structs/enums for threaded comments and its related fields.
Added those structs to Workbook and Worksheet
Parses the xml of xlsx file to create them

To Do:
add methods for users to read, edit, add threaded comments
save threaded comments back to a file

Stumbling blocks:
I couldn't figure out how to do Encode, Decode on NaiveDateTime so that's just stored as a String. To get Encode, Decode to work on uuid I had to add features to

### DIFF
--- a/base/Cargo.toml
+++ b/base/Cargo.toml
@@ -13,13 +13,15 @@ readme = "README.md"
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
 ryu = "1.0"
-chrono = "0.4"
+chrono = { version="0.4", features=["serde"] }
 chrono-tz = "0.10"
 regex = { version = "1.0", optional = true}
 regex-lite = { version = "0.1.6", optional = true}
-bitcode = "0.6.8"
+bitcode = {version = "0.6.8", features = ["uuid"] }
 csv = "1.3.0"
 statrs = { version = "0.18.0", default-features = false, features = [] }
+uuid = { version = "1.2.2", features = ["serde", "v4"] }
+
 
 [features]
 default = ["use_regex_full"]
--- a/base/Cargo.toml
+++ b/base/Cargo.toml
@@ -17,7 +17,7 @@ chrono = { version="0.4", features=["serde"] }
 chrono-tz = "0.10"
 regex = { version = "1.0", optional = true}
 regex-lite = { version = "0.1.6", optional = true}
-bitcode = {version = "0.6.8", features = ["uuid"] }
+bitcode = {version = "0.6.8", features = ["uuid", "serde"] }
 csv = "1.3.0"
 statrs = { version = "0.18.0", default-features = false, features = [] }
 uuid = { version = "1.2.2", features = ["serde", "v4"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,6 +86,7 @@ dependencies = [
  "bytemuck",
  "glam",
  "serde",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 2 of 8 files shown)


base/Cargo.toml
```diff
@@ -13,13 +13,15 @@ readme = "README.md"
...
-chrono = "0.4"
+chrono = { version="0.4", features=["serde"] }
 chrono-tz = "0.10"
 regex = { version = "1.0", optional = true}
 regex-lite = { version = "0.1.6", optional = true}
-bitcode = "0.6.8"
+bitcode = {version = "0.6.8", features = ["uuid", "serde"] }
 csv = "1.3.0"
 statrs = { version = "0.18.0", default-features = false, features = [] }
+uuid = { version = "1.2.2", features = ["serde", "v4"] }
```
nhatcher, line 23: We shouldn't be adding more dependencies. We remove dependencies :)

Tjis is a great project to work on! But we definitely should sync before doing large amounts of work.
Comments are related to two very important and complex tasks:
* Users
* Collaboration 

We need to make sure everything makes sense

deanm0000: I didn't add bitcode, just features. I added uuid because the fields _are_ UUIDs. ~~When adding (I didn't do that yet) new threaded comments they'll need new UUIDs so I can't even really say to just store them as String unless you also want to DIY making random UUIDs without rand.~~ (edit: I see you have rand under non-wasm build with a random function for both wasm/non-wasm builds so I now see that's doable)

On the bigger picture, I'm in full agreement that it all needs to make sense. I think what I've done is mostly boiler plate and leaves plenty of room to change things. I'll await further feedback before proceeding.

One self critique I have is storing a vec for threadedcomments at the worksheet level rather than perhaps storing it in the Cell struct or, frankly, at a minimum, having it be a hashmap with the cell address as the key.

deanm0000: >perhaps storing it in the Cell struct

On second thought, that's probably bad since it means there'd be an empty Vec attached to most cells which would needlessly waste memory.

> having it be a hashmap with the cell address as the key.

I think this is best, having a field in the `Worksheet` as a `HashMap<(row, column), Vec<ThreadedComment>>`

One potential footgun is that in the spec the ref attribute (which cell holds the comment) is marked as optional so the above doesn't account for that. I see that in other places, (ie. SheetData) that row and column indices are i32 rather than u32 so perhaps comments with no ref could be saved at (-1, n) where n is just a counter.

deanm0000:
Hi @nhatcher just wanted to check back in on this. I know you mentioned not wanting to pick up a dependency which isn't too hard to work around. Did you have any other feedback or direction besides that?

nhatcher:
> Hi @nhatcher just wanted to check back in on this. I know you mentioned not wanting to pick up a dependency which isn't too hard to work around. Did you have any other feedback or direction besides that?

Hi @deanm0000, no sorry. I think this is tough at the moment and blocked since we need to understand and design how we want to have users in the system. There is already a concept of "view" that I _think_ is wrong and we need to retire. I am afraid we need to go back to the drawing board for this.

Does that make sense?

Issue #753, opened 2026-02-08
parse_number() panics on lone '+' or '-' due to out-   of-bounds index


rezat-c:
parse_number() (and callers) panic when value is just '-' or '+'. 

The code increments position after reading the sign and then immediately indexes bytes[position] without checking bounds, so a single-character input causes out-of- bounds access.

```
    // get the sign
    let sign = if bytes[0] == b'-' {
        position += 1;
//               ^^^
        -1.0
    } else if bytes[0] == b'+' {
        position += 1;
//               ^^^
        1.0
    } else {
        1.0
    };

    if bytes[position] == group_separator {
//               ^^^
        return Err("Cannot parse number".to_string());
    }

```

Issue #752, opened 2026-02-08
Formula with special character after "=" returns #NAME? instead of text


rezat-c:
When entering a formula like =x where x is a special character (e.g., $ or =), returns #NAME?. 

Excel treats this input as plain text and displays it instead. 

<img width="402" height="526" alt="Image" src="https://github.com/user-attachments/assets/03de840a-3d43-4e05-8eae-dccece1a7706" />

<img width="769" height="637" alt="Image" src="https://github.com/user-attachments/assets/09496626-aa22-4f9c-aa70-4ece82b2ca98" />

Issue #741, opened 2026-02-06
Apostrophe prefix not preserved when saving cell values


rezat-c:
When entering a value like '01, the apostrophe is stripped during saving, though the value appears correctly initially. The apostrophe should be preserved to maintain the string formatting intent.

Expected behavior: The apostrophe prefix should be saved and persisted with the cell value.
Actual behavior: The apostrophe is lost on save, so '01 becomes 01, breaking the intended string formatting and causing subsequent edits to lose the leading zero.

<img width="452" height="422" alt="Image" src="https://github.com/user-attachments/assets/9b1aef31-fc4c-4e53-86b3-2c6a463430f8" />

elsaminsut:
I just tested the fix and it only seems to work for numbers and formulas (and booleans). If the cell content is a date or a string, the quote prefix is not preserved: https://testing.ironcalc.com/?model=MgM9U-sNw4Q-A8sUw

nhatcher:
Fix, and fixed deployed, please test

PR #754, opened 2026-02-08
Accept + and - as formula prefixes

### LINKED ISSUE #726: Accept `+` and `-` as formula prefixes (Excel compatibility)
## Summary
IronCalc currently requires formulas to start with `=`. In Microsoft Excel, users can start formulas with `+` or `-`. Excel automatically normalizes these inputs to a standard `=`-prefixed formula.

Supporting `+` as a formula prefix would improve Excel parity and reduce friction for:
- Users migrating sheets or copying from older workbooks.
- Financial/Accounting users who rely heavily on the Numpad.

## Current Behavior
Entering a formula that starts with `+` (e.g., `+SUM(1,2)` or `+A1+B1`) is not treated as a formula. It is interpreted as a text string or simply fails to evaluate.

## Expected Behavior
The parser should treat a leading `+` (or `-`) as a formula indicator, equivalent to entering the same string with `=`:

| Input | Normalizes To | Result |
| :--- | :--- | :--- |
| `+1+2` | `=1+2` | `3` |
| `+SUM(A1:A3)` | `=SUM(A1:A3)` | Sum of range |
| `+A1+B1` | `=+A1+B1` | Sum of cells |

## Proposed Implementation (High Level)
In the user-input layer (UI or API), detect the prefix before parsing:
1.  If the first non-whitespace character is `+`, convert `+<expr>` → `=<expr>` (or `=+<expr>` to preserve Excel's exact normalization).
2.  Ensure the serialized representation remains `=`-prefixed for `.xlsx` export consistency.

## Edge Cases to Handle
- **Text Escaping:** Text starting with `+` should still be possible using the apostrophe prefix. `'+123` should remain text `+123`.
- **Single Values:** `+1` should evaluate to `1` (Excel treats this as a formula entry).
- **Whitespace:** `+ SUM(A1)` should ideally be normalized to `=SUM(A1)`.

## Acceptance Criteria
- [ ] `+`-prefixed formulas evaluate identically to the same formula with `=`
- [ ] Works in web app cell edit and formula bar
- [ ] Export/import round-trip does not change computed results
- [ ] Unit tests added for mixed operators (e.g., `+A1-B1`)

## References
- Microsoft Learn: [Excel expression evaluation compatibility](https://learn.microsoft.com/en-us/answers/questions/4772156/in-excel-formula-what-does-adding-mean)

rezat-c:
Allow formulas that begin with +/- to parse as formulas
 
Note that there are several differences from Excel behavior:

1. If you enter '+foo in a cell, then copy and paste it, you now get an error because the ' prefix is not preserved. See #741.
2. Any cells that start with + or - but are not formulas (e.g., ++, +-, +=2) now return an error. See #752.
3. In cases like:

+A1+A2
+4+2
+B
+B1

the leading + is removed, whereas Excel preserves it when the next character is not a digit:

=+A1+A2
=4+2
=+B
=+B1

Close #726, #753

### DIFF
--- a/base/src/model.rs
+++ b/base/src/model.rs
@@ -280,17 +280,22 @@ impl<'a> Model<'a> {
         }
     }
 
-    fn is_formula_with_prefix(&self, value: &str) -> bool {
+    fn formula_without_prefix<'b>(&self, value: &'b str) -> Option<&'b str> {
         if let Some(stripped) = value.strip_prefix('=') {
-            return !stripped.is_empty();
-        }
-        if let Some(stripped) = value.strip_prefix(['+', '-']) {
-            if stripped.is_empty() || self.cast_number(stripped).is_some() {
-                return false;
+            if stripped.is_empty() {
+                None
+            } else {
+                Some(stripped)
             }
-            return true;
+        } else if let Some(stripped) = value.strip_prefix(['+', '-']) {
+            if stripped.is_empty() || self.cast_number(stripped).is_some() {
+                None
+            } else {
+                Some(value)
+            }
+        } else {
+            None
         }
-        false
     }
 
     pub(crate) fn evaluate_node_in_context(
@@ -1089,14 +1094,7 @@ impl<'a> Model<'a> {
             .worksheet(target.sheet)
             .map_err(|e| format!("Could not find target worksheet: {e}"))?
             .get_name();
-        if self.is_formula_with_prefix(value) {
-            let formula;
-            if let Some(eq_formula) = value.strip_prefix('=') {
-                formula = eq_formula
-            } else {
-                formula = value;
-            }
-
+        if let Some(formula) = self.formula_without_prefix(value) {
             let cell_reference = CellReferenceRC {
                 sheet: source_sheet_name.to_owned(),
                 row: source.row,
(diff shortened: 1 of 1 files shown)


base/src/utils.rs
```diff
@@ -127,7 +127,7 @@ impl ParsedReference {
 ///  * a boolean
 ///  * an error (i.e "#VALUE!")
 pub(crate) fn value_needs_quoting(value: &str, language: &Language) -> bool {
-    value.starts_with('=')
+    value.starts_with(['=', '+', '-'])
```
rezat-c, line 130: tests pass even without this change

base/src/formatter/format.rs
```diff
@@ -929,6 +929,10 @@ fn parse_number(
         1.0
     };
 
+    if position >= len {
```
nhatcher, line 932: Why is this needed? Is is a different bugfix?

base/src/model.rs
```diff
@@ -1076,7 +1089,14 @@ impl<'a> Model<'a> {
...
             .map_err(|e| format!("Could not find target worksheet: {e}"))?
             .get_name();
-        if let Some(formula) = value.strip_prefix('=') {
+        if self.is_formula_with_prefix(value) {
+            let formula;
+            if let Some(eq_formula) = value.strip_prefix('=') {
+                formula = eq_formula
+            } else {
+                formula = value;
+            }
```
nhatcher: This few lines of code are repeated every time. Do you think we can factor them in the `is_formula_with_prefix` ?

base/src/formatter/format.rs
rezat-c: this is close #753

base/src/model.rs
rezat-c: done

rezat-c:
> Hi @rezat-c , I think things are a tad more complex. Many inputs that start with `+` or `-` are not necessarily a formula
> 
> `--------` `+++++`
> 
> for instance. It would be nice to figure out the correct pattern before hand.

Yes, I mentioned this issue in the PR comment and in #752.

Issue #756, opened 2026-02-09
Precision issues in GCD, LCM


elsaminsut:
The GCD and LCM functions should reject negative arguments. However, very small negative values (like -1E-10) are not being detected as negative, probably due to floating-point precision issues during parsing. This behavior is similar to issue #676.

## Test File
[LCM_GCD.xlsx](https://github.com/user-attachments/files/25196605/LCM_GCD.xlsx)

## Reproduction
The following examples show the failing cases:

| Cell | Formula | Excel | IronCalc | Notes | 
|---|---|---|---|---|
| D23 | `=LCM(-0.00037, 0.0001)` | `#NUM!` | 0 | Near zero, negative |
| E23 | `=GCD(-0.00037, 0.0001)` | `#NUM!` | 0 | Near zero, negative |

Issue #551, opened 2025-11-14
Edge case failures in LCM and GCD


elsaminsut:
## Summary
Running tests with `cargo test` in the IronCalc engine shows several edge case failures in the LCM and GCD functions using the workbook `LCM_GCD_initial.xlsx`. The results differ from Excel, and the load/save verification also fails for this file.

## Test File
[LCM_GCD.xlsx](https://github.com/user-attachments/files/23552622/LCM_GCD.xlsx)

### Type mismatch and error propagation
Several cells return errors or incorrect types as compared to Excel:
  - Expected: Numeric output (Excel)
  - Actual: `#VALUE!` / `#NUM!` / `#N/A` errors instead of numbers
  - Issue: Should return numbers, but errors are returned, affecting dependent cells.

| Cell(s) | Arguments | Actual | Expected | Description |
| --- | --- | --- | --- |--- |
| D17, E17 | `10`, `=""` | #VALUE! | 10 | Blank with quotes |
| D27, E27 | `eight`, `2` | #VALUE! | 2 | Mix number/string |
| D29, E29 | `="@"`, `5` | #VALUE! | 5 | Symbol |
| D33, E33 | `=" 2024-01-10 "`, `7` | #VALUE! | 7 | Non coerced text date |
| D35, E35 | `=LCM(,B35,C35)` | #N/A | 5 | Blank argument |
| D45, E45 | `TRUE`, `2` | #VALUE! | 2 | Mix number/bool |
| D46, E46 | `FALSE`, `2` | #VALUE! | 2 | Mix number/bool |
| D52, E52 | `Banana`, `1` | #VALUE! | 1 | Mix number/string|

### Different numbers returned by formulas
Some outputs are numerically different from those provided by Excel. Note that `GCD` (in column E) returns the correct value for some cases.
  - Expected: Excel’s numeric result
  - Actual: Different numeric value in IronCalc
  - Issue: Formula implementation for LCM/GCD may diverge from Excel logic.

| Cell(s) | Arguments | Actual | Expected | Description |
| --- | --- | --- | --- |--- |
| D24 | `="8"`, `2` | #NUM! | 0 | Coerced text numbers to numeric |
| D25 | `2`, `="8"` | #NUM! | 0 | Coerced text numbers to numeric |
| D26, E26 | `"= 10"`, `3` | #NUM! | 0 | Coerced text numbers to numeric |
[...]

PR #757, opened 2026-02-10
fix: bug fixes in GCD and LCM

elsaminsut:
This PR fixes most issues in #551. Changes affect cells, ranges and arrays and include:
- Adds a new upper limit for any argument in GCD or the result of LCM (MAX_LCM_GCD = 2^53, according to the [Excel documentation](https://support.microsoft.com/en-us/office/lcm-function-7152b67a-8bb5-4075-ae5c-06ede5563c94))
- Adds support for numeric strings
- Non-numeric strings now return `#VALUE!` 
- Booleans are not handled and also return `#VALUE!` 
- Any empty arguments (like `=LCM( , 2)`) result in `#N/A` 
- Empty cells and arguments in ranges are treated as 0

It also includes a major refactor of the two functions in their own separate file, sharing most of their code through a helper function `fn gcd_lcm_impl`

All cases (and more) are covered in the xlsx test file.

There is still a (maybe minor) precision issue described in #756


nhatcher (approved):
Lgtm!

Issue #751, opened 2026-02-08
Bug: Invalid formula text return wrong error type for non-Latin characters


rezat-c:
When entering a formula with an invalid reference using non-Latin characters (e.g., =ä or =ы), returns #ERROR! instead of the expected #NAME? error.

Expected behavior: #NAME?
Actual behavior: #ERROR! for non-Latin character references.
Example:
- =ä returns #ERROR! (should be #NAME?)
- =ы returns #ERROR! (should be #NAME?)

PR #826, opened 2026-03-28
Fix/issue 751 lexer non ascii

CoalUnicorn:
Fix to #751 .


nhatcher:
Hi @CoalUnicorn , I resolved the merged conficts manually. I still haven't had the time to properly review this important PR, sorry about that 
:/

nhatcher (changes requested):
Hi, this is great, but why don't you just chnage `is_valid_identifier` to admit unicode characters.

We should add tests for defined names, sheets, LAMBDA functions, etc...

PR #850, opened 2026-04-06
update: add reusable `Select` component

### LINKED ISSUE #857: widget: implement a reusable `Select` component
We will implement a reusable, MUI-free select component. 

### Requirements
1. Selects should be consistent in styling with existing inputs
2. Should also be consistent with `Input` in terms of properties (size, label, etc)
3. It should include a `ghost` variant (no border)
4. Should adapt its positioning so it doesn't crop when it's too close to the edges or the display is too narrow
5. It should include a story for Storybook with some examples and cases

dg-ac:
This PR introduces a new `Select` component (`Select.tsx` + `select.css`) following the same patterns as the existing `Input` component. It also updates 2 other elements with this new component, and adds a Storybook story.

### Changes

**New files**
- `Select.tsx`: the reusable select component. We will be able to toggle `label`, `helperText`, `error`, `mandatory`... we also include a `ghost` variant (simple select, no borders)
  - `select.css`: styles following the same conventions as in `input.css`
- `Select.stories.tsx`: Storybook stories covering different cases

**Edited files**
- `EditNamedRange.tsx`: updated the select so it reflects the last changes
  - `edit-name-range.css`: remove unused classes
- `RegionalSettings.tsx`: updated the select so it reflects the last changes
  - `regional-settings.css`: remove unused classes


webapp/IronCalc/src/components/Select/select.css
```diff
@@ -1,16 +1,61 @@
...
+.ic-select > label {
+  display: block;
+  align-self: flex-start;
+  color: var(--palette-common-black);
+}
+
+.ic-select > label[data-required]::after {
+  content: " *";
+  color: var(--palette-error-main);
+}
```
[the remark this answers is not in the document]
dg-ac: Makes sense

webapp/IronCalc/src/components/Select/select.css
```diff
@@ -112,25 +189,28 @@
   min-width: 0;
   display: inline-flex;
   align-items: baseline;
+  text-wrap: nowrap;
```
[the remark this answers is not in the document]
dg-ac: Just noticed I placed this in the wrong place, so I've moved it to `.ic-select-trigger-value`:

```
.ic-select-trigger-value {
  flex: 1;
  min-width: 0;
  display: block;
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

webapp/IronCalc/src/components/Select/select.css
```diff
@@ -49,18 +83,60 @@
...
+}
+
+.ic-select-control.sm .ic-select-trigger-icon {
+  right: 6px;
+}
+
+.ic-select-control.md .ic-select-trigger {
+  height: 32px;
+  padding: 8px 32px 8px 10px;
+}
```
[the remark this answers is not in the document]
dg-ac: I'll change it

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -209,75 +228,85 @@ export function Select({
...
+          aria-haspopup="listbox"
+          aria-expanded={open ? "true" : "false"}
+          aria-controls={listboxId}
+          aria-invalid={error || undefined}
+          aria-describedby={helperText ? helperId : undefined}
+          aria-required={required || undefined}
+          disabled={disabled}
+          onClick={toggleMenu}
+          onKeyDown={handleTriggerKeyDown}
+        >
```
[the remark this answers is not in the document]
dg-ac: Removed

webapp/IronCalc/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,174 @@
...
+          required
+        />
+        <Select
+          label="Error"
+          options={fruitOptions}
+          value="cherry"
+          onChange={() => {}}
+          error
+          helperText="This value is invalid."
+        />
```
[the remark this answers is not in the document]
dg-ac: Oops, good catch. Fixed

webapp/IronCalc/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,174 @@
...
+  render: () => {
+    const [value, setValue] = useState("de");
+    return (
+      <div style={{ width: 120 }}>
+        <Select
+          label="Breakfast"
+          options={fruitOptions}
+          value={value}
+          onChange={setValue}
+        />
```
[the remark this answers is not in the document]
dg-ac: Fixed

webapp/IronCalc/src/components/Select/select.css
```diff
@@ -1,16 +1,63 @@
...
   font-size: var(--typography-font-size);
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
+}
+
+/* Label */
+.ic-select > label {
+  display: block;
+  align-self: flex-start;
```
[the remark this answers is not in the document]
dg-ac: Got it. Added

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -209,75 +228,82 @@ export function Select({
...
-            role="listbox"
-            aria-labelledby={ariaLabelledBy}
-          >
-            {options.map((option, index) => {
-              const isSelected = option.value === value;
-              const isActive = index === activeIndex;
+        {open ? (
+          <div className="ic-select-menu-wrapper">
+            <div id={listboxId} className="ic-select-menu" role="listbox">
+              {options.map((option, index) => {
```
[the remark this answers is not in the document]
dg-ac: I've connected `aria-labelledby` from the listbox (and trigger) to the visible `<label>` using a stable label ID.

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -209,75 +228,82 @@ export function Select({
...
-    <div ref={rootRef} className={rootClassName}>
-      {name ? <input type="hidden" name={name} value={value} /> : null}
+    <div className={["ic-select", className].filter(Boolean).join(" ")}>
+      {label && (
+        <label htmlFor={selectId} data-required={required || undefined}>
+          {label}
+        </label>
+      )}
+
+      <div ref={rootRef} className={controlClassName}>
```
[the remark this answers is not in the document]
dg-ac: Amazing, this is great. Changed.

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -209,75 +230,90 @@ export function Select({
...
-            {options.map((option, index) => {
-              const isSelected = option.value === value;
-              const isActive = index === activeIndex;
+        {open ? (
+          <div className="ic-select-menu-wrapper">
+            <div
+              id={listboxId}
+              className="ic-select-menu"
+              role="listbox"
+              aria-labelledby={label ? labelId : undefined}
```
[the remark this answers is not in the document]
dg-ac: Make sense, let's fall back to `valueId`.

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -209,75 +230,90 @@ export function Select({
...
-    <div ref={rootRef} className={rootClassName}>
-      {name ? <input type="hidden" name={name} value={value} /> : null}
+    <div
+      ref={rootRef}
+      className={["ic-select", className].filter(Boolean).join(" ")}
+    >
+      {label && (
+        <label htmlFor={selectId} data-required={required || undefined}>
+          {label}
+        </label>
```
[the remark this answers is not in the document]
dg-ac: Made some adjustments to fix this.

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -83,40 +170,75 @@ export function Select({
...
+      const trigger = triggerRef.current;
+      const menu = menuRef.current;
+      if (!trigger || !menu) {
+        return;
+      }
+
+      const { top, left, minWidth } = getMenuPosition(trigger, menu);
+
+      setMenuStyle({
+        position: "fixed",
```
nhatcher: Why do you do that? You can just set the position as fixed in the css.

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -175,109 +301,154 @@ export function Select({
...
         break;
       }
-      case "Tab": {
-        setOpen(false);
-        break;
-      }
     }
   }
 
+  const { onKeyDown: handleMenuKeyDown } = useKeyDown({
```
nhatcher: You don't need this at all. This is handling:

* Tab
* Esc

Both are handled somewhere else. Also in  this particular case (the dropdown) you possibly want tab to focus on a different element and not stay in the dropdown.

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -35,30 +103,38 @@ export function Select({
...
   const optionRefs = useRef<Array<HTMLButtonElement | null>>([]);
 
   const selectedIndex = Math.max(
     0,
     options.findIndex((option) => option.value === value),
   );
 
-  // FIXME: This fallback to 0 can cause issues if the value is not in options.
-  // See: https://github.com/ironcalc/IronCalc/pull/834#discussion_r3015897822
+  // Clamp missing values (-1) to 0 to keep a valid, focusable selection.
```
nhatcher: I don't think this is fixed. And the comment is a bit enigmatic.
What the comments is saying:

If there are no options:

```html
<select value="whatever">
</select>
```

then `option[0]` is the wrong thing to do.

To pile on. That line of code is selecting the first element if the value is not in the list:

```html
<select value="qwerty7">
  <option value="qwerty0">qwerty0</option>
  <option value="qwerty1">qwerty1</option>
  <option value="qwerty2">qwerty2</option>
  <option value="qwerty3">qwerty3</option>
</select>
```

That hopefully will never happen. But If you are using the element and make a type you won't notice because the select is going to work.
Could we have an "invalid state"? Does it make sense?

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -35,30 +103,38 @@ export function Select({
...
   const selectedOption = options[selectedIndex] ?? options[0];
 
   const [open, setOpen] = useState(false);
   const [activeIndex, setActiveIndex] = useState(selectedIndex);
+  const [menuStyle, setMenuStyle] = useState<CSSProperties>({});
```
nhatcher: Those are not the full set of CSSProperties though. The only thing you are changing here are the position (top, left and min-width). This feels too much

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -175,109 +301,154 @@ export function Select({
...
+          aria-haspopup="listbox"
+          aria-expanded={open ? "true" : "false"}
+          aria-controls={listboxId}
+          aria-labelledby={label ? `${labelId} ${valueId}` : undefined}
+          aria-required={required || undefined}
+          aria-invalid={error || undefined}
+          aria-describedby={helperText ? helperId : undefined}
+          disabled={disabled}
+          onClick={toggleMenu}
+          onKeyDown={handleTriggerKeyDown}
```
nhatcher, line 327: The trigger should only handle opening and closing the dropdown. A strightfoward:

```ts
  function handleTriggerKeyDown(event: KeyboardEvent<HTMLButtonElement>) {
    if (disabled || options.length === 0 || open) {
      return;
    }
    switch (event.key) {
      case "ArrowDown":
      case "ArrowUp":
      case "Enter":
      case " ": {
        event.preventDefault();
        openMenu();
        break;
      }
    }
  }
```

would work. Also Do we nned ArrowUp "Enter" and " " to open the menu? Or only "ArrowDown"?
See:

https://semantic-ui.com/modules/dropdown.html

It is what MUI does:

https://mui.com/material-ui/react-select/

In pure HTML:

```html
<select value="qwerty2">
  <option value="qwerty0">qwerty0</option>
  <option value="qwerty1">qwerty1</option>
  <option value="qwerty2">qwerty2</option>
  <option value="qwerty3">qwerty3</option>
</select>
```

Arrow keys will chnage selection even with the menu closed (Google Chrome Linux). I personally find that behaviour annoying and surprising

That being said I am not a huge fan of MUI. What is the standard here? Is there something about the "correct" or "expected" behaviour here

webapp/IronCalc/src/components/Select/useKeyDown.ts
```diff
@@ -0,0 +1,62 @@
+import { type KeyboardEvent, useCallback } from "react";
```
nhatcher: I think this file can be comopletely deleted

webapp/IronCalc/src/components/RightDrawer/RegionalSettings/RegionalSettings.tsx
```diff
@@ -180,26 +171,17 @@ const RegionalSettings = (properties: RegionalSettingsProps) => {
...
-                  triggerLabel: timezone,
-                }))}
-                compact
-              />
-              {/* FIXME: https://github.com/ironcalc/IronCalc/pull/824#discussion_r3003370209 */}
-              <p className="ic-regional-settings-helper-text">
-                {t("regional_settings.timezone.timezone_helper")}
-              </p>
-            </div>
+            <Select
```
nhatcher, line 174: You have removed the "compact" version here. So vissually changes from:

Image
To

Image

Is that intentional?

we will add a searchable dropdown one day, of course.

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -1,32 +1,100 @@
...
+
+/**
+ * Reusable Select with label, helper text, error state, and size variants.
+ * Sizes: sm, md
+ * States: default, hover, focused, error, disabled.
+ */
+
+export type SelectVariant = "default" | "ghost";
+
+export type SelectSize = "sm" | "md";
```
nhatcher, line 26: Do we use the "sm" variety anywhere in the code base?

webapp/IronCalc/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const [tr, setTr] = useState("banana");
+    const [bl, setBl] = useState("durian");
+    const [br, setBr] = useState("elderberry");
+
+    return (
+      <div
+        style={{
+          position: "relative",
+          height: "100vh",
+          boxSizing: "border-box",
```
nhatcher: I think I would do something like this:
```css
          position: "absolute",
          top: "10%",
          left: "10%",
          width: "80%",
          height: "80%",
          boxSizing: "border-box",
          border: "1px dashed gray",
```
So it looks something like this:

Image

So basically a bit smaller than the storybook window so we don't relay on it

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -1,32 +1,100 @@
...
+  error?: boolean;
+  required?: boolean;
   disabled?: boolean;
-  ariaLabelledBy?: string;
-  ariaDescribedBy?: string;
   className?: string;
-  compact?: boolean;
+}
+
+function getMenuPosition(trigger: HTMLElement, menu: HTMLElement) {
```
nhatcher, line 50: This is great but I feel it needs a bit of documentation. This function _could_ be made generic (I'm not saying we should) but this could be extracted to a general utility for many of our menus, so we don't repeat code over and over

webapp/IronCalc/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,236 @@
...
+            label="Bottom-right"
+            options={fruitOptions}
+            value={br}
+            onChange={setBr}
+          />
+        </div>
+      </div>
+    );
+  },
+};
```
nhatcher, line 302: Maybe you could add a couple of stories:

```tsx
export const NoOptions: Story = {
  args: defaultArgs,
  render: () => {
    const [value, setValue] = useState("durian");
    const options: SelectOption[] = [];
    return (
      <div style={{ width: 120 }}>
        <Select
          label="No options"
          options={options}
          value={value}
          onChange={setValue}
        />
      </div>
    );
  },
};

export const Invalid: Story = {
  args: defaultArgs,
  render: () => {
    const [value, setValue] = useState("orange");
    return (
      <div style={{ width: 120 }}>
        <Select
          label="Breakfast"
          options={fruitOptions}
          value={value}
          onChange={setValue}
        />
      </div>
    );
  },
};
```

webapp/IronCalc/src/components/Select/Select.tsx
```diff
@@ -1,32 +1,100 @@
...
   useEffect,
   useId,
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
 } from "react";
+import { createPortal } from "react-dom";
+
 import "./select.css";
```
nhatcher, line 15: I was thinking since we are now using a portal would it make sense to split the select.css into two files:

* select.css
* dropdown-menu.css

Remember that the portal is not a child of select anymore

nhatcher (changes requested):
Looks great!

I added some coments

webapp/IronCalc/src/components/RightDrawer/RegionalSettings/RegionalSettings.tsx
dg-ac: Yep, it felt odd to have different styling here.

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: We will in an update I'm gonna do in the next couple of weeks (in the Name Manager).

webapp/IronCalc/src/components/Select/Select.stories.tsx
dg-ac: If you think it's easier to understand like this I can add it.

webapp/IronCalc/src/components/RightDrawer/RegionalSettings/RegionalSettings.tsx
nhatcher: sounds good.

webapp/IronCalc/src/components/Select/Select.stories.tsx
nhatcher: I am not sure it is easier ti understand. I think it is, but it also gives you the oportunity to do other things, like resizing that part of the windowand see what happens. In larger screens there is just too much space around

webapp/IronCalc/src/components/Select/Select.tsx
nhatcher: Can we add the sm class then instead of now?

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: Makes sense. I've divided the css in 2 files.

webapp/IronCalc/src/components/Select/Select.stories.tsx
dg-ac: Added!

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: This makes sense but I haven't made the change yet. Shall we leave it for a different PR?

webapp/IronCalc/src/components/Select/Select.stories.tsx
dg-ac: Gotcha, I've added it.

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: As discussed, I'll use in a small update I want to tackle right after this PR is approved.

webapp/IronCalc/src/components/Select/useKeyDown.ts
dg-ac: Removed!

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: Hmm you're right. I've refined this part.

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: Can you check again?

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: I see, I see. Please see the latest version, we can have a call if it's still not as desired.

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: Removed, you're right.

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: True. Moved to the `.css` file.

webapp/IronCalc/src/components/Select/Select.tsx
nhatcher: You can add the documentation in a comment  (or series of comments) now. There is no reason to do that later.
The comments should explain the logic and when to use is so if I see the comment I know if I can use it for a different component or not

dg-ac: Makes sense. Added

webapp/IronCalc/src/components/Select/Select.tsx
nhatcher: Let's discuss this on a call.
The point is you ar not changing the CSSProperties. this should be:
```ts
const [menuPosition, setMenuPostion] = ...
```

webapp/IronCalc/src/components/Select/Select.tsx
nhatcher: Can you explain what you did?

webapp/IronCalc/src/components/Select/Select.tsx
nhatcher: agreed!

webapp/IronCalc/src/components/Select/Select.tsx
nhatcher: Let's discuss this on a call.

webapp/IronCalc/src/components/Select/dropdown-menu.css
```diff
@@ -0,0 +1,63 @@
+/* Dropdown menu */
+.ic-select-menu-wrapper {
```
nhatcher: I _think_ all the classes in this file should be namespaced with`ic-dropdown-` and not `ic-select`.
A different posibility would be to have all the classes in here `ic-select-menu-` and on the other file `ic-select-trigger-`

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: Oh of course, I missed updating this. Done.

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: I've changed some class names int he last push. Those inside `dropdown-menu.css` don't mention `select` anymore (`ic-select-option` → `ic-dropdown-menu-option`).

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: After the discussion, I've added a few more context aware comments, et me know if they make sense.

webapp/IronCalc/src/components/Select/Select.stories.tsx
dg-ac: I've finally removed the container we were using and placed the select examples in the edges. Note that the menu positioning depends on the window's width, but we may consider changing this in the future.

webapp/IronCalc/src/components/Select/Select.tsx
dg-ac: With the change, hitting ArrowUp and ArrowDown does the same: opens the menu and focuses on the selected option.

nhatcher (approved):
LGTM!

PR #861, opened 2026-04-11
UPDATE: Add Ukrainian (uk-UA) language support

Dreykp:
Summary                                                                                                                                                                                                         
  
  - Add Ukrainian translation files for both @ironcalc/workbook and the frontend app                                                                                                                              
  - Register uk-UA locale in i18n configs, language switcher menu, and storage utilities
  - Add browser language auto-detection and locale code mapping for Ukrainian

### DIFF
--- a/webapp/IronCalc/src/i18n.ts
+++ b/webapp/IronCalc/src/i18n.ts
@@ -4,6 +4,7 @@ import translationEN from "./locale/en_us.json";
 import translationES from "./locale/es_es.json";
 import translationFR from "./locale/fr_fr.json";
 import translationIT from "./locale/it_it.json";
+import translationUK from "./locale/uk_ua.json";
 
 const resources = {
   "en-US": { translation: translationEN },
@@ -12,6 +13,7 @@ const resources = {
   "fr-FR": { translation: translationFR },
   "de-DE": { translation: translationDE },
   "it-IT": { translation: translationIT },
+  "uk-UA": { translation: translationUK },
 };
 
 const instance: typeof i18n = i18n.createInstance({
--- a/webapp/IronCalc/src/locale/de_de.json
+++ b/webapp/IronCalc/src/locale/de_de.json
@@ -197,14 +197,16 @@
         "es": "Español",
         "fr": "Français",
         "de": "Deutsch",
-        "it": "Italiano"
+        "it": "Italiano",
+        "uk": "Українська"
       },
       "display_language_current_lang": {
         "en": "Englisch",
         "es": "Spanisch",
         "fr": "Französisch",
         "de": "Deutsch",
-        "it": "Italienisch"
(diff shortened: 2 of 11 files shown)


dg-ac:
Hey @Dreykp, this is fantastic. I will review and test it in the next couple of days.

On the other hand, I'm wondering, do you also translate the functions when using spreadsheets in Ukrainian? Or do you just use the English ones?

Dreykp:
Hello @dg-ac, for now the formulas remain in English, but I’m considering translating them into Ukrainian in the future. I also plan to add support for Ukrainian locales formats.

nhatcher:
This is great, thanks!

There are a few things mssing:

* We need to add the function names in Ukrainian as well. Do they exist?
* We need a Ukranian "backend" locale. See:
       - https://github.com/ironcalc/IronCalc/blob/main/generate_locale/locales_list.json
       - https://github.com/ironcalc/IronCalc/tree/main/base/src/locale
       - https://github.com/ironcalc/IronCalc/tree/main/base/src/language

* I think we will need new fonts to support cyrillic as the font-family being loaded is not Inter anymore.

One issue we need to be mindfull. How much code (kb) does this add to the system. After this language I think we need to have a way to load languages and locales dynamically

nhatcher (changes requested):
Please see my comment above

Dreykp:
Hi @nhatcher,

I’ve been thinking about translating formulas for quite a while, but came to the conclusion that it doesn’t make much sense at this stage. If you look at tools like Microsoft Excel or Google Sheets, they don’t translate formulas either, and users are already привыкли to this format.

nhatcher:
H @dg-ac , what is the status of this? Are we still waiting on something or we can't merge this?

As an aside: this would be really cool!

dg-ac:
@nhatcher I'd say this is not ready, for many reasons:

- First, we've added so much new stuff in the past weeks that there are many conflicts in the translation files, and some things are not translated. This should be the easiest part to fix
- We also don't fully tackle here other details regarding locales:
  - In formulas, do we separate arguments with comma, or semicolon? 
  - How are dates written in Ukrainian? mm/dd/yyyy? dd/mm/yyyy? @elsaminsut sent me [this](https://github.com/unicode-cldr/cldr-dates-modern/tree/master/main/ru-UA) as a reference and there may be many caveats
  - Is the currency symbol placed before or after quantity?
  - etc
  
As @Dreykp said before, my understanding is that function names are not translated when using other spreadsheet solutions in Ukrainian so we would be covered there.

I'd love to see this in IronCalc and think we will if we solve the remaining issues.

@Dreykp Feel free to jump in if you want to discuss further!

nhatcher:
Hi, let's park this for now.  @Dreykp if you want we can reopen. We will do this anyway at some point

PR #864, opened 2026-04-13
Storybook example in Shadow DOM

jvdp:
The first commit will make the shadow DOM incompatibility very obvious (and ugly) when you look at the storybook. ~The second is one way to deal with the issue; a `documentOrShadowRoot` property on `Workbook` (defaults to an auto-detected root) that is propagated via context.~

Let me know what you think.

Resolves #831.

### DIFF
--- a/webapp/app.ironcalc.com/frontend/.storybook/main.ts
+++ b/webapp/app.ironcalc.com/frontend/.storybook/main.ts
@@ -1,0 +1,21 @@
+import type { StorybookConfig } from "@storybook/react-vite";
+
+const config: StorybookConfig = {
+  stories: ["../src/**/*.mdx", "../src/**/*.stories.@(js|jsx|mjs|ts|tsx)"],
+  addons: [],
+  framework: {
+    name: "@storybook/react-vite",
+    options: {},
+  },
+  core: {
+    disableTelemetry: true,
+  },
+  viteFinal: (config) => {
+    config.server ??= {};
+    config.server.fs ??= {};
+    config.server.fs.allow = ["../../../"];
+    return config;
+  },
+};
+
+export default config;
--- a/webapp/app.ironcalc.com/frontend/.storybook/preview.ts
+++ b/webapp/app.ironcalc.com/frontend/.storybook/preview.ts
@@ -1,0 +1,10 @@
+import type { Preview } from "@storybook/react";
+import "../src/fonts.css";
+
+const preview: Preview = {
+  parameters: {
+    layout: "fullscreen",
+  },
+};
+
+export default preview;
--- a/webapp/app.ironcalc.com/frontend/package-lock.json
+++ b/webapp/app.ironcalc.com/frontend/package-lock.json
@@ -21,9 +21,12 @@
       },
       "devDependencies": {
         "@biomejs/biome": "2.4.7",
+        "@storybook/react": "^10.3.5",
+        "@storybook/react-vite": "^10.3.5",
         "@types/react": "^19.2.4",
         "@types/react-dom": "^19.2.3",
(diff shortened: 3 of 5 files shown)


webapp/IronCalc/src/components/AdvancedColorPicker.tsx/AdvancedColorPicker.tsx
```diff
@@ -5,7 +5,7 @@ import { HexColorInput, HexColorPicker } from "react-colorful";
 import { createPortal } from "react-dom";
 import { useTranslation } from "react-i18next";
 import { Button } from "../Button/Button";
-import "./advanced-color-picker.css";
```
nhatcher, line 8: Why do you remove all css files?

webapp/IronCalc/src/components/Workbook/Workbook.tsx
```diff
@@ -9,6 +9,7 @@ import {
   CLIPBOARD_ID_SESSION_STORAGE_KEY,
```
nhatcher, line 9: Hi @jvdp , this file has way too many changes for me to review. Maybe can you explain to me what are you trying to accomplish?

What is the problem you are solving?

You can give me a minimal html that includes IronCalc and it is broken somehow.

jvdp: If you checkout the code at https://github.com/ironcalc/IronCalc/pull/864/changes/a1c5fb975a2d49eb3d2108aa0151396d78b5436f and open the storybook you will see a completely broken workbook due to CSS rules being applied from 'outside'. That's basically the problem this is solving.

webapp/IronCalc/src/components/AdvancedColorPicker.tsx/AdvancedColorPicker.tsx
jvdp: Importing them in a javascript file is actually a side-effect. (You had to update the [`sideEffects` array in `package.json` for this](https://github.com/ironcalc/IronCalc/commit/86c2f4e2535e01e921c03d462eef34d3cce09932#diff-69bd1dcaa8c905e36856c212153e80cc6f037f2455b38a1f7cd41b4b8eda5943).) Vite will generate some code that adds the stylesheet to `document.head`, and that's what we (well, I) don't want.

jvdp:
I think the diff is a bit too big due to formatting issues, will see if I can fix this.

Also, the changes to the components are really just a suggestion. The main contribution in this PR is the storybook that illustrates the issue.

webapp/IronCalc/src/components/AdvancedColorPicker.tsx/AdvancedColorPicker.tsx
nhatcher: I am not sure what you mean here. But you can see how it works in the app.ironcalc.com.
The way we do it now creates a single css file that you will have to import in your project separated from the js. I actually like that. I also like that the css is imported in the files where they are used.

Maybe we we need to have a call about these issues?

jvdp: I have learned more about this single CSS file and it provides the flexibility that I need; I will update this PR.

jvdp:
I've added a storybook to the frontend project because that's where the CSS works representatively (ie. you have to explicitly import the CSS.) It embeds an IronCalc Workbook on a page with clashing CSS styles and so it can provide a way to test fixes for this. It currently looks so messed up I think it can't be right, but have a look if you're curious.

nhatcher:
Thank you for this! As explained on a call we can't do this here because the app might have it's own storybook. But we can definitely revisit this later

PR #518, opened 2025-11-07
merge duration, mduration #44

nhatcher:
merge price, yield #29

merge fvschedule #56

merge pricedisc, pricemat, yielddisc, yieldmat, disc, received, intrate #57

merge accrint, accrintm #58

merge coupdaybs, coupdays, coupdaysnc, coupncd, coupnum, couppcd #59

fix cursor

refactor

refactor

fix build

PR #865, opened 2026-04-13
UPDATE: Add ACCRINTM function (accrued interest at maturity)

albertodiazdurana:
## Summary

- Add `ACCRINTM(issue, settlement, rate, [par], [basis])` function
- Delegates to existing `fn_yearfrac` for day-count computation
- 13 unit tests covering all code paths, error conditions, and day-count edge cases

## Design decisions

- **Par is optional** (default 1000) per Microsoft/LibreOffice spec. Nico's scaffold had par as required; adjusted to `args_signature_scalars(arg_count, 3, 2)`.
- **Delegates to `fn_yearfrac`** rather than reimplementing day-count logic. This avoids the basis 1 (Actual/Actual) Procedure E bug found in PR #427's standalone `year_fraction` helper.
- **`par <= 0` returns `#NUM!`** per Microsoft spec (not zero as in #427).
- Locale translations added: en=ACCRINTM, de=AUFGELZINSF, fr=INTERET.ACC.MAT, es=INT.ACUM.V.

## Test coverage

13 unit tests:
- Microsoft reference example (basis 3)
- Default arguments (3-arg form)
- All 5 basis options
- Error conditions: date ordering, rate/par validation, basis range, arg count
- Basis 0 vs 4 month-end divergence
- Feb 28 leap-aware 30/360 adjustment (basis 0)
- Cross-year `is_feb_29_between_dates` edge case (basis 1)
- Leap vs non-leap Feb 28→Mar 1 contrast (basis 1)

20 additional test cases (covering all Procedure E branches, multi-year spans, basis float truncation) were validated during development and trimmed to avoid over-testing the existing `fn_yearfrac` implementation. Available on request.

XLSX integration tests deferred: no Excel-computed ACCRINTM reference file exists yet.

## Context

Aware of #427 (BrianHung, financial functions) and #518 (rebase). This PR covers only ACCRINTM with a minimal, self-contained approach. ACCRINT will follow as a separate PR.

Related: Discussion #762 (AI collaboration guidelines).


nhatcher (changes requested):
Hi @albertodiazdurana,

Thanks for a first issue! Several thoughts.
* We definitelly need an xlsx file covering all test cases
* ACCRINTM is an extremelly simple function. It is not a good test case for AI, ACCRINT is the good test case

That being said we could still incorporate this. It would be very nice to know what was the prodedure exactly.

* What tool did you use?
* What was the prompt/SKILL.md/CLAUDE.md

You could dump all this in the issue. I think the AI engine did identify a number of generic issues like languages/arguments/etc. So it is  agood first step.

albertodiazdurana:
Hi @nhatcher, thanks for the read and for being open to incorporating it.

**On the tooling.** The work was done with Claude Opus 4.6, driven by a methodology I developed called *Take an AI Bite*, built on top of DSM (Deliberate Structured Methodology). The premise is simple: the human stays in control, and every piece of work is cut into a bite small enough to think about and sharp enough to act on. In practice that means three human-approved gates on every artifact — concept, implementation, run — and a planning + research loop around each one. Concretely for this PR: a per-BL plan, a pack of research docs (YEARFRAC day-count audit, ACCRINTM spec, prior-PR analysis, edge cases), then the commit sequence you see (scaffold, body, rigorous tests), each a separate gate. Full write-up at https://take-ai-bite.com/about/.

**On ACCRINTM first.** Fair call that it's simple. That was intentional — it was the right-sized first bite: low-blast-radius, enough surface to learn the repo conventions (function-lookup macro, localization wiring, test patterns, clippy surface), then carry the same pipeline to ACCRINT where the multi-period logic and day-count basis earn the methodology. ACCRINT is the next PR.

**On the xlsx fixture.** On it. Will add one covering all thirteen test cases and push to this branch.

albertodiazdurana:
**Update on the xlsx fixture.** The fixture lands as `xlsx/tests/calc_tests/ACCRINTM.xlsx` with 11 rows covering the core happy paths (canonical MS example, 3-arg defaults, all five basis codes, US vs EU 30/360 month-end, Feb 28 leap vs non-leap, short spans crossing Feb 29).

It's 11 rows rather than the 13 test cases in `base/src/test/test_fn_accrintm.rs` — 8 formulas were deliberately excluded after authoring the file in LibreOffice Calc and running `test_xlsx`. The exclusions fall into two categories:

- **Error-code divergence (5 rows)**: LibreOffice returns `#VALUE!` for illegal-arg cases (issue ≥ settlement, rate ≤ 0, par ≤ 0, basis outside 0–4) where the Microsoft ACCRINTM spec and Excel return `#NUM!`. IronCalc follows the spec; the round-trip framework compares error codes byte-exact, so any tool divergence fails.
- **Numeric divergence (3 rows)**: LibreOffice's basis 4 computes basis-0 semantics for the Mar 31 month-end case; LibreOffice adjusts D1 on Feb 28 even in leap years (MS spec says not to); LibreOffice's Feb-29-in-range detector doesn't cross year boundaries. IronCalc matches Excel / MS spec on all three.

These 8 cases stay fully covered by the Rust unit tests in `test_fn_accrintm.rs`. The xlsx fixture's purpose is I/O round-trip validation, not duplicating semantic coverage — pinning it to the 11 portable rows keeps it tool-agnostic (Excel and LibreOffice produce the same cached values). Happy to add the divergent cases back if/when I can get Excel desktop or Office 365 web access; they should pass there.

.gitattributes
```diff
@@ -0,0 +1,3 @@
+# Enforce LF line endings for all text files.
```
nhatcher: Please remove this file

albertodiazdurana:
**Fixture extended to 18 rows.** `xlsx/tests/calc_tests/ACCRINTM.xlsx` now covers 18 of the 21 `ACCRINTM(...)` calls in `test_fn_accrintm.rs` (commit 2270f617). Authored in MS Excel (Office Web) so cached values come out spec-correct without the LibreOffice divergences we hit on the original 11-row pass. Round-trip (`test_calc_ACCRINTM`) passes.

Three unit-test cases remain unfixturable; they stay covered by the Rust unit tests:

1. **`error_arg_count` (too few args):** Excel's formula parser refuses to save `=ACCRINTM(a,b)`, you can't embed a malformed-arity formula as a cached `#NUM!`. Parser-side limit, not a deliberate exclusion.
2. **`error_arg_count` (too many args):** Same parser limit on `=ACCRINTM(a,b,c,d,e,f)`.
3. **`error_issue_equals_settlement`:** Excel Online returns `0` for same-day issue/settlement, contradicting Microsoft's own `ACCRINTM` documentation which states `issue >= settlement` produces `#NUM!`. IronCalc follows the documented spec, so this row would break the byte-exact round-trip. The unit test still asserts the spec-correct `#NUM!`.

nhatcher (approved):
LGTM!

Many thanks @albertodiazdurana !

.gitattributes
nhatcher: Sorry, this was a dangling comment, from a previous review. I'll remove it myself.

nhatcher:
Hi @albertodiazdurana, we have some nasty conflics. In this particular case the path of least resistance would be to forget the rebase and do everything from scratch.
The most problematic file would be the languages.bin file. But it would literally take you a few minutes to copy over the code in a new branch, regenerate the binaries and force push the branch. Let me know if you want to do that on a call

albertodiazdurana:
Done as suggested. Fresh branch off current `main`, single squashed commit (`af31c12`). Regenerated `language.bin`. Pre-flight green: `cargo fmt`, `cargo clippy -D warnings`, 1445 base tests, full `make lint` and `make test-rust`.

No content changes vs the prior head; 7 of 8 source files are byte-identical, the eighth (`test/mod.rs`) just picks up your `mod conditional_formatting;` from current main. Fixture and unit tests unchanged.

Ready for CI whenever you have a moment.

PR #866, opened 2026-04-14
update: add reusable `Tooltip` component

### LINKED ISSUE #858: widget: implement a reusable `Tooltip` component
We will implement a reusable, MUI-free tooltip component.

### Requirements
1. Selects should be consistent in styling with existing tooltips
2. Should handle layout properly so it doesn't crop when it's too close tho the window's edge
3. It should include a story for Storybook with some examples and cases

dg-ac:
Replaces the previous `Tooltip` with a custom one. The tooltip renders via a portal, positions itself relative to its trigger and has with viewport-aware clamping so it doesn't crop when it's too close to the window's edge.

### Changes

1. `Tooltip.tsx`
- Renders a wrapper span around the trigger that listens to mouseenter/mouseleave and focus/blur to control visibility.
- Renders the tooltip itself into document.body via createPortal, keeping it out of any overflow/stacking context

2. `useTooltipPosition.ts`
Consistent with what we recently added in the `Select` component.

3. `tooltip.css`
Very simple styling with a subtle animation. The tooltip takes 150ms to set the opacity to 1 (on mouse enter) and the same to set it to 0 (on mouse leave).

4. `Tooltip.stories.tsx`: 2 stories added:
- Default: simple button with a tooltip
- Corner positioning: 9 cases to see how the tooltip works on the corners and the sides

### DIFF
--- a/webapp/IronCalc/src/components/Tooltip/tooltip.css
+++ b/webapp/IronCalc/src/components/Tooltip/tooltip.css
@@ -4,7 +4,6 @@
 
 .ic-tooltip {
   position: fixed;
-  z-index: 9999;
   pointer-events: none;
   white-space: nowrap;
   font-family: var(--typography-font-family);
--- a/webapp/IronCalc/src/components/Tooltip/tooltip.css
+++ b/webapp/IronCalc/src/components/Tooltip/tooltip.css
@@ -1,0 +1,33 @@
+.ic-tooltip-trigger {
+  display: inline-flex;
+}
+
+.ic-tooltip {
+  position: fixed;
+  z-index: 9999;
+  pointer-events: none;
+  white-space: nowrap;
+  font-family: var(--typography-font-family);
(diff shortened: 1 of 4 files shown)


webapp/IronCalc/src/components/Tooltip/tooltip.css
```diff
@@ -0,0 +1,33 @@
+.ic-tooltip-trigger {
+  display: inline-flex;
+}
+
+.ic-tooltip {
+  position: fixed;
+  z-index: 9999;
```
nhatcher: We don't need z-index anymore

nhatcher (approved):
Aside from removing the z-index LGTM!

webapp/IronCalc/src/components/Tooltip/tooltip.css
dg-ac: That's true. Removed.

PR #867, opened 2026-04-14
fix: stop event propagation at Input component level

dg-ac:
We've moved `stopPropagation` for `keydown` and `click` events into the `Input` component directly, so callers don't need to handle it individually.


nhatcher (approved):
Thanks!

PR #870, opened 2026-04-16
update: add reusable `Modal` component 

### LINKED ISSUE #859: widget: implement a reusable `Dialog` component
We will implement a reusable, MUI-free `Dialog` component.

### Requirements
1. Clicking on the backdrop should close it
2. Design should be similar to the existing one
3. Header is optional
4. Hitting `esc` should close it as well
5. It should include its respective story on Storybook.

dg-ac:
This PR introduces a reusable Modal component with portal rendering, focus management, and keyboard handling.
It also migrates `SheetDeleteModal` and `FormatPicker` to use it, as they were using custom, non-consistent implementations.

Note that until now we were using 2 different stylings, one for normal modals and another for confirmation modals. **With this change we should only use one from now one**, which I think is better and will make maintenance easier.

**Before**
image image

**After**
image image

### Changes
Find below the files we're changing and an explanation of the changes.

In `src/components/Modal/`:

1. `Modal.tsx`
- New reusable modal. Includes:
  - Clickable backdrop
  - Optional header
  - Body
  - Footer
- Uses `createPortal` to render into `document.body`
- Supports `onConfirm` for `Cmd+Enter` (MacOS) and `Ctrl+Enter` (Windows/Linux)

2. `modal.css`
- Styles for backdrop, header, body, footer
- Footer stacks vertically and reverses order on mobile (max-width: 600px)
- We're getting rid of footer borders, header borders and unnecessary icons (in destructive modals)

3. `useModalFocus.ts`
- Saves previously focused element and restores it on close
- Focuses the modal container on open via `requestAnimationFrame`

4. `useModalKeyDown.ts`
Handles Escape (close), Cmd/Ctrl+Enter (confirm), Tab (focus trap), and plain Enter (blocked unless a button is focused)

5. `Modal.stories.tsx`
- Stories for Default, Destructive (only difference is the variant of the buttons) and NoHeader

In `src/components/SheetTabBar/`:

6. `SheetDeleteModal.tsx`
- Migrated to the new `Modal` and other parts that are handled by the component
- The CSS file has been removed as it's not needed anymore

7. `sheet-delete-modal.css`
- Deleted (no longer needed)

In `src/components/FormatMenu/`:

8. `FormatPicker.tsx`
- Migrated to the new `Modal.
- Removed `onExited` prop (was unused)

9. `FormatMenu.tsx`
- Removed `onExited` prop (was unused)

10. `Toolbar/Toolbar.tsx`
- Removed `onExited={(): void => {}} pass-through` (also unused afaik)


webapp/IronCalc/src/components/Dialog/dialog.css
```diff
@@ -0,0 +1,72 @@
+.ic-dialog-backdrop {
+  position: fixed;
+  inset: 0;
+  z-index: 1400;
```
nhatcher: Are we still using this, why?

webapp/IronCalc/src/components/Dialog/dialog.css
```diff
@@ -0,0 +1,72 @@
...
+}
+
+.ic-dialog {
+  width: 280px;
+  max-width: calc(100% - 40px);
+  display: flex;
+  flex-direction: column;
+  border-radius: 8px;
+  background: var(--palette-common-white);
+  box-shadow: 0px 1px 3px 0px rgb(39 37 37 / 10%);
```
nhatcher: Maybe we can create a variable for this?

webapp/IronCalc/src/components/Modal/modal-dialog.css
```diff
@@ -0,0 +1,72 @@
...
+
+.ic-dialog-footer {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  gap: 8px;
+  padding: 8px 12px;
+}
+
+@media (max-width: 600px) {
```
nhatcher, line 66: What is thgis doing?

webapp/IronCalc/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+    onClose: closeDialog,
+    onConfirm,
+  });
+
+  if (!open) {
+    return null;
+  }
+
+  return createPortal(
+    // biome-ignore lint/a11y/noStaticElementInteractions: FIXME
```
nhatcher: Can you explain what is going on here and why we can't dix it?

nhatcher (changes requested):
Hi Dani. I think having a generic Dialog is too general.

I think we should have three Modal dialogs (there are equivalent in js)

* Alert
* Confirm
* Prompt

I will need Alert right away so you might as well created.
Modern UX design uses "diapearing boxes" instead of alerts. Old folks like me hate them because I don't have time to read or understnad what happend. And at any rate we could start with an Alert

We also need alert when something has gone wrong with the model as we are currently using the javascrip.

Let me know what you think

webapp/IronCalc/src/components/Dialog/dialog.css
dg-ac: You mean the `z-index`, right? I've just removed it.

webapp/IronCalc/src/components/Modal/modal-dialog.css
dg-ac: This changes the direction of the flexbox to `column` in narrow displays. For mobile devices it's more standard[1] to use full-width buttons. Also, the reason to use `column-reverse` and not just `column` is that we want to keep the main action on top (so at the beginning of the flexbox).

---
[1] Open [shadcn's dialog](https://ui.shadcn.com/docs/components/radix/dialog) and narrow the screen

webapp/IronCalc/src/components/Dialog/dialog.css
dg-ac: Good idea. I've added a ` --shadow-sm` variable to `theme.css`.

webapp/IronCalc/src/components/Dialog/Dialog.tsx
dg-ac: I was having some issues with the linter as biome was complaining of the use of `role="presentation"`, saying that "static elements should not be interactive". My understanding was that tackling this needed a bigger refactor that could be out of scope so I added the `biome-ignore`.

Diving a bit deeper, I've learned that changing the role to `"none"` seems to fix the problem with no downside (afaik) so I've gotten rid of the ignore. This is just for the backdrop, and keyboard users can close the dialog hitting the `Esc` key, so my thought is that it can be ignored.

dg-ac:
Hey @nhatcher, I've made a few changes:

- The previous `Dialog.tsx` is now divided into `Alert`, `Confirm` and `Prompt`
- They are all similar so they use the same css file (`modal-dialog.css`)
- We're using now the terminology `modal` instead of `dialog`
- The storybook file has now 3 stories, one for each type of modal
- Updated the files that were using modals (SheetDeleteDialog and FormatPicker)

webapp/IronCalc/src/components/Modal/Confirm.tsx
```diff
@@ -0,0 +1,92 @@
+import { X } from "lucide-react";
+import { type ReactNode, useId } from "react";
+import { createPortal } from "react-dom";
+import { Button } from "../Button/Button";
+import { IconButton } from "../Button/IconButton";
+import "./modal-dialog.css";
```
nhatcher: Hmm, I don't think we should import this file from every file. Maybe we can have an index.ts and import it there
(I _think_ the build system should be smart enough not to include it three times, but I am not sure

webapp/IronCalc/src/components/Modal/modal-dialog.css
nhatcher: Is this the idea?

Image

The left being the "narrow" screen.

webapp/IronCalc/src/components/FormatMenu/FormatMenu.tsx
```diff
@@ -379,7 +378,6 @@ const FormatMenu = (properties: FormatMenuProperties) => {
         onChange={onSelect}
         open={isPickerOpen}
         onClose={() => setPickerOpen(false)}
-        onExited={properties.onExited}
```
nhatcher, line 382: Why are you removing this?

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
```diff
@@ -1,170 +1,39 @@
...
 import { useTranslation } from "react-i18next";
-import { Button } from "../Button/Button";
-import { IconButton } from "../Button/IconButton";
-import { Input } from "../Input/Input";
-import "./format-picker.css";
+import { Prompt } from "../Modal/Prompt";
 
-// FIXME: This control should be a modal prompt
 // FIXME: the stopPropagation everywhere is because of my bad implementation
 // of keyboard handling in the spreadsheet
```
nhatcher: I _think_ you should add this comment (and the stopPropagation calls) to the base implementation

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
```diff
@@ -1,170 +1,39 @@
...
+      defaultValue={properties.numFmt}
+      confirmLabel={t("num_fmt.save")}
+      inputProps={{
+        name: "format_code",
+        autoFocus: true,
+        spellCheck: false,
+        onFocus: (event) => event.target.select(),
+        onPaste: (event) => event.stopPropagation(),
+        onCopy: (event) => event.stopPropagation(),
+        onCut: (event) => event.stopPropagation(),
```
nhatcher: You could add all of those to the base implementation. Once we fix the underlying issue we could just remove them

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
```diff
@@ -1,170 +1,39 @@
...
+      onSubmit={properties.onChange}
+      title={t("num_fmt.title")}
+      className={properties.className}
+      defaultValue={properties.numFmt}
+      confirmLabel={t("num_fmt.save")}
+      inputProps={{
+        name: "format_code",
+        autoFocus: true,
+        spellCheck: false,
+        onFocus: (event) => event.target.select(),
```
nhatcher: What is this doing?

webapp/IronCalc/src/components/Modal/Alert.tsx
```diff
@@ -0,0 +1,74 @@
+import { X } from "lucide-react";
+import { type ReactNode, useId } from "react";
+import { createPortal } from "react-dom";
+import { Button } from "../Button/Button";
+import { IconButton } from "../Button/IconButton";
+import "./modal-dialog.css";
```
nhatcher: Please include this fiule only once.

webapp/IronCalc/src/components/Modal/useModalFocus.ts
nhatcher, line 1: Could you add a comment explaining what this file is doing and why we are using it?

MAybe with an example of what would hapen if we don't have it

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
```diff
@@ -0,0 +1,69 @@
...
+      }
+
+      if (event.key === "Enter") {
+        if (event.metaKey || event.ctrlKey) {
+          event.preventDefault();
+          onConfirm?.();
+          return;
+        }
+
+        // Prevent Enter from submitting when focus is not on a button
```
nhatcher: I don't understand this. What would happen if this is not in place?

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
```diff
@@ -0,0 +1,69 @@
...
+
+      if (event.key === "Tab") {
+        const modal = modalRef.current;
+        if (!modal) {
+          return;
+        }
+
+        const focusable = Array.from(
+          modal.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTORS),
+        );
```
nhatcher: We do have a list of the focusable elements, we could pass this list so that tab goes in the right order and we remove this code.

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        if (focusable.length === 0) {
+          event.preventDefault();
+          return;
+        }
+
+        const first = focusable[0];
```
nhatcher: This is interesting you are only using first and last of the "focusable" and leting the browser do it's thing for the rest. So you could only pass these two. In general:
Alert: [X, Ok]
Prompt: [X, Input, Cancel, Confirm]
Confrim: [X, Cancel, Confirm]

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
```diff
@@ -0,0 +1,69 @@
...
+        const last = focusable[focusable.length - 1];
+
+        if (event.shiftKey) {
+          if (document.activeElement === first) {
+            event.preventDefault();
+            last?.focus();
```
nhatcher: last *always* exists

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
```diff
@@ -0,0 +1,69 @@
...
+          }
+        } else if (document.activeElement === last) {
+          event.preventDefault();
+          first?.focus();
```
nhatcher: first *always* exists

nhatcher (changes requested):
Looks great, thanks! I left some minor comments.

webapp/IronCalc/src/components/Modal/modal-dialog.css
dg-ac: Correct.

webapp/IronCalc/src/components/FormatMenu/FormatMenu.tsx
dg-ac: We were passing this from `Toolbar.tsx` where we had `onExited={(): void => {}}`. This is not really doing anything and seems like a leftover, but I may be wrong?

webapp/IronCalc/src/components/Modal/Confirm.tsx
dg-ac: Noted, I also prefer it that way.

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
dg-ac: Added!

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
dg-ac: Added, thanks.

webapp/IronCalc/src/components/Modal/modal-dialog.css
```diff
@@ -0,0 +1,71 @@
+.ic-modal-dialog-backdrop {
+  position: fixed;
+  inset: 0;
```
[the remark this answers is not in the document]
nhatcher: Don't do that, we are moving _away_ from `z-index` and use "stacked" html. We will see if we can but as of now I see no reason why we shouldn't.

webapp/IronCalc/src/components/Modal/useModalFocus.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  useEffect(() => {
+    if (!open) {
+      return;
+    }
+
+    previousFocusRef.current = document.activeElement as HTMLElement | null;
+
+    requestAnimationFrame(() => {
+      modalRef.current?.focus();
```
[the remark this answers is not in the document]
nhatcher: What would be the difference?

webapp/IronCalc/src/components/Modal/Prompt.tsx
```diff
@@ -0,0 +1,124 @@
...
+            (typeof message === "string" ? <p>{message}</p> : message)}
+          <Input
+            {...inputProps}
+            value={value}
+            onChange={(event) => setValue(event.target.value)}
+            onKeyDown={(event) => {
+              event.stopPropagation();
+              if (event.key === "Escape") {
+                event.preventDefault();
+                closeModal();
```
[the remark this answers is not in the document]
nhatcher: True. It might be ok, though

webapp/IronCalc/src/components/Modal/Confirm.tsx
```diff
@@ -0,0 +1,92 @@
+import { X } from "lucide-react";
+import { type ReactNode, useId } from "react";
+import { createPortal } from "react-dom";
+import { Button } from "../Button/Button";
+import { IconButton } from "../Button/IconButton";
```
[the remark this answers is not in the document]
nhatcher: That's true. I I don't like any of those names. They are both Modals and Dialogs.

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
dg-ac: When we click on the input, it selects all the text inside.

webapp/IronCalc/src/components/Modal/Prompt.tsx
```diff
@@ -0,0 +1,124 @@
...
+        onKeyDown={onKeyDown}
+        role="dialog"
+        aria-modal="true"
+        aria-labelledby={titleId}
+        tabIndex={-1}
+      >
+        <div className="ic-modal-dialog-header">
+          <h2 id={titleId}>{title}</h2>
+          <IconButton icon={<X />} aria-label="Close" onClick={closeModal} />
+        </div>
```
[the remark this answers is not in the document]
nhatcher: Good catch!

webapp/IronCalc/src/components/Modal/Alert.tsx
dg-ac: Yes I'll add an `index.ts` file for that.

webapp/IronCalc/src/components/Modal/Confirm.tsx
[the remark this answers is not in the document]
dg-ac: @nhatcher both are widely used but in the end I leaned to modal. I would keep it like that tbh but let me know if you prefer to revert back to Dialog.

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
dg-ac: Sorry, added the comment in the wrong place.

But this is removed, see the comment about the focus below.

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
nhatcher: Is it? I can't find the comment there. BTW, you should also remove it from here

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
dg-ac: You're right. I've changed this a bit so we only consider first and last elements and we rely on the browser for the rest.

I've removed `FOCUSABLE_SELECTORS` and `querySelectorAll` as I don't think we need them anymore.

Then we mark which is the first and last button for each one of the 3 components (for example, in `Prompt.tsx` we use `closeButtonRef` and `submitButtonRef`) so the focus knows when to star and end.

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
dg-ac: Fixed in the next version.

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
dg-ac: Fixed in the next version.

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
dg-ac: This is to avoid applying forms by hitting `Enter`. It's a way to avoid accidental submits.
To submit changes in modals you need to use `cmd/ctrl+Enter`. This is how this comment editor in GitHub works, for example (although I know the use case is different).

You can, on the other hand, hit `Enter` if a button is focused, of course.

webapp/IronCalc/src/components/Modal/useModalFocus.ts
dg-ac: Added in the next version

dg-ac:
Just pushed some comments:

- Moved `stopPropagation` events to the `Input.tsx` component
- Simplified `useModalKeyDown.ts` to consider only first and last items
  - Marked first and last items in all 3 components
- Added a comment explaining `useModalFocus.tsx`
- Added an `index.ts` file to avoid importing the css file in 3 places
  - Updated root in `SheetDeleteDialog.tsx` and in components

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
nhatcher: I still don't understand.
Could you be more specific:

1. Open the Alert dialog
2. Focus on the X
3. Hit Enter

This would happen but what should happen is blah, blah.

Also in html there are [forms](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form) that have buttons of type="submit", but I don't think this applies to us as we do not use forms.

webapp/IronCalc/src/components/Modal/Prompt.tsx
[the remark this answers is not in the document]
dg-ac: Let me change it in the next version.

webapp/IronCalc/src/components/Modal/Alert.tsx
```diff
@@ -0,0 +1,74 @@
...
+        onKeyDown={onKeyDown}
+        role="dialog"
+        aria-modal="true"
+        aria-labelledby={titleId}
+        tabIndex={-1}
+      >
+        <div className="ic-modal-dialog-header">
+          <h2 id={titleId}>{title}</h2>
+          <IconButton icon={<X />} aria-label="Close" onClick={closeModal} />
+        </div>
```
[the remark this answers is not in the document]
dg-ac: This is the same comment in a separate file. Will be fixed in the next push.

webapp/IronCalc/src/components/Modal/Confirm.tsx
```diff
@@ -0,0 +1,92 @@
...
+        onKeyDown={onKeyDown}
+        role="dialog"
+        aria-modal="true"
+        aria-labelledby={titleId}
+        tabIndex={-1}
+      >
+        <div className="ic-modal-dialog-header">
+          <h2 id={titleId}>{title}</h2>
+          <IconButton icon={<X />} aria-label="Close" onClick={closeModal} />
+        </div>
```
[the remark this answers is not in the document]
dg-ac: This is the same comment in a separate file. Will be fixed in the next push.

webapp/IronCalc/src/components/Modal/Prompt.tsx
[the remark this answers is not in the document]
dg-ac: That's intentional... the focus trap only fires on the first and last elements.

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
nhatcher: Doesn't it do that atomatically?
Can we add all those properties to the prompt? (name, autoFocus, spellCheck and onFocus)

also do we need the name?

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
```diff
@@ -1,170 +1,39 @@
...
+        onPaste: (event) => event.stopPropagation(),
+        onCopy: (event) => event.stopPropagation(),
```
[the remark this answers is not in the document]
dg-ac: Makes sense. Added

webapp/IronCalc/src/components/Input/Input.tsx
```diff
@@ -84,8 +84,14 @@ export const Input = forwardRef<HTMLInputElement, InputProperties>(
             required={required}
             aria-invalid={error || undefined}
             aria-describedby={helperText ? helperId : undefined}
+            // FIXME: the stopPropagation everywhere is because of my bad implementation
```
nhatcher: Could you include my name in the comment please?
```
// FIXME: the stopPropagation everywhere is because of my (Nicolás Hatcher) bad implementation of keyboard handling in the spreadsheet
```

nhatcher: (I'll add an issue later and reffer to it in the code)

dg-ac: Sure. Done!

webapp/IronCalc/src/components/Modal/Confirm.tsx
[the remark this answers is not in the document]
dg-ac: Fixed.

webapp/IronCalc/src/components/Modal/Confirm.tsx
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  const handleConfirm = (): void => {
+    onConfirm();
+    closeModal();
+  };
+
+  const { onKeyDown } = useModalKeyDown({
+    first: closeButtonRef,
+    last: confirmButtonRef,
```
nhatcher: `first` and `last` are a bit akward names. I _think_ the way I would do it is something like:

```ts
focusableElements: [one, two, three]
```
Then you can use tab to cycle in th orther you want. In the present state if you specify the wrong first and last, the focus it might still get out of the dialog.

nhatcher (approved):
LGTM!

webapp/IronCalc/src/components/Modal/Confirm.tsx
dg-ac: I've replaced first/last with `focusableElements: RefObject<HTMLElement | null>[];`. Like in here:

**before**
```
useModalKeyDown({ first: closeButtonRef, last: confirmButtonRef, ... })
```
**after**
```
useModalKeyDown({ focusableElements: [closeButtonRef, confirmButtonRef], ... })
```

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
nhatcher: ping

webapp/IronCalc/src/components/Modal/useModalFocus.ts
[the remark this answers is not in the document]
nhatcher: ping

webapp/IronCalc/src/components/Modal/Confirm.tsx
nhatcher: That's good enough, but let's discuss this on a call.

webapp/IronCalc/src/components/Modal/useModalFocus.ts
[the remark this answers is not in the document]
dg-ac: I'd keep it as it is, I've tried it and there's no real benefit.

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
dg-ac: OK I gave it a 2nd thought and removed this part. `Prompt`'s input already handles `Enter` itself and stops it from reaching this handler. And even without that pressing `Enter` on a focused div does nothing by default...

webapp/IronCalc/src/components/FormatMenu/FormatPicker.tsx
dg-ac: Makes sense. I've added `spellCheck` to `Input.tsx` and removed `name` (not needed since there's no form) and `autoFocus` (the input already selects all text on focus via `onFocus` in `Input.tsx`).

webapp/IronCalc/src/components/Modal/useModalFocus.ts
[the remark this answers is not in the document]
nhatcher: Yeah, I'm not even sure what's the point.

webapp/IronCalc/src/components/Modal/useModalKeyDown.ts
nhatcher: ok

PR #873, opened 2026-04-16
FIX: editor font not themed

jvdp:
I'm deploying IronCalc without the Inter font (configuring another via theme variable instead) and the cell editor showed up as Times. This change fixes it.


nhatcher (approved):
Lgtm!

PR #881, opened 2026-04-22
fix: import css from all 3 modal types

dg-ac:
This PR should fix the issue we're facing where the css is not properly imported for the modals.


nhatcher (approved):
Lgtm!

PR #883, opened 2026-04-23
Implement INFO function

elsaminsut:
### Summary
This PR implements the INFO function and introduces a build.rs script that embeds version/build metadata at compile time. Previously INFO was not implemented and just returned `#NIMPL!` for all arguments.

### Changes
`build.rs` (new file)
A Cargo build script that determines the version string at compile time and exposes it as the GIT_VERSION environment variable. The version is found by running `git describe` in two passes:
- Exact tag (`git describe --tags --exact-match`). Succeeds when the current commit is directly tagged (e.g. _v0.7.1_).
- Dirty/always fallback (`git describe --tags --dirty --always`). Produces a string like _v0.7.1-15-g0703e60_ including version tag, number of commits ahead and commit id or _v0.7.1-15-g0703e60-dirty_ when there are uncommitted changes.
- If git is unavailable, the value falls back to "unknown".

`base/src/functions/information.rs`
type_text | Description
-- | --
"RELEASE" | Git-derived version string from env!("GIT_VERSION")
"SYSTEM" | Output of `std::env::consts::OS` on native targets or "browser" on WASM. 
"RECALC" | "Automatic" As of now, IronCalc always recalculates automatically
"NUMFILE" | Count of worksheets in the current workbook (similarly to SHEETS function)
"DIRECTORY", "ORIGIN", "OSVERSION" | #VALUE! (not yet implemented)

`base/src/test/test_cell_info_n_sheets.rs`
The test for INFO("recalc") previously expected #NIMPL!. It now asserts the correct return value "Automatic".

### DIFF
--- a/base/build.rs
+++ b/base/build.rs
@@ -17,20 +17,5 @@ fn main() {
         .or_else(|| run_git(&["describe", "--tags", "--dirty", "--always"]))
         .unwrap_or_else(|| "unknown".into());
 
-    let commit = run_git(&["rev-parse", "HEAD"]).unwrap_or_else(|| "unknown".into());
-    let branch =
-        run_git(&["rev-parse", "--abbrev-ref", "HEAD"]).unwrap_or_else(|| "unknown".into());
-
-    let dirty = run_git(&["status", "--porcelain"])
-        .map(|s| if s.is_empty() { "false" } else { "true" })
-        .unwrap_or("unknown");
-
-    let build_time = {
-        use std::time::{SystemTime, UNIX_EPOCH};
-        let secs = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .unwrap()
-            .as_secs();
-        secs.to_string()
-    };
+    println!("cargo:rustc-env=GIT_VERSION={}", version);
 }
--- a/base/build.rs
+++ b/base/build.rs
@@ -1,3 +1,5 @@
+// At build time, we grab the release version from the git tag.
+// This is used by INFO("release")
 use std::process::Command;
 
 fn run_git(args: &[&str]) -> Option<String> {
--- a/base/build.rs
+++ b/base/build.rs
@@ -1,0 +1,36 @@
+use std::process::Command;
+
+fn run_git(args: &[&str]) -> Option<String> {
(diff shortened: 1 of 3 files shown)


base/src/functions/information.rs
```diff
@@ -456,14 +466,36 @@ impl<'a> Model<'a> {
...
         }
-        CalcResult::Error {
-            error: Error::NIMPL,
-            origin: cell,
-            message: "Info function not implemented".to_string(),
+        let type_text = match self.get_string(&args[0], cell) {
+            Ok(s) => s.to_uppercase(),
+            Err(e) => return e,
+        };
+        let release = env!("GIT_VERSION");
```
nhatcher, line 481: Can you add a comment here of how this is done?

base/build.rs
```diff
@@ -0,0 +1,21 @@
+use std::process::Command;
```
nhatcher, line 3: Can you add a comment here about this file and why we do need it?

nhatcher (changes requested):

nhatcher (changes requested):
Minor issues but looks great

base/src/functions/information.rs
elsaminsut: Done! I added a reference to build.rs (and also the reference back over there)

nhatcher (approved):
LGTM!

For the future we could implement the osversion via the uaAgent:

```ts
#[cfg(target_arch = "wasm32")]
use wasm_bindgen::prelude::*;

#[cfg(target_arch = "wasm32")]
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(thread_local_v2, js_name = navigator)]
    static NAVIGATOR: JsValue;
}

#[cfg(target_arch = "wasm32")]
pub fn get_property(name: &str) -> Option<String> {
    use js_sys::{JsString, Reflect};
    NAVIGATOR.with(|nav| {
        Reflect::get(nav, &JsValue::from_str(name))
            .ok()
            .and_then(|v| v.dyn_into::<JsString>().ok())
            .map(|s| s.into())
    })
}

#[cfg(not(target_arch = "wasm32"))]
pub fn get_property(_name: &str) -> Option<String> {
    return None;
}
```

PR #886, opened 2026-04-24
update: allow confirm via `enter` in `Alert` and `Confirm` modals

dg-ac:
This PR should fix the issue that prevented `Alert` and `Confirm` modals from being closed when pressing the `Enter` key.

### DIFF
--- a/webapp/IronCalc/src/components/Modal/Alert.tsx
+++ b/webapp/IronCalc/src/components/Modal/Alert.tsx
@@ -39,7 +39,6 @@ export function Alert({
     focusableElements: [closeButtonRef, confirmButtonRef],
     onClose: closeModal,
     onConfirm: closeModal,
-    enterConfirm: true,
   });
 
   if (!open) {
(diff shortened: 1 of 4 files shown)


nhatcher:
Please rebase main and push again, the failing test is my fault

nhatcher (changes requested):
I don't think this is what should happen. What should happen is that the "ok" button has the focus of the Keyboard in those cases. So enter => click ok => closes

nhatcher:
As an aside:
RANDARRAY can generate integers. I have a test:

```Rust
#[test]
fn test_randarray_whole_number() {
    let mut model = new_empty_model();
    model._set("A1", "=RANDARRAY(1,1,1,6,TRUE)");
    model.evaluate();
    let val: f64 = model._get_text("A1").parse().unwrap();
    assert_eq!(val, val.floor());
    assert!((1.0..6.0).contains(&val));
}
```

That generates an array of ints in between 1 and 6 (including both). RANDARRAY (like RAND or RANDBTWEEN) are _volatile_ functions. Meaning their result changes every time you run them. Thats why testing them is a bit more tedious. You can't do:
```
assert_eq!(get_random_number(), 42)
```
as every time the number is going to be different. So yo tests properties of the function. In this case I was testing that the number is between 0 and 6 (excluding 6). The test passes 5 out of 6 times...
It is a britle tests, and it is the wrost kind.

It is fixed in main where the last assert is:

```Rust
assert!((1.0..=6.0).contains(&val));
```

dg-ac:
@nhatcher Thanks for the explanation and the idea. I've reverted changes and added `autoFocus` to the OK buttons so we can close the modals just hitting `enter`. What do you think?

nhatcher (approved):
LGTM!

PR #888, opened 2026-04-25
xlsx import: tolerate whitespace text nodes in worksheet rels

duane1024:
`load_sheet_relations` iterated every child of `<Relationships>` and called get_attribute(&rel, "Type") on each, which panicked on the trailing-newline text node that some writers (e.g. LibreOffice, which produced SUM1991S.xlsx) leave between `<Relationship/>` and `</Relationships>`.

Filter children to elements named "Relationship", matching the pattern already used by `load_relationships` in import/mod.rs. Added SUM1991S.xlsx to xlsx/tests/calc_tests/ as a regression (picked up automatically by xlsx/build.rs as test_calc_SUM1991S).


nhatcher (changes requested):
This is great! and I wonder if it is a bug in the xml parser upstream

Could you:

1. Rename the file to something like `issue_888.xlsx` or `libreoffice_888_example.xlsx`
2. Move it to th xlsx/tests/ folder
3. Add a similar test to: https://github.com/ironcalc/IronCalc/blob/main/xlsx/tests/test.rs#L413-L441

```Rust
// This tests the `xl/worksheets/_rels/sheet*` are parsed corectly
fn test_libreoffice_example() {
    let mut model = load_from_xlsx("tests/issue_888.xlsx", "en", "UTC", "en").unwrap();
    model.evaluate();
}
```

duane1024:
Updated filename and added explicit unit test.

nhatcher:
Lint is failing. You need to run `cargo fmt`. Everything is great!

duane1024:
Lint formatting should be updated now.

nhatcher (approved):
Lgtm!

PR #890, opened 2026-04-27
update: implements CELL("filename") 

elsaminsut:
The already implemented CELL function now takes a "filename" argument which returns the filename in the format `[WorkbookName.xlsx]SheetName`. This implementation is tested in a unit test and the current xlsx is adapted to the current IronCalc implementation.

Some failing cases in CELL are already described in #533


nhatcher (approved):
LGTM!

PR #891, opened 2026-04-27
update: adds unit testing for GCD, LCM

elsaminsut:
This PR adds argument testing for GCD and LCM functions, basically testing that passing no arguments to the functions returns an `#ERROR!`. It also includes a few array smoke tests to check that the functions accept them as arguments.

A full xlsx test with many more test cases can be found in #757


nhatcher (approved):
LGTM!

PR #892, opened 2026-04-27
Make SUM and PRODUCT functions more consistent with Excel

elsaminsut:
_Note: This PR was originally #236 by @gian-hancock and has been rebased to current main._
___
This PR updates SUM and PRODUCT functions to match Excel's behavior for type conversion and range handling.

### Changes
**Core improvements:**
- Both functions now convert non-numeric direct arguments to numbers (e.g., `=SUM("100")` returns `100`, `=PRODUCT(TRUE)` returns `1`)
- Important: Values within ranges or cell references are NOT converted, matching Excel's behavior
- Errors are properly propagated in all cases

**Implementation:**
- Made `cast_to_number()` public (`pub(crate)`) so both functions use the same conversion logic
- References are converted to single-cell range equivalents for consistent handling
- Added comprehensive docstrings documenting the special cases

### Test Coverage
Added 12 unit tests covering:
- Text string conversion (including scientific notation)
- Invalid text error handling
- Range/reference special cases (no conversion)
- Boolean value handling (`TRUE`=1, `FALSE`=0)
- INDIRECT function behavior

### Excel Compatibility
This aligns IronCalc with Excel's quirky but important behavior: direct arguments get converted to numbers, but values within ranges don't. This matters for real-world spreadsheet interoperability.


nhatcher (approved):
LGTM!

PR #894, opened 2026-04-28
Treat unset font color as None instead of black (import + engine default)

duane1024:
## Summary

Fix two related sources of font-color round-trip ambiguity, so that "user never set a color" is no longer indistinguishable from "user explicitly chose black."

- **XLSX import**: a `<font>` with no `<color>` child now imports as `Font.color = None`, matching the existing `<color auto="1"/>` behavior. Previously it was collapsed to `Some("#000000")`.
- **Engine default**: `Font::default()` now uses `color: None` instead of `Some("#000000")`. Fresh, non-imported workbooks no longer emit a spurious `<color rgb="FF000000"/>` on export.
- **Frontend**: `CellStyleFont.color` is now optional (`color?: string`) to match the Rust `Option<String>`, with a fallback to the theme's default text color where the canvas reads it and to `#000000` where the toolbar consumes it.
- New regression test for the import case in `xlsx/tests/font_color_import.rs`.

## Why

In OOXML, three inputs are semantically equivalent — all mean "use the automatic/default color":

- `<font>` with no `<color>` child
- `<font><color auto="1"/>...</font>`

…and the engine's own notion of an unset font color belongs in the same bucket. The codebase was collapsing two of these to`Some("#000000")`, which:

1. Made them indistinguishable from an explicit `<color rgb="FF000000"/>` (a real user choice of black).
2. Caused round-trip fidelity regressions: a `<font>` with no `<color>`, or a fresh non-imported workbook's default font, came back out as `<font><color rgb="FF000000"/>...</font>`.
3. Diverged internally — the importer's `<color auto="1"/>` path already produced `None`, but the no-`<color>` path and the engine default did not.

Behavior matrix after this PR:

| Input                                | Before                  | After                   |
|--------------------------------------|-------------------------|-------------------------|
| `<font>` (no color child)            | `Some("#000000")`       | `None` ✅                |
| `<font><color auto="1"/>`            | `None`                  | `None` (unchanged)      |
| `<font><color rgb="FF000000"/>`      | `Some("#000000")`       | `Some("#000000")` (unchanged) |
| `<font><color theme="..."/>`         | resolved RGB            | resolved RGB (unchanged) |
| `Font::default()` (engine)           | `Some("#000000")`       | `None` ✅                |

## Out of scope / not changed

- `<color theme="..."/>` resolution remains lossy on export (the resolved RGB is written, not the original theme reference). Pre-existing fidelity issue, separate concern.

## Test plan

- [x] `cargo test --workspace` — all tests pass (1182 base unit + 155 xlsx integration + others)
- [x] `make lint` — clean (strict clippy `-W clippy::unwrap_used -W clippy::expect_used -W clippy::panic -D warnings` plus biome on both webapp packages)
- [x] New `xlsx/tests/font_color_import.rs::test_font_with_no_color_element_is_none` — passes; verified to **fail** without the parser fix.
- [x] Manual round-trip: fresh `Model::new_empty()` workbook now exports a `<font>` with no `<color>` element and reloads as `Font.color == None`. Pre-fix it would have emitted `<color rgb="FF000000"/>` and reloaded as explicit black.
- [x] Manual round-trip: `PRODUCT_SUM.xlsx` (which contains a `<font>` with no `<color>` child) preserves that absence on save.


nhatcher (changes requested):
I think you forgot to commit a file?

nhatcher (changes requested):
Can you:

1. Move the test to a different file 
2. Allow expect used (see how it is done in other test files)

duane1024:
I fixed the issues that you mentioned, and also expanded the scope of this PR slightly; I felt that it would be better to address that gap rather than leave obviously different behavior in the xlsx import path vs. new IronCalc-native spreadsheet.  With respect to font color, I think that all of the paths that deal with font color should be consistent now, if the font color was not explicitly set, it should remain as null/None/not set.  In the webapp UI, it will default to #000000 (highest place that we can push out the logic).  It also works well for me now to automatically set a contrasting font color in my l123 TUI, instead of having an explicit color set that the UI has to work around.  Let me know if it looks ok!

duane1024:
Rebased with main and verified latest tests.  Let me know if you agree with the scope of the changes.

nhatcher:
@dg-ac, could you check this works as expected?

dg-ac (approved):
LGTM. I've tested it and it seems to work well.

The funny thing is that I couldn't change the default font color in Excel for Mac, it only worked on the Windows version. Anyway, once imported into IronCalc, it had the expected colour, so all good!

nhatcher (approved):
LGTM!

PR #896, opened 2026-04-28
FIX: few frontend ux and performance improvements

phanirithvij:
I was writing a nix derivation for IronCalc as part of [Nix@NGI team](https://nixos.org/community/teams/ngi/) effort. wip nixpkgs pr - https://github.com/NixOS/nixpkgs/pull/500348

During which I encountered these issues.

- 500 error when model is not found in the db instead of 404
- init_db.sql is not idempotent
- model doesn't load from the url when lang is set to `en-GB` in the UI. due to `en-GB` being handled specially in `getShortLocaleCode`
- ~~one of the tests was failing due to incorrect error message~~

### DIFF
--- a/webapp/IronCalc/src/components/Worksheet/usePointer.ts
+++ b/webapp/IronCalc/src/components/Worksheet/usePointer.ts
@@ -139,7 +139,7 @@ const usePointer = (options: PointerSettings): PointerEvents => {
       } else if (isInsertingRef.current) {
         const { workbookState } = options;
         const editingCell = workbookState.getEditingCell();
-        if (!editingCell || !editingCell.referencedRange) {
+        if (!editingCell?.referencedRange) {
           return;
         }
         const range = editingCell.referencedRange.range;
--- a/webapp/app.ironcalc.com/frontend/src/App.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/App.tsx
@@ -65,25 +65,32 @@ function App() {
       const languageId = getLanguageFromLocale(language);
       // If there is a model name ?model=modelHash we try to load it
       // if there is not, or the loading failed we load an empty model
+      let loadedModel: Model | null = null;
       if (modelHash) {
         // Get a remote model
         try {
           const model_bytes = await get_model(modelHash);
-          const importedModel = Model.from_bytes(model_bytes, languageId);
+          loadedModel = Model.from_bytes(model_bytes, languageId);
           localStorage.removeItem("selected");
-          setModel(importedModel);
         } catch (_e) {
+          console.error(_e);
+          alert(t("errors.model_not_found"));
           console.log("Failed to load model from hash:", modelHash);
         }
       } else if (exampleFilename) {
         try {
           const model_bytes = await get_documentation_model(exampleFilename);
-          const importedModel = Model.from_bytes(model_bytes, languageId);
+          loadedModel = Model.from_bytes(model_bytes, languageId);
           localStorage.removeItem("selected");
-          setModel(importedModel);
         } catch (_e) {
+          console.error(_e);
+          alert(t("errors.example_not_found"));
           console.log("Failed to load example model:", exampleFilename);
         }
+      }
+
+      if (loadedModel) {
+        setModel(loadedModel);
       } else {
         // try to load from local storage
         const result = loadSelectedModelFromStorage();
--- a/webapp/app.ironcalc.com/frontend/src/App.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/App.tsx
@@ -54,6 +54,7 @@ function App() {
 
   const { t, i18n } = useTranslation();
 
+  // biome-ignore lint/correctness/useExhaustiveDependencies: Run only for i18n.language dependency
   useEffect(() => {
     async function start() {
       await init();
@@ -123,6 +124,18 @@ function App() {
     }
   }, [model, localStorageId]);
 
+  useEffect(() => {
+    if (!model) return;
+    // We try to save the model every second
+    const interval = setInterval(() => {
+      const queue = model.flushSendQueue();
+      if (queue.length !== 1) {
+        saveSelectedModelInStorage(model);
+      }
+    }, 1000);
+    return () => clearInterval(interval);
+  }, [model]);
+
   if (!model) {
     return (
       <Loading>
@@ -131,14 +144,6 @@ function App() {
       </Loading>
     );
   }
-
-  // We try to save the model every second
-  setInterval(() => {
-    const queue = model.flushSendQueue();
-    if (queue.length !== 1) {
-      saveSelectedModelInStorage(model);
-    }
-  }, 1000);
 
   // Handlers for model changes that also update our models state
   const handleNewModel = () => {
--- a/webapp/app.ironcalc.com/frontend/src/App.tsx
+++ b/webapp/app.ironcalc.com/frontend/src/App.tsx
@@ -19,6 +19,7 @@ import {
   createNewModel,
   deleteModelByUuid,
   deleteSelectedModel,
+  getLanguageFromLocale,
   getShortLocaleCode,
   isStorageEmpty,
   loadDefaultLocaleFromStorage,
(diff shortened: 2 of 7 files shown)


nhatcher:
Thanks! There are some tests falling. Do you know how to fix this?

phanirithvij:
Will take a look soon

phanirithvij:
- I've fixed the formatting issue with the react App.tsx file.
- For the `test::user_model::test_to_from_bytes::errors` test I reverted my fix. I am not sure the reason but the bitcode dependency is giving a different error message `bitcode error` instead of `invalid packing` when I build with nix.
- I've added a way to allow the server to serve the frontend webapp by specifying `IRONCALC_WEBAPP_DIR` env var - this allows people to have a single process serving the UI along with the api, without the need for caddy/nginx etc., let me know if this is undesirable and I will drop it for this pr.

phanirithvij:
- I ran `cargo update -p atomic@0.7.0 --precise 0.6.1` because https://crates.io/crates/atomic/versions has no 0.7.0 version, and https://crates.io/crates/state/versions is also missing 0.7.0

nhatcher:
Hi @phanirithvij, happy to deal with this if you don't want to. Maybe what would be important is to add regression tests so these kind of things don't happen all the time.

phanirithvij:
- I ran `npm run check-write` on all three folders with biome checks in them.
- Fixed the copilot's suggestion for empty string checking.
- And logging the error to stderr and the UI now doesn't see the internal sqlx error.

phanirithvij:
I'm seeing some infinite loop and tab is getting stuck in the browser, after my recent changes, will try to investigate as soon as I am able to.

phanirithvij:
It is ready for review now @nhatcher

webapp/embed.ironcalc.com/example.html
```diff
@@ -7,22 +7,22 @@
```
nhatcher, line 7: Why are you chnaging this file? Can this be done ins a separate PR?

webapp/IronCalc/src/components/Worksheet/usePointer.ts
```diff
@@ -139,7 +139,7 @@ const usePointer = (options: PointerSettings): PointerEvents => {
       } else if (isInsertingRef.current) {
         const { workbookState } = options;
         const editingCell = workbookState.getEditingCell();
-        if (!editingCell || !editingCell.referencedRange) {
+        if (!editingCell?.referencedRange) {
```
nhatcher, line 142: Was this a linter issue?

webapp/app.ironcalc.com/frontend/src/App.tsx
```diff
@@ -52,37 +53,46 @@ function App() {
   };
 
   const { t, i18n } = useTranslation();
+  const isInitialized = useRef(false);
```
nhatcher: Why do you need this?

webapp/app.ironcalc.com/frontend/src/App.tsx
```diff
@@ -115,6 +128,18 @@ function App() {
     }
   }, [model, localStorageId]);
 
+  useEffect(() => {
```
nhatcher, line 127: Why do you need to put this inside a useEffect?

webapp/app.ironcalc.com/frontend/src/App.tsx
```diff
@@ -95,15 +105,18 @@ function App() {
           setModel(newModel);
         }
       }
-      i18n.changeLanguage(language);
+      if (i18n.language !== language) {
```
nhatcher: Why do you need this?

nhatcher (changes requested):
Hi @phanirithvij , I think we need most of your changes.
So... sorry to make you work extra. But could you split the PR:

1. Lint issues in embed (so you even need to fix those?)
2. server updates. Feel free to delete Cargo.lock and recreate it
3. Frontend updates. Here is the place where we might have some real discussions.

In the last point I would need some explanation. Bug fixes is the best. First give me a failing example and then do the fix. Otherwise to a React newbie like myself it looks like random changes :D

webapp/IronCalc/src/components/Worksheet/usePointer.ts
phanirithvij: yes it was suggested by biome check.

webapp/embed.ironcalc.com/example.html
phanirithvij: yes it was formatted when I ran npm run check-write in the embed.ironcalc.com directory.

I did it because I don't see github actions immediate response feedback when I push changes and having to wait for manual approval from you, so I did all the biome suggested fixes.

webapp/app.ironcalc.com/frontend/src/App.tsx
phanirithvij: Without this `useEffect` which runs only when the model gets updated and clearing the previous timer with the clearInterval call, the number of `setIntervals` will keep increasing in the background. It is a memory hog.

webapp/app.ironcalc.com/frontend/src/App.tsx
phanirithvij: This is to avoid an additional re-render if the current language is the same as the new one. I don't know if React-i18next already does it but I did it as one more performance fix.

webapp/app.ironcalc.com/frontend/src/App.tsx
phanirithvij: To solve the infinite loop happening due to introducing `t` in the `useEffect` call's dependencies.

This was failing in the ci https://github.com/ironcalc/IronCalc/actions/runs/25068830423/job/73481235110#step:6:49

Initially, I added `t` (the translation function) to the `useEffect` dependency array to fix the linter error. But that caused an infinite loop as App.tsx calls i18n.changeLanguage() inside that same effect.
In React-i18next, changing the language causes the t function to be recreated.

So the fix was to use a ref, I added a `useRef(false)` to track if the application has already loaded the model. If true, the effect exits immediately and Refs do not trigger re-renders.

webapp/embed.ironcalc.com/example.html
nhatcher: That's good. I haven't added yet to the CI. I'll do that shortly

webapp/IronCalc/src/components/Worksheet/usePointer.ts
nhatcher: ok. It does look ok. I wonder how it dis pass in other builds. Tomorrows problem :D

webapp/app.ironcalc.com/frontend/src/App.tsx
nhatcher: Thanks. This is the issue I will have to check carefully. (I'm 99.9% certain that you are right though, I just need to get this through my skull)

phanirithvij: Other fix is to do `// biome-ignore lint/correctness/useExhaustiveDependencies: Only run once on mount`

webapp/IronCalc/src/components/Worksheet/usePointer.ts
phanirithvij: It wasn't an error but a suggested fix I think.

webapp/app.ironcalc.com/frontend/src/App.tsx
nhatcher: Would taht remove the `isInitialized` variable? Then yes please do that

phanirithvij: Ok removed it.

webapp/app.ironcalc.com/frontend/src/App.tsx
phanirithvij: I've dropped this as well, to keep it simple.

webapp/app.ironcalc.com/server/src/main.rs
```diff
@@ -124,7 +123,15 @@ async fn upload(data: Data<'_>, name: &str) -> io::Result<Vec<u8>> {
 
 #[launch]
 fn rocket() -> _ {
-    rocket::build()
+    let mut rocket = rocket::build()
         .attach(IronCalcDB::init())
-        .mount("/", routes![upload, download, share, get_model, get_model_list])
+        .mount("/", routes![upload, download, share, get_model, get_model_list]);
+
+    if let Ok(frontend_path) = std::env::var("IRONCALC_WEBAPP_DIR") {
```
nhatcher, line 130: This is great!!!

nhatcher (approved):
Hi @phanirithvij, thanks for your changes and for bearing with me.

Let me know if you need any further asistance. Using Nix has been on our radar for at least a year!

webapp/app.ironcalc.com/server/src/main.rs
nhatcher: Hi @dg-ac, this means we don't need caddy during development anymore :)

phanirithvij:
Thanks for the very timely replies and reviews. After I got biome locally (`nix-shell -p biome`) it was smooth sailing.

> Using Nix has been on our radar..

Hopefully you can initially try nix for provisioning developer shell (cargo, nodejs, biome etc.) and see if it works out and depending on your experience you could slowly try to adopt it for CI and then maybe even production?! - All running the exact same versions of software, mathematically guaranteed. 

Personally I can't live without nix - even if it is not perfect, it is still the sanest way to capture all the dependencies.

PR #919, opened 2026-05-06
update: bump `lucide` to v1.14.0 and add `line-style` icon

dg-ac:
We were finally able to include [one of our icons](https://lucide.dev/icons/line-style), `line-style`, in Lucide's library! This PR removes the current custom-made svg and uses the one from the official library.

We are also bumping the library from `v0.562.0` to `v1.14.0` as we were a bit behind.


nhatcher (changes requested):
Please remove pnpm-lock.yaml file

PR #920, opened 2026-05-06
update: add unit testing for FACT, FACTDOUBLE and mark some functions as available in the docs

elsaminsut:
This PR adds unit testing in Rust to check the number of arguments taken by FACT and FACTDOUBLE functions. Also marks them as implemented in their respective documentation pages. 

While doing this, other implemented functions were found with the wrong description in their pages. Some functions in the following categories were fixed:
- Logical
- Mathematical and trigonometry
- Information
- Lookup and reference
- Dynamic array
- Text (some functions here did not even have their own page)

### DIFF
--- a/base/src/test/mod.rs
+++ b/base/src/test/mod.rs
@@ -13,6 +13,7 @@ mod test_datedif_leap_month_end;
 mod test_days360_month_end;
 mod test_degrees_radians;
 mod test_error_propagation;
+mod test_fact_factdouble;
 mod test_fn_average;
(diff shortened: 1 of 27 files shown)


nhatcher:
Can you split this in two?
Also why are you updating the docs? If you are can you just update all of them? There is an open PR for that

nhatcher (changes requested):
Can we split this?

Also As far as I know this two functions are blocked on the implementation, no?

nhatcher (approved):
LGTM!

PR #933, opened 2026-05-08
update[app]: remove dependencies

dg-ac:
We're finally getting rid of MUI and friends. Build size is almost half now!


nhatcher (approved):
LGTM!

Does the qr code work? (as in it doesn't have a peer dependency?)

dg-ac:
@nhatcher Yep it works but has some deps [1]. We can revisit this in the future, maybe we can get rid of it.

[1] https://www.npmjs.com/package/react-qr-code?activeTab=dependencies

PR #937, opened 2026-05-09
fix: overlapping chevrons in toolbar

dg-ac:
This PR fixes an annoying bug that was making the toolbar chevron buttons overlap unrelated elements, like dialogs and drawers. These buttons appear on narrow displays and help users reach all buttons in the toolbar.

IMG_6567 IMG_6568

We may want to revisit the chevron button solution in the future, but for now this just makes sure it's not in the way of other elements.


nhatcher:
Why can't you just remove the z-index as you suggested?

dg-ac:
@nhatcher I tried but it causes overlapping with the toolbar buttons.

nhatcher:
> @nhatcher I tried but it causes overlapping with the toolbar buttons.

Let's discuss then.

nhatcher (approved):
LGTM!

PR #939, opened 2026-05-12
fix: scalar-context conversion for arrays in non-spilling positions

wermelinger:
## Problem

When a formula that was statically analyzed as scalar produces a multi-cell array at runtime, `Model::evaluate_cell` falls into a `None` branch that:

1. Triggers `debug_assert!(false, ...)` — crashes debug builds.
2. In release, sets the cell to `#NIMPL!` and returns `Err(...)` from `evaluate_cell`, causing all dependent cells to cascade to `#ERROR!`.

This was reproducible loading a real-world workbook in our project. A single hard-to-predict cell tripped this path and the entire sheet became `#ERROR!`.

## Fix

Handle the case explicitly:

* **1×1 array** → unwrap to its single value (it really *is* scalar, just wrapped).
* **N×M array (N>1 or M>1)** → emit `#VALUE!`.

This matches what Excel and LibreOffice produce in the same situation, and aligns with [ODF OpenFormula](https://docs.oasis-open.org/office/OpenDocument/v1.4/os/v1.4-os.pdf) §6.3.2 ("Conversion to Scalar") and §6.3.3 ("Forced Recalculation / Implied Intersection").

`evaluate_cell` now returns `Ok(())` so dependents can compute normally instead of cascading.

## Verification

I used the existing `xlsx/src/bin/test.rs` comparator (which compares IronCalc-evaluated values against the cached Excel values inside the xlsx) on the failing workbook. Before this fix: thousands of cells diverged. After: only the cosmetic differences and `RAND()`-derived cells remain.

## Notes

I'm not a spreadsheet-engine expert; I worked through this with AI assistance to understand the parser and the implicit-intersection machinery. The change is intentionally minimal — it only replaces the "should not happen" assertion with a defined fallback. I haven't been able to construct a minimal Rust test that triggers this path because the parser's static analysis is otherwise quite good at preventing it; in our case it slipped through because of complex defined-name chains in the original workbook.

### DIFF
--- a/base/src/model.rs
+++ b/base/src/model.rs
@@ -896,9 +896,11 @@ impl<'a> Model<'a> {
                     // Scalar formula produced an array at runtime. We only coerce safely
                     // when the array is 1x1 (the result is genuinely a single value just
                     // wrapped in an array). For larger arrays, Excel would apply implicit
-                    // intersection (legacy) or auto-spill (dynamic arrays); neither is
-                    // implemented here, so picking [0][0] could silently produce wrong
-                    // results. Emit #VALUE! instead so the divergence is visible.
+                     // intersection (legacy) or, for formulas identified as dynamic/array
+                     // at parse time, auto-spill. In this `original_range == None` path we
+                     // do not have that array/dynamic context, so neither behavior is
+                     // available here; picking [0][0] could silently produce wrong results.
+                     // Emit #VALUE! instead so the divergence is visible.
                     let coerced = if array_width == 1 && array_height == 1 {
                         match self.get_value_from_array(array, 1, 1) {
                             Some(node) => array_node_to_formula_value(node),
--- a/base/src/model.rs
+++ b/base/src/model.rs
@@ -893,10 +893,28 @@ impl<'a> Model<'a> {
                     return Ok(());
                 }
                 None => {
-                    // this should not happen, but we need to handle it anyway
-                    // It means we got an array in a cell that we 'thought' could only return a scalar
-                    // It means we made a mistake in parse time
-                    debug_assert!(false, "Unexpected array result in non-array formula");
+                    // Scalar formula produced an array at runtime. We only coerce safely
+                    // when the array is 1x1 (the result is genuinely a single value just
+                    // wrapped in an array). For larger arrays, Excel would apply implicit
+                    // intersection (legacy) or auto-spill (dynamic arrays); neither is
+                    // implemented here, so picking [0][0] could silently produce wrong
+                    // results. Emit #VALUE! instead so the divergence is visible.
+                    let coerced = if array_width == 1 && array_height == 1 {
+                        match self.get_value_from_array(array, 1, 1) {
+                            Some(node) => array_node_to_formula_value(node),
+                            None => FormulaValue::Error {
+                                ei: Error::VALUE,
+                                o: "".to_string(),
+                                m: "Unexpected array result".to_string(),
+                            },
+                        }
+                    } else {
+                        FormulaValue::Error {
+                            ei: Error::VALUE,
+                            o: "".to_string(),
+                            m: "Array result in scalar context".to_string(),
+                        }
+                    };
                     *self.workbook.worksheets[sheet as usize]
                         .sheet_data
                         .get_mut(&row)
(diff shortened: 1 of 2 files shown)


base/src/model.rs
```diff
@@ -893,10 +893,28 @@ impl<'a> Model<'a> {
                     return Ok(());
                 }
                 None => {
-                    // this should not happen, but we need to handle it anyway
-                    // It means we got an array in a cell that we 'thought' could only return a scalar
-                    // It means we made a mistake in parse time
-                    debug_assert!(false, "Unexpected array result in non-array formula");
```
nhatcher, line 899: It is nice that the debug build crashes no?

nhatcher (changes requested):
Thank you for catching this!

Do you have an xlsx that will trigger the error?

It would be very nice to have a regression test

wermelinger:
FYI: I created these three PRs together to make our use-case work 
- #939 
- #940 
- #941

base/src/model.rs
wermelinger: The `debug_assert!` claims this branch is unreachable, but I have a real workbook that hits it — so keeping it would crash debug builds on legitimate input. 
The new fallback (1×1 → unwrap, larger → #VALUE!) should match Excel/LibreOffice for the same situation.

I can't share the original workbook, but I'll try to strip it down to a minimal reproducer and post it here.

nhatcher: > but I'll try to strip it down to a minimal reproducer and post it here.

That would be amazing!

wermelinger: In this file, we can see how Excel handles a situation like this: 
`=2 * IF(TRUE, OFFSET(B1, 2, 0), 0)`

* static analysis correctly classifies the whole expression as scalar (3-arg OFFSET of a single-cell ref has the shape of that cell)
* OFFSET returns a 1x1 array at runtime
* when such a 1×1 array reaches a scalar context (here: arithmetic / cell assignment), it should be transparently unwrapped

[array_in_scalar_repro.xlsx](https://github.com/user-attachments/files/27697372/array_in_scalar_repro.xlsx)

I also added this as a new automated test in test_implicit_intersection.rs: offset_returning_1x1_array_in_scalar_context

wermelinger: Also re-added a debug_assert! for the >1×1 case. 
With the split it now only triggers on the unexpected case (a larger array reaching a scalar-context cell). 
Mirrored in both set_cells_with_result and evaluate_cell (as pointed out by Copilot reviewer) so the two coercion sites stay in sync.

base/src/model.rs
```diff
@@ -905,13 +923,9 @@ impl<'a> Model<'a> {
...
-                        v: FormulaValue::Error {
-                            ei: Error::NIMPL,
-                            o: "".to_string(),
-                            m: "Unexpected array result".to_string(),
-                        },
+                        v: coerced,
                     };
-                    return Err("Unexpected array result".to_string());
+                    return Ok(());
                 }
```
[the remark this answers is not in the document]
wermelinger: Fixed: `evaluate_cell `now applies the same coercion as `set_cells_with_result `(1×1 → unwrap, >1×1 → #VALUE!) for non-array formulas. Added a regression test pinning the 1×1 dependent-observation behavior. The >1×1 path is currently unreachable (implicit intersection collapses such subexpressions), so a `debug_assert!` guards it in both sites.

base/src/model.rs
```diff
@@ -893,10 +893,28 @@ impl<'a> Model<'a> {
...
-                    // this should not happen, but we need to handle it anyway
-                    // It means we got an array in a cell that we 'thought' could only return a scalar
-                    // It means we made a mistake in parse time
-                    debug_assert!(false, "Unexpected array result in non-array formula");
+                    // Scalar formula produced an array at runtime. We only coerce safely
+                    // when the array is 1x1 (the result is genuinely a single value just
+                    // wrapped in an array). For larger arrays, Excel would apply implicit
+                    // intersection (legacy) or auto-spill (dynamic arrays); neither is
+                    // implemented here, so picking [0][0] could silently produce wrong
+                    // results. Emit #VALUE! instead so the divergence is visible.
```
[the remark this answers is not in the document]
wermelinger: applied

nhatcher (approved):
Amazing, thanks!

nhatcher:
Force merging as the lint issue is unrelated to this change

PR #941, opened 2026-05-12
feat: accept legacy STDEV / NORMINV function names

wermelinger:
## What

Accept the pre-Excel-2010 function names `STDEV` and `NORMINV` as aliases for `STDEV.S` and `NORM.INV` respectively. Both names are common in older workbooks and are fully documented by Microsoft as "compatibility functions".

* `STDEV(...)` → dispatched to the same evaluator as `STDEV.S`
* `NORMINV(...)` → dispatched to the same evaluator as `NORM.INV`

## Implementation

Added two extra lookup keys in `impl_function_lookup!` that point at the existing `Function::StDevS` / `Function::NormInv` variants. No new evaluators or formula semantics — the legacy names are pure aliases.

The translation entries are stored as `stdev_legacy` / `norminv_legacy` so the non-legacy names remain the ones used when a workbook is written back out.

## Tests

* `base/src/test/statistical/test_fn_stdev.rs::stdev_legacy_alias`
* `base/src/test/statistical/test_fn_norm_dist.rs::norminv_legacy_alias`

Both verify the alias produces the same value as the modern name.

## Why

Same project as the other PRs — the formulas use the legacy names; without them the sheet fails to evaluate.

## Notes

I'm not a spreadsheet-engine expert; I worked through this with AI assistance. If you'd like the alias mechanism done differently (e.g. a dedicated `Function::StDev` variant with a deprecation note instead of a direct lookup-key mapping), I'm happy to adjust.


wermelinger:
FYI: I created these three PRs together to make our use-case work 
- #939 
- #940 
- #941

nhatcher:
Ooops, sorry, I missed this. This is half done already. Will add NORMINV latrer

PR #942, opened 2026-05-12
update[app]: use Menu component in remaining menus

dg-ac:
Now that the `Menu` component is done we can replace the custom made we had in the app. This PR does the following changes:

- Refactor `FileMenu.tsx`
  - Added also locale codes as `secondaryText`
- Refactor `HelpMenu.tsx`
  - Added an 'About Ironcalc' link that leads to the website
- Refactor `WorkbookMenu.tsx`
  - Remove unnecessary code from `useWorkbookMenu.ts`
- Refactor `MobileMenu.tsx`

- Removed `navigation-menus.css, as it's not needed anymore
- Locales have been updated with the 'About' text
- It also adds a few imports to `index.ts`


nhatcher (approved):
LGTN!

PR #947, opened 2026-05-14
Fix: Preserve original date case during autofill

### LINKED ISSUE #902: Autofilling month names should keep original case
When writing day or month names in IronCalc and autofilling the values to next cells, the new values should match the user's original case. For example, jan feb mar should be completed with apr may jun, not Apr May Jun. 

### Reproduction
- Write jan in A1, feb in A2, mar in A3
- Drag the cell selector to A6
- The new cells now have Apr May Jun

Image

This also happens with day names and in different languages (in Spanish, day and month names are written in lowercase):

Image

Image

Image

rezat-c:
date autofill now preserves the users original letter case
  * jan, feb → mar, apr, ...
  * JANUAR, FEBRUAR → MÄRZ, APRIL, ...
  
month letters are no longer supported by autofill, aligning the behavior with Excel
added tests
refactoring

fix #902

### DIFF
--- a/base/src/test/user_model/test_autofill_columns.rs
+++ b/base/src/test/user_model/test_autofill_columns.rs
@@ -490,6 +490,93 @@ fn short_month_progression() {
 }
 
 #[test]
+fn short_month_progression_lowercase() {
+    let model = new_empty_model();
+    let mut model = UserModel::from_model(model);
+    model.set_user_input(0, 1, 1, "jan").unwrap(); // A1
+    model.set_user_input(0, 1, 2, "feb").unwrap(); // B1
+    model
+        .auto_fill_columns(
+            &Area {
+                sheet: 0,
+                row: 1,
+                column: 1,
+                width: 2,
+                height: 1,
+            },
+            4,
+        )
+        .unwrap();
+
+    assert_eq!(
+        model.get_cell_content(0, 1, 3), // C1
+        Ok("mar".to_string())
+    );
+    assert_eq!(
+        model.get_cell_content(0, 1, 4), // D1
+        Ok("apr".to_string())
+    );
+}
+
+#[test]
+fn short_month_progression_uppercase() {
+    let model = new_empty_model();
+    let mut model = UserModel::from_model(model);
+    model.set_user_input(0, 1, 1, "JAN").unwrap(); // A1
+    model.set_user_input(0, 1, 2, "FEB").unwrap(); // B1
+    model
+        .auto_fill_columns(
+            &Area {
+                sheet: 0,
+                row: 1,
+                column: 1,
+                width: 2,
+                height: 1,
+            },
+            4,
+        )
+        .unwrap();
+
+    assert_eq!(
+        model.get_cell_content(0, 1, 3), // C1
+        Ok("MAR".to_string())
+    );
+    assert_eq!(
+        model.get_cell_content(0, 1, 4), // D1
+        Ok("APR".to_string())
+    );
+}
+
+#[test]
+fn short_month_progression_left_uppercase() {
+    let model = new_empty_model();
+    let mut model = UserModel::from_model(model);
+    model.set_user_input(0, 1, 10, "JAN").unwrap(); // J1
+    model.set_user_input(0, 1, 11, "FEB").unwrap(); // K1
+    model
+        .auto_fill_columns(
+            &Area {
+                sheet: 0,
+                row: 1,
+                column: 10,
+                width: 2,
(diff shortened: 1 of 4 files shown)


rezat-c:
Excel recognizes these:

* "March. "
* "April. "

or these:

* "March "
* "April "

as months and can continue autofill.

But these:

* "March ."
* "April ."

do not work.

Right now I'm just trimming '.' or ' ' from the beginning and end

elsaminsut:
Hi! I just tested your branch and found a few cases where the autofill doesn't properly work:
- In Spanish and French (where the correct names are `enero, febrero, ...` in lowercase) using capitalized names `Enero, Febrero` autofills to their lowercase names anyway
- In German and French, short month names don't autofill: `JAN, FEB` becomes `JAN, FEB, JAN, FEB`. Italian & Spanish work fine
- And maybe a minor thing but in French `Janvier, Fevrier` doesn't autofill if `Février` is missing its accent mark

You can find these cases I tested here: https://testing.ironcalc.com/?model=stVF0-l4QpB-NJwgk

rezat-c:
> Hi! I just tested your branch and found a few cases where the autofill doesn't properly work:
> 
>     * In Spanish and French (where the correct names are `enero, febrero, ...` in lowercase) using capitalized names `Enero, Febrero` autofills to their lowercase names anyway
> 
>     * In German and French, short month names don't autofill: `JAN, FEB` becomes `JAN, FEB, JAN, FEB`. Italian & Spanish work fine
> 
>     * And maybe a minor thing but in French `Janvier, Fevrier` doesn't autofill if `Février` is missing its accent mark
> 
> 
> You can find these cases I tested here: https://testing.ironcalc.com/?model=stVF0-l4QpB-NJwgk

Hi!

Many thanks for testing
Fixed

Note that in your table, in columns O and P, there are "jan" and "fév" as short French month names, while in [locales.json](https://github.com/ironcalc/IronCalc/blob/main/generate_locale/locales.json#L493) they are "janv" and "févr"

Also, in Excel, `day_names_short` autofills dates with a trailing ".", while `months_short` does not. In [`locales.json`](https://github.com/ironcalc/IronCalc/blob/main/generate_locale/locales.json), both include a ".". Should I reproduce this behavior?

elsaminsut:
Hey, thanks for the fixes! 

About French short month names - you're completely right, my bad! I've been checking the CLDR standard (for example, in  [German](https://github.com/unicode-cldr/cldr-dates-modern/blob/master/main/de/ca-gregorian.json)) and I see there are some short forms with a trailing period and some without. I also see that IronCalc mixes both `stand-alone/abbreviated` and `stand-alone/short` forms. Using `stand-alone` is the right choice for this case, as opposed to `format` forms, which are used within whole dates (like 12. Feb. 2026).

Honestly this feels very confusing but let's just go with one form: `stand-alone/abbreviated`. Also making sure that the sequence detector works both if the user types a period in these languages or not. In case of doubt, we can always follow what Excel does!

Hope this all makes sense to you - happy to have a chat if it doesn't though!

nhatcher:
@rezat-c , @elsaminsut what is the status of this?

rezat-c:
> @rezat-c , @elsaminsut what is the status of this?

I think there's a small issue where `months_short` are auto-filled with a period at the end, even though they shouldn't be

To fix this, I think we need to add stand-alone dates to [locales.json](https://github.com/ironcalc/IronCalc/blob/main/generate_locale/locales.json#L493)

elsaminsut:
Changing the locales file might not be necessary. Checking out what Excel does in these languages, it always follows convention: a user input of `janv, févr` (no period) will be completed with `mars, avr.` (with period). I think this is a rare edge case and we can follow Excel here. 

But in any case I think we should **detect** the day/month even if written without its trailing period and complete the sequence in the proper locale form from `locales.json`

elsaminsut:
After speaking with @nhatcher we realized we _might_ have to make changes to `locale.json` 😅

So please forget about the trailing period for now, we can approve these changes here and tackle that in another PR. 

Btw I still see that `Enero, Febrero` autofills to `marzo, abril`, returning the months as they appear in the json and not following user input. Once that's fixed, we can merge this one!

rezat-c:
> Btw I still see that `Enero, Febrero` autofills to `marzo, abril`, returning the months as they appear in the json and not following user input. Once that's fixed, we can merge this one!

Are you sure? I can't reproduce it

elsaminsut (approved):
My bad, everything works perfectly thanks!

LGTM

PR #962, opened 2026-05-18
fix: autofill on double-click

### LINKED ISSUE #911: Autofill: double-clicking cell handle overwrites cells instead of autofilling
When double-clicking the fill handle to autofill a range, all cells are filled with the first value instead of continuing the pattern or sequence. Double-clicking should autofill down to match the length of an adjacent range. Manually dragging the cell handle downwards correctly fills the range with the right values.

### Reproduction
1. In column A enter a range of data with 6 values like A, B, C, D, E, F
2. In column B, enter a sequence 1, 2, 3
3. Double-click the cell handle
4. The range is now autofilled with ones instead of following the sequence 

https://github.com/user-attachments/assets/166ba7cf-0e3c-4826-91f5-aed9789aec5b

rezat-c:
Use selected range for double-click autofill instead of repeating the first value.

fix #911

### DIFF
--- a/webapp/IronCalc/src/components/WorksheetCanvas/outlineHandle.ts
+++ b/webapp/IronCalc/src/components/WorksheetCanvas/outlineHandle.ts
@@ -206,6 +206,20 @@ export function attachOutlineHandle(
       lastUsedRow = r;
     }
 
+    for (let r = rowEnd + 1; r <= lastUsedRow; r += 1) {
+      let isAnyCellNotEmpty = false;
+      for (let c = columnStart; c <= columnEnd; c += 1) {
+        if (worksheet.model.getFormattedCellValue(sheet, r, c) !== "") {
+          isAnyCellNotEmpty = true;
+          break;
+        }
+      }
+      if (isAnyCellNotEmpty) {
+        lastUsedRow = r - 1;
+        break;
+      }
+    }
+
     if (lastUsedRow <= rowEnd) {
       return;
     }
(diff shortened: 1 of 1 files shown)


dg-ac:
Hey @rezat-c, just tested this and it's looking great, we're almost there. I've noticed 2 issues though:

1. Let's say we select 3 cells and autofill them by double click. After the double click, we keep the original selection (the 3 cells) instead of expanding the selection itself so it covers all newly filled cells. 
I think **it'd expand the selection**. Excel for Mac does it this way (but not on Excel for Web!). IMO this is handy as you can right away style or edit these new cells.

2. Right now we overwrite those cells that are on the way.
I think we should **stop the autofill as soon as there's content in a cell**.

I've made a video where both issues are shown. Let me know what you think!

https://github.com/user-attachments/assets/7b1fddf7-3f7f-4396-888f-b0cddd1af2ab

> @nhatcher @elsaminsut feel free to jump in if you have arguments against what I said in n1.

rezat-c:
> Hey @rezat-c, just tested this and it's looking great, we're almost there. I've noticed 2 issues though:
> 
>     1. Let's say we select 3 cells and autofill them by double click. After the double click, we keep the original selection (the 3 cells) instead of expanding the selection itself so it covers all newly filled cells.
>        I think **it'd expand the selection**. Excel for Mac does it this way (but not on Excel for Web!). IMO this is handy as you can right away style or edit these new cells.
> 
>     2. Right now we overwrite those cells that are on the way.
>        I think we should **stop the autofill as soon as there's content in a cell**.
> 
> 
> I've made a video where both issues are shown. Let me know what you think!
> autofill.mov
> 
> > @nhatcher @elsaminsut feel free to jump in if you have arguments against what I said in n1.

1. I thought about that, but decided to mirror the behavior of the web version

fixed both issues

elsaminsut (approved):
Thanks for the fixes! LGTM

PR #963, opened 2026-05-18
chore: exclude nodejs from coverage

rezat-c:


nhatcher:
Hi, why do you want to do this?

rezat-c:
> Hi, why do you want to do this?

hi
on [codecov](https://app.codecov.io/gh/ironcalc/IronCalc), `bindings/nodejs/src` shows 0% coverage
if I understand correctly, it is already tested outside Rust (`bindings/nodejs/__test__/index.spec.mjs`), similar to `bindings/wasm`, which has already been excluded

nhatcher (approved):
Thanks!

PR #985, opened 2026-05-20
update: add testing for GAMMA functions

elsaminsut:
This PR adds unit tests for the GAMMA functions (GAMMA, GAMMA.DIST and GAMMA.INV), including some smoke tests, domain tests and checks for the right number of arguments.

While testing I found that booleans were not accepted as a valid input so a fix for that is included here as well. 

Some failing test cases were found while testing, see #986


nhatcher (approved):
LGTM!

nhatcher:
NB: This PR also fixes some of the issues found elsewhere

PR #991, opened 2026-05-20
fix: submenu behavior and menu anchor positioning

### LINKED ISSUE #970: [App]: Language switch broken in main
Hi @dg-ac, I think changing the Language does not work anymore since the Menu updates.
Do you think you can fix that?

dg-ac:
This PR fixes 2 issues we were having with the Menu component:

1. Submenu items were not applying changes: clicking items inside a submenu was closing the parent menu before the `onClick` handler could fire, because clicks inside nested menus were incorrectly treated as outside clicks. 

2. Submenus were not properly aligned when placed near the bottom of the viewport

### Changes:

- `Menu.tsx`: ignore outside-click events when the target is inside a nested menu wrapper
- `MenuItem.tsx`: pass `flipY` anchor coordinate when computing submenu position
- `useAnchorPosition.ts`: handle `flipY` to align submenu bottom edge when near the viewport bottom
- `Menu.stories.tsx`: improve submenu stories so we can catch issues with clicking and positioning

After the fix, this how the submenu is aligned when near the bottom edge:

image

---

Fixes #970

### DIFF
--- a/webapp/IronCalc/src/components/Menu/Menu.tsx
+++ b/webapp/IronCalc/src/components/Menu/Menu.tsx
@@ -93,8 +93,13 @@ export function Menu(props: MenuProperties) {
         ? (triggerPosition.triggerRef.current?.contains(target) ?? false)
         : false;
 
-      const insideAnyMenu = (target as Element).closest?.(".ic-menu-wrapper") !== null;
-      if (!triggerContains && !(menuRef.current?.contains(target) ?? false) && !insideAnyMenu) {
+      const insideAnyMenu =
+        (target as Element).closest?.(".ic-menu-wrapper") !== null;
+      if (
+        !triggerContains &&
+        !(menuRef.current?.contains(target) ?? false) &&
+        !insideAnyMenu
+      ) {
         close();
       }
     }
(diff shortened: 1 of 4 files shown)


webapp/IronCalc/src/components/Menu/Menu.tsx
```diff
@@ -93,7 +93,13 @@ export function Menu(props: MenuProperties) {
...
         : false;
 
-      if (!triggerContains && !(menuRef.current?.contains(target) ?? false)) {
+      const insideAnyMenu =
+        (target as Element).closest?.(".ic-menu-wrapper") !== null;
+      if (
+        !triggerContains &&
+        !(menuRef.current?.contains(target) ?? false) &&
+        !insideAnyMenu
+      ) {
```
dg-ac, line 103: @nhatcher This is the fix for the broken submenu. For reference, submenus are rendered as separate DOM elements, so the parent menu didn't recognize clicks inside them and  would close before any action was triggered. 
With this fix we don't close the menu if the click happens inside any other menu (which we identify with the class `.ic-menu-wrapper`). Let me know what you think.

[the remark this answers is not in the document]
dg-ac: Fixed.

nhatcher: Hmm, let's discuss this. The problem with this is that introduces coupling between the submenu and a class name. Somebody might come two years from now change the name of the class snd all of the sudden everything is broken

nhatcher (approved):

PR #996, opened 2026-05-21
fix: MMULT edge cases (empty/boolean inputs, overflow, omitted args)

wermelinger:
This is a follow-up PR for MMULT (merged in https://github.com/ironcalc/IronCalc/pull/940) supporting missing edge-cases.
Thanks @elsaminsut for providing the MMULT.xlsx

- Empty cells in input ranges now return #VALUE! instead of being coerced to 0
- Boolean cells in input ranges now return #VALUE! instead of being coerced to 0/1
- Overflow (inf/NaN result) now returns #NUM! instead of propagating infinity
- Omitted arguments (e.g. =MMULT(,)) now return #VALUE! caught early via EmptyArgKind
- Add unit tests for edge cases
- Add MMULT.xlsx test workbook with Excel-verified expected values


elsaminsut (approved):
Thanks for the fixes! LGTM

PR #1053, opened 2026-05-26
update: in CF/Icon Sets, change default select to first item in the list

dg-ac:
There was a tiny bug by which we were selecting the 5th preset inside Icon Sets by default.

image

It makes more sense to select the first item on the list:

image


webapp/IronCalc/src/components/RightDrawer/ConditionalFormatting/IconSetsRule.tsx
```diff
@@ -462,7 +462,7 @@ const IconSetsRule = ({
   const [colorOpenIndex, setColorOpenIndex] = useState<number | null>(null);
 
   const selectedPreset =
-    ALL_PRESETS.find((p) => p.id === selected) ?? DIRECTIONAL[0];
+    ALL_PRESETS.find((p) => p.id === selected) ?? INDICATORS[0];
```
nhatcher, line 465: Auch!

nhatcher (approved):

PR #1063, opened 2026-05-27
fix: modal dialogs rendered inside the worksheet now work correctly

dg-ac:
This PR fixes 2 issues that affected modals opened from within the cell editor:

- Enter key did not close the dialog because `focusWorkbook()` was stealing focus from the autoFocused button. Fixed by adding `.ic-modal-dialog` to the focus-stealing guard.

- Clicking the backdrop also triggered cell selection. Fixed by stopping pointer event  propagation on the backdrop in all 3 modal components: Alert, Confirm, and Prompt.

### DIFF
--- a/webapp/IronCalc/src/components/Modal/Alert.tsx
+++ b/webapp/IronCalc/src/components/Modal/Alert.tsx
@@ -49,6 +49,7 @@ export function Alert({
     <div
       className="ic-modal-dialog-backdrop"
       onClick={closeModal}
+      // HACK: prevents the workbook from stealing focus while the modal is open
       onPointerDown={(event) => event.stopPropagation()}
       role="none"
     >
--- a/webapp/IronCalc/src/components/Modal/Alert.tsx
+++ b/webapp/IronCalc/src/components/Modal/Alert.tsx
@@ -46,7 +46,12 @@ export function Alert({
   }
 
   return createPortal(
-    <div className="ic-modal-dialog-backdrop" onClick={closeModal} role="none">
+    <div
+      className="ic-modal-dialog-backdrop"
+      onClick={closeModal}
+      onPointerDown={(event) => event.stopPropagation()}
+      role="none"
+    >
       <div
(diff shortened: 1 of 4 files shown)


webapp/IronCalc/src/components/Workbook/Workbook.tsx
```diff
@@ -74,7 +74,7 @@ const Workbook = (props: { model: Model; workbookState: WorkbookState }) => {
     // Basically prevents the workbook from stealing focus when the user is interacting with the color picker, border picker or advanced color picker
     if (
       active?.closest(
-        ".ic-color-picker, .ic-advanced-color-picker-panel, .ic-border-picker",
+        ".ic-color-picker, .ic-advanced-color-picker-panel, .ic-border-picker, .ic-modal-dialog",
       )
```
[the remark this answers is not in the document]
dg-ac: Fair, I've updated it.

webapp/IronCalc/src/components/Modal/Alert.tsx
```diff
@@ -46,7 +46,12 @@ export function Alert({
   }
 
   return createPortal(
-    <div className="ic-modal-dialog-backdrop" onClick={closeModal} role="none">
+    <div
+      className="ic-modal-dialog-backdrop"
+      onClick={closeModal}
+      onPointerDown={(event) => event.stopPropagation()}
```
nhatcher, line 53: Hmmm, this feels like a HACK.
Anything we can do make it a componnet so we don't need to repeat it everywhere?

nhatcher (changes requested):
Just one question

webapp/IronCalc/src/components/Modal/Alert.tsx
dg-ac: Added a comment, let me fix that in another PR today.

nhatcher (approved):

PR #1072, opened 2026-05-28
update: add a common modal component

dg-ac:
This PR (a follow-up from #1063) extracts the repeated portal + backdrop + dialog shell from Alert, Confirm, and Prompt into a shared ModalDialog component.

This way we can centralice the onPointerDown stopPropagation hack that prevents the workbook from stealing focus while a modal is open.

### DIFF
--- a/webapp/IronCalc/src/components/Modal/ModalDialog.tsx
+++ b/webapp/IronCalc/src/components/Modal/ModalDialog.tsx
@@ -22,6 +22,7 @@ export function ModalDialog({
     <div
       className="ic-modal-dialog-backdrop"
       onClick={onClose}
+      // HACK: prevents the workbook from stealing focus while the modal is open
       onPointerDown={(event) => event.stopPropagation()}
       role="none"
     >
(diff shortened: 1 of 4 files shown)


webapp/IronCalc/src/components/Modal/ModalDialog.tsx
```diff
@@ -0,0 +1,43 @@
...
+  onClose,
+  onKeyDown,
+  className,
+  children,
+}: ModalDialogProperties) {
+  return createPortal(
+    <div
+      className="ic-modal-dialog-backdrop"
+      onClick={onClose}
+      onPointerDown={(event) => event.stopPropagation()}
```
nhatcher, line 26: HACK missing

nhatcher (approved):

webapp/IronCalc/src/components/Modal/ModalDialog.tsx
dg-ac: Added 😄

nhatcher (approved):

PR #1077, opened 2026-06-01
UPDATE: Add ACCRINT function (accrued interest)

albertodiazdurana:
## Summary

- Add `ACCRINT(issue, first_interest, settlement, rate, par, frequency, [basis], [calc_method])`
- Computes accrued interest for a security that pays periodic interest
- 34 unit tests + an Excel-authored xlsx round-trip fixture (39 cases)

## Design decisions

- **Follows Mayle SSCM, generalized by the [Microsoft DAX `ACCRINT` documentation](https://learn.microsoft.com/en-us/dax/accrint-function-dax):** `AI = par * (rate / frequency) * Σᵢ (Aᵢ / NLᵢ)` over the NC quasi-coupon periods.
- **Quasi-coupon boundaries** are walked backward from `first_interest` at `12/frequency`-month intervals with end-of-month snapping.
- **Odd-long first coupon (Actual bases 1/2/3):** the front-stub accrual is normalized by the regular coupon length, matching Excel.
- **Day-count reuses the existing `fn_yearfrac`** rather than reimplementing it.
- **`calc_method`** selects accrual start: `issue` (TRUE, default) or `first_interest` (FALSE).
- **Returns `#NUM!`** for `rate <= 0`, `par <= 0`, `frequency` not in {1,2,4}, `basis` outside 0..4, or `issue >= settlement`.
- Locale names: en=ACCRINT, it=INT.MATURATO.PER, fr=INTERET.ACC, de=AUFGELZINS, es=INT.ACUM.

## Test coverage

34 unit tests in `base/src/test/test_fn_accrint.rs`:
- All five basis codes; frequencies 1, 2, 4
- `calc_method` TRUE/FALSE; par and frequency defaults
- Odd-long first coupons (front-stub normalization) across frequencies 2 and 4
- Leap-year boundaries (settlement on Feb 29, issue on Feb 29)
- Validation: rate, par, frequency, basis range, arg count, `issue >= settlement`

XLSX round-trip fixture at `xlsx/tests/calc_tests/FINANCIAL/ACCRINT.xlsx` (authored in MS Excel, Office Web): 39 cases (28 numeric + 11 `#NUM!`), all cross-checked against Excel-cached values. `test_calc_FINANCIAL_ACCRINT` passes.

## Context

Follow-up to #865 (ACCRINTM, merged). This PR covers only ACCRINT, with the same minimal, self-contained approach.

One spec case where the DAX documented spec and shipping Excel diverge (frequency-4 end-of-month day-count) is discussed in #1076; this PR ships the documented DAX value, which is exact and reproducible. That single case is excluded from the xlsx fixture; all 39 fixture cases match shipping Excel.


nhatcher (approved):
Thank you!

nhatcher:
Hi @albertodiazdurana, looks great, since I have a bunch of financail functions coming later in the week and I want to avoid conflicts I merged that as is. Since it is thouroughly tested it is pretty safe to do so. If and when issues come we will deal with theme in separate issues

PR #1080, opened 2026-06-03
fix: use stand-alone date forms

### LINKED ISSUE #1042: About locale formatting conventions
Having support for different languages in IronCalc means considering all types of corner cases. One of them is how to abbreviate day and month names in different locales. Most software and browsers follow the Unicode CLDR - same as IronCalc does.

While working on the autofill feature, @rezat-c noticed that in some languages, `day_names_short` have a trailing period but `months_short` don't. Some locales use trailing periods on all abbreviated forms, others on none.

Right now our [locales.json](https://github.com/ironcalc/IronCalc/blob/main/generate_locale/locales.json) is being generated in [main.rs](https://github.com/ironcalc/IronCalc/blob/main/generate_locale/src/main.rs) using the CLDR standard.

We should check that all supported languages use the correct form from the standard. In our case, our main references should be the `ca-gregorian.json` files:
- [English](https://github.com/unicode-cldr/cldr-dates-modern/blob/master/main/en/ca-gregorian.json)
- [French](https://github.com/unicode-cldr/cldr-dates-modern/blob/master/main/fr/ca-gregorian.json)
- [German](https://github.com/unicode-cldr/cldr-dates-modern/blob/master/main/de/ca-gregorian.json)
- [Italian](https://github.com/unicode-cldr/cldr-dates-modern/blob/master/main/it/ca-gregorian.json)
- [Spanish](https://github.com/unicode-cldr/cldr-dates-full/blob/master/main/es/ca-gregorian.json)

And what's the correct form to use? In [this article](http://cldr.unicode.org/translation/date-time/date-time-symbols#standalone-vs-format-styles), the Unicode standard is clear - the **standalone version** is the one we'd use to write just the month in shortened form without a day.

In [dates.rs](https://github.com/ironcalc/IronCalc/blob/main/generate_locale/src/dates.rs) (line 57) we explicitly say that we are using the **format** mode. This is the wrong form for our use case, since we need **standalone** rather than **format** forms. So my suggestion is to tweak the script just enough to grab the **standalone** day and month versions instead.

rezat-c:
fix #1042


elsaminsut (approved):
Sorry for the delay. It took us a while to check that every change came from the generate_locales script but everything looks great!

LGTM!

PR #1100, opened 2026-06-12
update: Color picker cleanup and improvements

dg-ac:
This PR includes a round of improvements to the `ColorPicker` component, listed below:

image

### Storybook
- Added a new story for `ColorPicker`: theme can be switched from the Storybook settings.

### Css cleanup
- Structure cleanup: sections are now consistent in terms of margin and padding, all styling is placed in classes now. We don't rely on individual margins or paddings as before
- Removed some dead css

### Consistency improvements
- The toolbar font/fill color buttons now show the active (pressed) state while their picker is open
- Removed the full-screen backdrop: the picker closes via a document-level `pointerdown` listener, so clicking another toolbar menu closes the picker and opens the target menu in a single click
- Positioning now lives in a dedicated `useAnchorPosition` hook (same as in the `Menu` component)

### Fixes
- "Add new color" button now uses the shared `Tooltip` component
- Keyboard navigation works smoothly and no longer skips the standard colors row
- Removed the 14-color limit on recent colors
- Added language keys to the text in the 'Default' button

### DIFF
--- a/webapp/IronCalc/src/components/ColorPicker/ColorPicker.stories.tsx
+++ b/webapp/IronCalc/src/components/ColorPicker/ColorPicker.stories.tsx
@@ -2,6 +2,7 @@ import type { Color, IronCalcTheme } from "@ironcalc/wasm";
 import { getThemeList } from "@ironcalc/wasm";
 import type { Meta, StoryObj } from "@storybook/react";
 import { useEffect, useRef, useState } from "react";
+import { useTranslation } from "react-i18next";
 import { init } from "../../index";
 import ColorPicker from "./ColorPicker";
 
@@ -29,6 +30,7 @@ function ColorPickerStory({ theme: themeName }: ColorPickerStoryProps) {
   // must be initialized before rendering the picker.
   const [themes, setThemes] = useState<IronCalcTheme[] | null>(null);
   const anchorRef = useRef<HTMLDivElement>(null);
+  const { t } = useTranslation();
 
   useEffect(() => {
     let cancelled = false;
@@ -64,7 +66,7 @@ function ColorPickerStory({ theme: themeName }: ColorPickerStoryProps) {
       <ColorPicker
         color={color}
         defaultColor="#000000"
-        title="Default"
+        title={t("color_picker.default")}
         onChange={setColor}
         onClose={() => {}}
         anchorEl={anchorRef}
--- a/webapp/IronCalc/src/components/ColorPicker/ColorPicker.stories.tsx
+++ b/webapp/IronCalc/src/components/ColorPicker/ColorPicker.stories.tsx
@@ -1,0 +1,100 @@
+import type { Color, IronCalcTheme } from "@ironcalc/wasm";
+import { getThemeList } from "@ironcalc/wasm";
+import type { Meta, StoryObj } from "@storybook/react";
+import { useEffect, useRef, useState } from "react";
+import { init } from "../../index";
+import ColorPicker from "./ColorPicker";
+
+// Names of the builtin themes returned by getThemeList()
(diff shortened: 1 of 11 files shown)


webapp/IronCalc/src/components/Tooltip/Tooltip.tsx
```diff
@@ -58,7 +60,7 @@ export function Tooltip({ title, children }: TooltipProperties) {
         >
           {title}
         </div>,
-        document.body,
+        container ?? document.body,
```
nhatcher: Why are you doing this? This will break with @jvdp changes, correct?

webapp/IronCalc/src/components/ColorPicker/useAnchorPosition.ts
```diff
@@ -0,0 +1,104 @@
+import {
+  type CSSProperties,
```
nhatcher, line 2: Isn't this file repeated accross many popups and menus?

nhatcher (approved):
LGTM!

webapp/IronCalc/src/components/ColorPicker/useAnchorPosition.ts
dg-ac: Yeah, maybe it's time to move it and make it reusable?

webapp/IronCalc/src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -261,31 +190,20 @@ const ColorPicker = ({
...
-          // Otherwise the sheet would grab the keyboard focus
-          event.stopPropagation();
-        }}
-      >
+    <div
+      ref={panelRef}
+      className="ic-color-picker"
+      style={position}
+      role="dialog"
+      aria-modal="true"
```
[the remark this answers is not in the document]
dg-ac: True, true. Removed.

webapp/IronCalc/src/components/ColorPicker/ColorPicker.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+    let cancelled = false;
+    init().then(() => {
+      if (!cancelled) {
+        setThemes(getThemeList());
+      }
+    });
+    return () => {
+      cancelled = true;
+    };
+  }, []);
```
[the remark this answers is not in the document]
dg-ac: Good catch, but I'd rather keep it consistent with `Workbook.stories.tsx` (the only other story calling `init()`), which doesn't handle the rejection either.

We could do this but I think it should be a separate and wider project.

webapp/IronCalc/src/components/ColorPicker/color-picker.css
```diff
@@ -3,7 +3,6 @@
   z-index: 1300;
   max-width: 220px;
   min-width: 180px;
```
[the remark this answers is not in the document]
dg-ac: Actually, I'm changing my mind. I'll cap it to 29 colors (3 lines).

webapp/IronCalc/src/components/Tooltip/Tooltip.tsx
dg-ac: You're right on that. Also that branch has already some conflicts and this is a very minimal thing (we can live without it for sure) so I've reverted the change.

We can wait until that one is merged, and then maybe polish here and there.

webapp/IronCalc/src/components/ColorPicker/useAnchorPosition.ts
nhatcher: Next PR.

PR #1101, opened 2026-06-12
update: add hyperlink colors to EditTheme 

dg-ac:
This PR adds the 2 colors we were missing in theme customization. For the total of 10 available colors, the last 2 are related to hyperlink and visited hyperlink.

1. Color and hex code inputs added both both cases

image

2. Translation keys added as well for all locales
 
3. There's also a small change regarding wording: 'Edit theme' is not 'Customize theme'. This affects both the drawer header and button tooltip.

image image


nhatcher (approved):
Lgtm!

PR #1102, opened 2026-06-13
fix: accept booleans as arguments in EXPON.DIST

elsaminsut:
Small fix to handle booleans as arguments for the EXPON.DIST function to match Excel's behavior.


nhatcher (approved):
Thank you!

PR #1104, opened 2026-06-13
fix: accept booleans as arguments in GAMMALN and GAMMALN.PRECISE

elsaminsut:
Small fix to handle booleans as arguments for the GAMMALN and GAMMALN.PRECISE functions to match Excel's behavior.


nhatcher (approved):
Lgtm!

PR #1109, opened 2026-06-13
fix: accept booleans as arguments in GAUSS function

elsaminsut:
Small fix to handle booleans as arguments for the GAUSS function to match Excel's behavior.


nhatcher (approved):
Lgtm!

PR #1110, opened 2026-06-13
update: add xlsx test cases and extend unit tests for GAUSS

elsaminsut:
Improve test coverage for the GAUSS function with xlsx tests and adding a small unit test to test the fix in #1109.


nhatcher (approved):
Thank you @elsaminsut !

PR #1113, opened 2026-06-15
wasm: load xlsx in the browser

sylvinus:
Adds Model.loadFromXlsx(bytes, name, locale, timezone, language_id) to the wasm binding so xlsx files can be parsed entirely client-side, plus getDimension and getMergeCells helpers. The ironcalc (xlsx) crate is added as a dependency of bindings/wasm, and xlsx's zip dependency drops default features (keeping only deflate) — the default pulls in bzip2, a C library that can't target wasm32. xlsx archives only use DEFLATE.


nhatcher:
Hi @sylvinus , we can't do this at the moment. The reason is that it significantly increases the build.

also there is still a previous attempt at this:

https://github.com/ironcalc/IronCalc/pull/381

Doing two different builds is something we can definitely consider. Let me think on how to combine those two as soon as possible.

sylvinus:
Ok, makes sense! I'll follow up on #381

PR #1114, opened 2026-06-15
xlsx: make xfId optional in cellXfs

sylvinus:
xfId on a cellXfs <xf> is optional per the spec (it references cellStyleXfs), and many real Excel/LibreOffice files omit it. We currently hard-error with Missing "xfId" XML attribute. Default to 0 when absent.


nhatcher:
Awesome, do you have an example? Would love to add a regression test

sylvinus:
@nhatcher done!

nhatcher:
Interesting. Was the example file produced with Apple numbers? I am happy merging this, but I am unsure how far in the rabithole we should go.

There are several things that Apple Numbers does differently and it would be a huge undertake to create full compatibility with the standard.

nhatcher (approved):
Thanks!

sylvinus:
Thanks for the merge! Not exactly sure where it came from but I think we should aim at broad compatibility... users won't understand that a file that works fine in their app doesn't load in IronCalc !

PR #1115, opened 2026-06-15
formatter: accept locale suffix in currency tokens

### LINKED ISSUE #934: Dollar formatted cells aren't correctly imported to IronCalc
Cells with a dollar format are imported with a different dollar format to IronCalc. After importing, these cells have a `[$$-409]#,##0` whereas the IronCalc current currency format is `"$"#,##0.00`.

### Example
[dollar-format.xlsx](https://github.com/user-attachments/files/27531518/dollar-format.xlsx)
https://testing.ironcalc.com/?model=d9oEd-gE0N8-0w9Ik

### Reproduction
1. Import the example workbook to IronCalc
2. All cells with dollar format (`E3:E13`) show `#VALUE!` instead of their actual values
3. Applying a currency format with the `Format as currency` button correctly shows the cell values

sylvinus:
The number-format lexer only parsed [$<symbol>] and errored on the common [$<symbol>-<locale>] form (e.g. [$€-40C]), which Excel/LibreOffice emit for localized currency. Such cells rendered as #VALUE!. Consume and ignore the optional -<locale> suffix.

### DIFF
--- a/base/src/formatter/lexer.rs
+++ b/base/src/formatter/lexer.rs
@@ -344,11 +344,22 @@ impl Lexer {
                                 Token::ILLEGAL
                             }
                         } else if c == '$' {
-                            // currency
+                            // currency: [$<symbol>] or [$<symbol>-<locale>]
+                            // (e.g. [$€-40C]). The optional "-<locale>" suffix
+                            // is consumed and ignored. Locale-only tokens like
+                            // [$-40C] keep the previous unsupported behaviour.
                             self.read_next_char();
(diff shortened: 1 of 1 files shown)


nhatcher:
Thanks for bringing this up. This is an important one that actually breaks many imports. We need to add tests though.
FYI @dg-ac , @elsaminsut

nhatcher:
This would fix: https://github.com/ironcalc/IronCalc/issues/934

nhatcher:
Thanks for kicking this off!. Fixed in #1186

PR #1116, opened 2026-06-15
webapp: read-only mode via canEdit

sylvinus:
Adds an optional canEdit prop (default true) to <IronCalc>, threaded down to Workbook. When false, the toolbar and formula-bar editor are hidden, paste/cut and editing/Delete/F2 key handlers are gated, and sheet add/rename/delete affordances are removed — while keeping the canvas, keyboard navigation, and sheet tabs for browsing. Useful for embedding IronCalc as a viewer. Reuses the existing canEdit plumbing already present on Toolbar/FormulaBar/Worksheet.

### DIFF
--- a/webapp/IronCalc/src/IronCalc.tsx
+++ b/webapp/IronCalc/src/IronCalc.tsx
@@ -11,6 +11,8 @@ import { type PartialIronCalcThemeVariables, setThemeVariables } from "./theme";
 interface IronCalcProperties {
   model: Model;
   themeVariables?: PartialIronCalcThemeVariables;
+  /** When false, renders without the toolbar/formula bar and blocks all edits. */
+  canEdit?: boolean;
 }
 
 export interface IronCalcHandle {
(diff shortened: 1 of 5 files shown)


dg-ac:
Hi @sylvinus, thank you for working on this! I've tested it and it works well, although I have a few comments:

1. **Toolbar:** Just hiding it may not be enough for what you want to achieve. Note that even when hidden, there's a 40px empty space above the formula bar. That's because it uses `top: var(--toolbar-height);`, so you may want to toggle this variable when editing is disabled.

2. **Storybook:** What do you think about adding a simple [control/toggle](https://storybook.js.org/docs/essentials/controls?ref=ui) to the Workbook story? It would make testing this a lot easier.

3. In `useKeyboardNavigation.ts`, it might be worth grouping all the shortcuts that should be gated behind `canEdit` together. We'll likely be adding more shortcuts at some point (like insert row/col, autosum, and so) so having them organized could help.

4. To fix the failing tests, you can run `npm run check-write`

That'd be all, thanks again for the contribution!

(@nhatcher feel free to add anything if you want to)

nhatcher:
Hi @dg-ac, can you take over this branch?

dg-ac:
Closing this PR as we have #1218

PR #1122, opened 2026-06-16
FIX: Remove unused lists in docs.ironcalc.com

nhatcher:
As we agreed I removed the redundant lists on each page.

Also removed the link to the cube functions


dg-ac (approved):
LGTM!

PR #1131, opened 2026-06-18
fix: long formulas overflow in Named Ranges drawer and menu

### LINKED ISSUE #1127: Can't edit names with long formulas in defined names
See, for instance:

Image

### LINKED ISSUE #1128: Defined names with long names break in the name manager dropdown
See, for instance:

Image

Related #1127

dg-ac:
Fixes #1127 and #1128

### DIFF
--- a/webapp/IronCalc/src/components/FormulaBar/formula-bar-menu.css
+++ b/webapp/IronCalc/src/components/FormulaBar/formula-bar-menu.css
@@ -23,10 +23,8 @@
   top: calc(100% + 4px);
   left: -4px;
   z-index: 1000;
-
-  max-width: 460px;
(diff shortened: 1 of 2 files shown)


nhatcher (changes requested):
Why are you using webkit specific rules?

Does this work on FF snd other browsers?

Unless we absolutely must we shouldn't do this. And if we do we should document it thoroughly

dg-ac:
@nhatcher After our discussion, I've decided to remove the line-clamp and webkit, and just use the normal height. For formulas there's no limitation in terms of lines, that way you can read the whole thing from the list.

nhatcher (approved):

PR #1138, opened 2026-06-18
update: add a separator between canvas and row/col headers

### LINKED ISSUE #1134: [Widget]: Missing line dividing the column headers and row headers from the sheet
This is specially notizable without a grid. Note that we could have the grid in a different color in the future, so we need to account for that

Image

dg-ac:
Fixes #1134, it'd look like this now:

image

@nhatcher please let me know if you like this approach.


dg-ac:
Closed as we will follow a different approach.

PR #1178, opened 2026-06-24
fix: double-click autofill stops incorrectly when target range has values

### LINKED ISSUE #1167: Double-click autofill doesn't work when range already contains values
## Description
When double-clicking the fill handle to autofill a series, the operation fails if the target range already contains values. Manually dragging the fill handle works correctly in the same scenario.

## Steps to Reproduce
1. Enter numbers 1 to 10 in cells A1:A10
2. Replace cells A2 and A3 with `1`
3. Select the range A1:A3
4. Double-click the fill handle
5. The range A1:A10 should be autofilled with `1` but double-clicking does nothing

https://github.com/user-attachments/assets/29bebf68-ec1b-4126-a356-c2d12074101e

rezat-c:
First target row non-empty => overwrite mode (stop at first empty row)
First target row empty => skip mode (stop at first non-empty row)
fix #1167

### DIFF
--- a/webapp/IronCalc/src/components/WorksheetCanvas/outlineHandle.ts
+++ b/webapp/IronCalc/src/components/WorksheetCanvas/outlineHandle.ts
@@ -188,23 +188,24 @@ export function attachOutlineHandle(
     let lastUsedRow = rowEnd;
     let testColumn = columnStart - 1;
 
-    // The "test column" is the column to the left of the selected range,
-    // or the column to the right if the left one is unavailable or empty.
-    if (
-      testColumn < 1 ||
-      worksheet.model.getFormattedCellValue(sheet, rowStart, testColumn) === ""
-    ) {
-      testColumn = columnEnd + 1;
-      if (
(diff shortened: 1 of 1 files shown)


elsaminsut (changes requested):
Thanks for the quick fix! It generally works great but I found that it doesn't autofill when the range is isolated with no adjacent columns, like:

Image

Could you take a look into that?

nhatcher:
Also does it have regression tests?

elsaminsut (changes requested):
Thanks for the changes! It seems to work great now.

As @nhatcher said, we just need some tests to avoid regressions and we're good to go.

nhatcher:
@elsaminsut , do you want to take over this?

elsaminsut:
Sure, I can add some tests here

nhatcher (approved):
sorry for the tremendous delay, but looks good. Its a bit tough to make tests so let's go as such. Thank you both!

elsaminsut (approved):
LGTM!

PR #1180, opened 2026-06-24
fix: use locale symbols for decimal/group separators

### LINKED ISSUE #1179: hardcoded group separator breaks parsing
now we hardcodes group separators in formatter to only ',' or '.'
in locales like French, the group separator is U+202F (NARROW NO-BREAK SPACE), so "1.000" should not be parsed as 1000

1. choose fr locale
2. Type "1.000" in A1

now it parse as "1 000", but should be string "1.000"

rezat-c:
removed hardcoded decimal/group separator logic in parse_formatted_number
in some locales
fix #1179

### DIFF
--- a/base/src/formatter/format.rs
+++ b/base/src/formatter/format.rs
@@ -821,11 +821,9 @@ pub(crate) fn parse_formatted_number(
     let value = original.trim();
     let scientific_format = "0.00E+00";
 
-    let (decimal_separator, group_separator) = if locale.numbers.symbols.decimal == "," {
-        (b',', b'.')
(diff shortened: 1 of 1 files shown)


nhatcher:
Hi @rezat-c, this is very important but it needs:

* Regression tests
* Answer Copilot's comments.

nhatcher (changes requested):
As mentioned, could you add some regression tests?

nhatcher:
I'll take over this PR.

nhatcher:
See: #1240

PR #1191, opened 2026-06-25
update: templates cleanup

dg-ac:
This PR:

1. Removes some of the new templates from the welcome and templates dialogs, as some of them are not fully ready yet
2. It also uploads the completed ones to the right folder


nhatcher:
Superseeded by #1214

PR #1203, opened 2026-06-30
rootContainer prop

jvdp:
Replacement for #871. It solves basically the same issues but in a different way: there is now a `rootContainer` concept, tied to the repurposed `.ic-root` class, which is an element within which both the widget and the modals are to be mounted. It defaults to `document.body` and can be specified as a property on the widget. This provides more flexibility for embedding IronCalc, eg. in a shadow DOM.

Notes:
- `.ic-root` no longer points to the widget; a new `.ic-widget` class is introduced for that purpose instead.
- The default CSS theme variables no longer select on`:root` but on `.ic-root` instead.
- `createAnchoredPortal` is a new helper function that finds the closest `.ic-root` parent node from an anchor element. 
- Portals and the worksheet canvas don't render while the root element is not found, which may be the case during the initial render.

### DIFF
--- a/webapp/IronCalc/src/IronCalc.tsx
+++ b/webapp/IronCalc/src/IronCalc.tsx
@@ -1,16 +1,21 @@
 import type { Model } from "@ironcalc/wasm";
-import { forwardRef, useEffect, useImperativeHandle, useRef } from "react";
+import { forwardRef, useEffect, useImperativeHandle } from "react";
 import { I18nextProvider } from "react-i18next";
 import Workbook from "./components/Workbook/Workbook.tsx";
 import { WorkbookState } from "./components/workbookState.ts";
 import i18n from "./i18n";
 import "./theme/theme.css";
 import "./index.css";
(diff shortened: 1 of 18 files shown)


dg-ac (approved):
LGTM!

There's an open comment from Copilot, feel free to reply to it. Otherwise, it's good to go.

And of course, thank you for your help!

jvdp:
> There's an open comment from Copilot, feel free to reply to it. Otherwise, it's good to go.

The point about unmounting and properly cleaning up after itself is valid. New changes:
- The `document.body` fallback is now treated identically to a specified root container (class is added/removed.)
- Theme variables are now also cleaned up (since they're outside of the widget tree now)

It touches on something not previously designed yet: mounting multiple workbook widgets with the same root container. Since this gets hairy around cleanup (and theme variable clashes) I think it should be unsupported; this will now log a warning.

> And of course, thank you for your help!

No problem, happy to help!

jvdp:
🥳

PR #1216, opened 2026-07-04
FIX: When editing a spill cell we should show no content

nhatcher:
This shows an empty string when editing a "spill" cell where previously we would show the value.
It is still not clear to me what we should do.
For dynamic arrays:

What should we show in the anchor and spill cells both in the cell editor and the formula bar, both when editing and when only displaying content

this is what Excel onthe web does:

image

image

Google sheets does this:

image

(formula is not displayed)

That is close to what we do in IronCalc now.

This is what zoho does:

image

(yes their implementation of SEQUENCE is wrong)

Waiting on your input @dg-ac

### DIFF
--- a/base/src/test/user_model/mod.rs
+++ b/base/src/test/user_model/mod.rs
@@ -13,6 +13,7 @@ mod test_delete_row_column_formatting;
 mod test_diff_queue;
 mod test_duplicate_sheet;
 mod test_dynamic_arrays;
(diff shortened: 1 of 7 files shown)


nhatcher:
After a discussing we decided to follow Excel on the desktop:

* Show greyed out for children
* When editing a spill cell you see nothing

PR #1220, opened 2026-07-07
fix: formula helper positioning

dg-ac:
This PR fixes an issue by which the formula helper wasn't properly positioned when editing cells near bottom and right edges of the viewport. 

image

We now use the same logic as in `Menu/useMenuPosition.ts`, by which it will show above or at the left side of the edited cell.

image

Also, the detail card now remembers its collapsed state for the session instead of always reopening expanded.


webapp/IronCalc/src/components/utils/useMenuPosition.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+  if (top + menuHeight > viewportHeight - margin) {
+    top = viewportHeight - menuHeight - margin;
+  }
+  if (top < margin) {
+    top = margin;
+  }
+
+  return { top, left };
+}
```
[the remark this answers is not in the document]
dg-ac: This is known and discussed, I'll tackle this in the future.

webapp/IronCalc/src/components/FormulaHelper/FormulaHelper.tsx
```diff
@@ -137,14 +137,22 @@ function FunctionList({
...
 }: {
   name: string;
   activeIndex: number;
 }) {
-  const [collapsed, setCollapsed] = useState(false);
+  const [collapsed, setCollapsed] = useState(lastCollapsed);
+  const toggleCollapsed = () => {
+    lastCollapsed = !collapsed;
+    setCollapsed(!collapsed);
+  };
```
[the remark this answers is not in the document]
dg-ac: Fixed!

webapp/IronCalc/src/components/Editor/Editor.tsx
```diff
@@ -427,13 +417,13 @@ const Editor = (options: EditorOptions) => {
...
+      {showHelper
         ? createAnchoredPortal(
             <div
+              ref={helperRef}
               style={{
                 position: "fixed",
-                left: helperPosition.left,
-                top: helperPosition.top,
+                ...helperPosition,
                 zIndex: 1000,
```
nhatcher: Can you remove that?

webapp/IronCalc/src/components/FormulaHelper/FormulaHelper.tsx
```diff
@@ -137,14 +137,24 @@ function FunctionList({
   );
 }
 
+// Details card unmounts whenever the helper closes so
+// user's choice (collapsed/uncollapsed) is remembered.
+let lastCollapsed = false;
```
nhatcher, line 142: Why do you need a global variable? Anyway you can do without this? Is this ok?

nhatcher (changes requested):
Two minor comments

webapp/IronCalc/src/components/FormulaHelper/FormulaHelper.tsx
dg-ac: Well the formula helper unmounts every time it closes, so it'd forget everything unless we place the var outside.

webapp/IronCalc/src/components/Editor/Editor.tsx
dg-ac: Removed, actually we didn't even need it.

nhatcher (approved):

PR #1223, opened 2026-07-08
Correct the built-in numFmtId table (ECMA-376)

scosman:
## Problem

`DEFAULT_NUM_FMTS` in `base/src/number_format.rs` is indexed by `numFmtId`, but
from id ~18 onward the entries are misaligned and incorrect.

- the id 37–40 parenthesis formats sit at indices 23–26,
- ids 45–49 sit at indices 28–32,
- indices 38–44 hold placeholder codes like `"t0.00"`.

So `get_num_fmt(id)` returns the wrong code for most standard built-ins.
Concretely, id 39 returns `"t0.00"`, which the formatter can't parse — a valid
number in a cell using built-in format 39 renders as `#VALUE!` instead of
`#,##0.00`.

### Repro

Open a workbook whose cells use built-in accounting/number formats (ids 37–44)
— e.g. Excel's Accounting/Currency/comma styles applied without hand-editing the
format code. Affected cells show `#VALUE!` though the underlying value is fine.

## Fix

Rebuild the table as the canonical ECMA-376 (18.8.30) built-ins indexed by id:
restore the reserved 23–36 gap (→ `"general"`) and place ids 37–49 correctly. A
workbook-defined `numFmt` still takes precedence — `get_num_fmt` checks the
user-defined map first.

## Tests

Assert the id→code mapping, user-defined precedence, the id 39 regression
specifically, and that every corrected code actually formats a value without
error.

## Note

id 47 (`"mmss.0"`) is the correct ECMA-376 code, but the formatter does not yet
parse that pattern — a separate formatter gap, documented in the test and
excluded from the "formats without error" check. Out of scope here.

Screenshot 2026-07-07 at 10 30 02 PM

### DIFF
--- a/base/src/number_format.rs
+++ b/base/src/number_format.rs
@@ -4,52 +4,62 @@ use crate::{
     types::NumFmt,
 };
 
+// The standard built-in number formats defined by ECMA-376 (§18.8.30), indexed by `numFmtId`
+// so that `DEFAULT_NUM_FMTS[id]` is the format code for that built-in id. Ids 23–36 are
+// reserved/undefined by the spec and resolve to "general". A workbook-defined `numFmt` always
+// takes precedence over this table (see `get_num_fmt`), so this is only consulted for a
+// referenced-but-undefined built-in id.
 const DEFAULT_NUM_FMTS: &[&str] = &[
-    "general",
-    "0",
-    "0.00",
-    "#,##0",
-    "#,##0.00",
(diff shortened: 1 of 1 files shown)


nhatcher:
Thanks, this has been in my mind for awhile now

This is done in #844 although that PR has linting issues (FYI @syabro , @CoalUnicorn)

nhatcher (approved):

PR #1255, opened 2026-07-11
i18n: add Korean (ko) translation resources

moduvoice:
## Motivation

This adds Korean (`ko`) translation support — to help Korean-speaking users use this
open-source project more comfortably in their own language.

한국 사람들의 오픈소스 이용에 도움이 되게 하기 위해서 한글화 작업을 하였습니다.

## Changes

- `webapp/app.ironcalc.com/frontend/src/locale/ko_kr.json` — new Korean translation file, 92 keys (full parity with `en_us.json`)
- `webapp/app.ironcalc.com/frontend/src/i18n.ts` — registers `ko-KR` as an i18next resource
- `webapp/IronCalc/src/locale/ko_kr.json` — new Korean translation file, 333 keys (full parity with `en_us.json`)
- `webapp/IronCalc/src/i18n.ts` — registers `ko-KR` as an i18next resource

**Note on the language switcher:** the `ko-KR` i18next resources are fully translated and
registered, but intentionally **not yet** added to the `LANGUAGES` arrays in
`FileMenu.tsx` / `MobileMenu.tsx` (the UI language switcher). The reason is that the
Rust/wasm calculation engine only ships compiled language support for `en`/`es`/`fr`/`de`/`it`
(generated from `generate_language/src/languages.json` into `base/src/language/language.bin`).
`model.setLanguage()` looks the code up via `get_language()` in `base/src/language/mod.rs`,
which returns `Err("Language is not supported: 'ko'")` for any code it doesn't recognize; the
wasm binding turns that `Err` into a thrown `JsError`, and neither `IronCalc.tsx` nor
`Workbook.tsx` wrap the `model.setLanguage()` call in a try/catch. So enabling `ko-KR` in the
switcher today would let a user pick it, desync the UI language from the engine's internal
locale state, and throw an uncaught error on `model.setLanguage("ko")`.

This PR therefore only adds the translation resources. Once Korean is added to the engine's
compiled language data (a separate, non-translation task), `ko-KR` can be added to the
`LANGUAGES` arrays to expose it in the switcher — the i18next side is already fully ready.

## Testing

- Verified key parity of both `ko_kr.json` files against their `en_us.json` counterparts programmatically (92/92 and 333/333 keys, no missing/extra/empty values).
- Ran `npm run lint:i18n` in `webapp/IronCalc` — passes: `✓ i18n OK — 333 keys across 6 locales`.
- Confirmed currency codes/symbols (EUR/GBP/USD, €/£/$) and formula syntax (`=`, cell references) are left untranslated as expected.
- Spot-checked translation quality against standard Excel/Google Sheets Korean terminology (e.g. `통합 문서` for workbook, `틀 고정` for freeze panes, `조건부 서식` for conditional formatting, `사용자 지정 숫자 서식` for custom number format).


nhatcher:
I am afraid there is still a lot of work to be done here. @dg-ac can I close this for now?

dg-ac:
@nhatcher Yes, let's revisit in the future as this may be more complicated that what it looks like. I'm also working on a translation guidelines article for the docs that should make these tasks easier.

nhatcher:
Closing for now. @moduvoice I think we need to put all the translation files in a separate build so we don't ship all the languages in one file. I added a "Worth revisiting" label

Very sorry it took me so long to review once we have the rest of the scalfolding in place we will revisit this PR.

PR #1259, opened 2026-07-12
fix(xlsx): accept ECMA-376 true/false forms for xsd:boolean attributes

scosman:
The SpreadsheetML attributes read by the importer's boolean helpers (customHeight, customWidth, customFormat, hidden, wrapText, the <xf> apply*/quotePrefix flags, the <tableStyleInfo> show* flags, ...) are all typed xsd:boolean in the ECMA-376 / ISO 29500 schema. Per W3C XML Schema Part 2 the lexical space of xsd:boolean is exactly {true, false, 1, 0}, so all four literals are spec-valid. get_bool/get_bool_false only recognized the Excel form "1"/"0" and silently mishandled the equally- valid "true"/"false" form that LibreOffice emits: customHeight="true" lost its flag (custom row heights fell back to the default) and wrapText="true" imported as non-wrapping. The pre-fix "1"/"0"-only code was the non-compliant one; it rejected valid true/false.

Parse xsd:boolean per the spec: "1"/"true" -> true, "0"/"false" ->
false; absent/empty/unrecognised -> the attribute's schema default. Matching is case-insensitive and whitespace-tolerant as a deliberate lenient read of real-world files; tokens outside the xsd:boolean lexical space ("yes", "on", garbage) are NOT over-accepted -- they fall back to the default rather than being read as true. Excel "1"/"0" files are unaffected. Route the row/column importer's inline customHeight/customWidth/customFormat/hidden checks through the shared helper so they are corrected the same way as wrapText.

### DIFF
--- a/xlsx/src/import/util.rs
+++ b/xlsx/src/import/util.rs
@@ -70,12 +70,107 @@ pub(super) fn get_color(node: Node, _theme: &Theme) -> Result<Color, XlsxError> 
     }
 }
 
+/// Parse an `xsd:boolean` attribute value.
+///
+/// The SpreadsheetML attributes read through these helpers (`customHeight`, `customWidth`,
+/// `customFormat`, `hidden`, `wrapText`, the `<xf>` `apply*` / `quotePrefix` flags, the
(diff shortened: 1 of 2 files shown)


nhatcher (approved):
Very nice! Thank you

PR #1292, opened 2026-07-25
Fix TRIM to collapse internal runs of spaces (Excel-compatible)

scosman:
TRIM previously returned s.trim(), which only stripped leading and trailing whitespace and, via str::trim, removed all Unicode whitespace. Excel's TRIM operates on the ASCII space (0x20) only: it trims leading and trailing spaces AND collapses each internal run of two or more spaces to a single space, while leaving tabs, non-breaking spaces, and other whitespace untouched.

Replace the normalization with a split(' ')/filter/join over the ASCII space so internal runs collapse and the 0x20-only scope is preserved. Add tests covering internal-run collapse, the interior-only regression cases, all-spaces input, and the tab/NBSP 0x20-only proofs.

### DIFF
--- a/base/src/test/text_functions/mod.rs
+++ b/base/src/test/text_functions/mod.rs
@@ -103,6 +103,40 @@ fn test_proper_basic() {
     assert_eq!(model._get_text("A3"), "It'S A Test");
 }
 
+// ── TRIM ──────────────────────────────────────────────────────────────────────
+
+#[test]
+fn test_trim_collapse_internal_runs() {
+    let mut model = new_empty_model();
+    // Leading/trailing spaces trimmed AND internal runs collapsed to a single space.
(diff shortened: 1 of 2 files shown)


base/src/test/text_functions/mod.rs
```diff
@@ -103,6 +103,40 @@ fn test_proper_basic() {
     assert_eq!(model._get_text("A3"), "It'S A Test");
 }
 
+// ── TRIM ──────────────────────────────────────────────────────────────────────
+
+#[test]
+fn test_trim_collapse_internal_runs() {
+    let mut model = new_empty_model();
```
nhatcher, line 110: Also good cases for an xlsx @elsaminsut

nhatcher (approved):

PR #1293, opened 2026-07-25
Fix DOLLAR to not parenthesize a value that rounds to zero

scosman:
A negative `number` whose magnitude rounds to zero (e.g. DOLLAR(-0.001, 2) or DOLLAR(-50, -3)) was formatted with the negative parentheses as "($0.00)". Other sheets emit the unsigned "$0.00" in that case, reserving the parenthesized form for values that are still non-zero after rounding.

Detect the rounds-to-zero case from the already-formatted magnitude and emit the non-negative form for it.

### DIFF
--- a/base/src/test/text_functions/mod.rs
+++ b/base/src/test/text_functions/mod.rs
@@ -51,6 +51,38 @@ fn test_dollar_basic() {
     assert_eq!(model._get_text("A4"), "$1,200");
 }
 
+#[test]
+fn test_dollar_negative_rounds_to_zero() {
+    let mut model = new_empty_model();
+    // A negative number whose magnitude rounds to zero must NOT be parenthesized.
+    model._set("A1", "=DOLLAR(-0.001, 2)");
+    // A real negative still uses parentheses.
+    model._set("A2", "=DOLLAR(-1234.567, 2)");
+    // A negative that rounds to zero to the left of the decimal point.
+    model._set("A3", "=DOLLAR(-50, -3)");
+    // Zero itself is unsigned.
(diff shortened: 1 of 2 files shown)


base/src/test/text_functions/mod.rs
```diff
@@ -51,6 +51,38 @@ fn test_dollar_basic() {
     assert_eq!(model._get_text("A4"), "$1,200");
 }
 
+#[test]
+fn test_dollar_negative_rounds_to_zero() {
+    let mut model = new_empty_model();
+    // A negative number whose magnitude rounds to zero must NOT be parenthesized.
+    model._set("A1", "=DOLLAR(-0.001, 2)");
```
nhatcher, line 58: That's not true on my Excel instance @dg-ac , could you try? In my case I get 
```
=DOLLAR(-0.001,2) => ($0.00)
```

In IronCalc you can just create a file DOLLAR.xlsx (done in Excel) with all the test cases you want

nhatcher (changes requested):
Hi @scosman , could you double check what you are saying is correct
I might have missunderstood something

base/src/test/text_functions/mod.rs
dg-ac: Yep I get the same result.

nhatcher: Mmm, I see that GSheets do as you say, We  aim at Excel compatibility not GSheets.

@scosman do you think this PR is worth saving?

scosman: Ah, I didn't test excel directly. No need to save it.

PR #1294, opened 2026-07-25
Fix ADDRESS to prefix ! for an empty sheet_text argument

scosman:
ADDRESS(1,1,1,TRUE,"") is "!$A$1" in Excel: a present sheet_text argument always emits the "!" separator, even when it is an empty string. Only an omitted sheet_text produces no prefix. Previously an empty sheet_text was treated like an omitted one and produced "$A$1".

Quoting of the name is unchanged (an empty name needs no quoting).

### DIFF
--- a/base/src/test/lookup_and_reference/test_fn_address_areas.rs
+++ b/base/src/test/lookup_and_reference/test_fn_address_areas.rs
@@ -68,6 +68,30 @@ fn test_address_large_column() {
     assert_eq!(model._get_text("A1"), "$AA$1");
 }
 
+#[test]
+fn test_address_empty_sheet_text() {
+    let mut model = new_empty_model();
+    // A present-but-empty sheet_text still emits the "!" separator.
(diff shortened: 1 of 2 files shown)


base/src/test/lookup_and_reference/test_fn_address_areas.rs
```diff
@@ -68,6 +68,30 @@ fn test_address_large_column() {
     assert_eq!(model._get_text("A1"), "$AA$1");
 }
 
+#[test]
+fn test_address_empty_sheet_text() {
+    let mut model = new_empty_model();
+    // A present-but-empty sheet_text still emits the "!" separator.
+    model._set("A1", r#"=ADDRESS(1,1,1,TRUE,"")"#);
```
nhatcher, line 75: @elsaminsut , maybe good for an xlsx test case?

nhatcher (approved):

PR #1295, opened 2026-07-25
Add array-constant support to XMATCH lookup_array

scosman:
XMATCH only accepted a range reference for lookup_array; an in-formula array constant such as XMATCH("ban*", {"apple","banana","cherry"}, 2) returned #VALUE!. Materialise lookup_array into a one-dimensional vector of values from either a range or a CalcResult::Array, then run the same linear/binary search over it. A 2-D array constant still yields #VALUE!.


nhatcher:
This is great, but I think we will need an xlsx test case. We can probably add them here:

https://github.com/ironcalc/IronCalc/tree/main/xlsx/tests/calc_tests/LOOKUP_AND_REFERENCE

scosman:
@nhatcher tests added!

elsaminsut (approved):
All tests pass! LGTM

PR #1304, opened 2026-07-30
update: improve testing for LOGNORM functions

elsaminsut:
Add xlsx test and arguments test for LOGNORM functions (LOGNORM.DIST and LOGNORM.INV). 

Includes a small fix to make these functions accept booleans as arguments.


nhatcher (approved):
Great!

PR #1305, opened 2026-07-30
UPDATER: Update python bindings

nhatcher:


yolonir:
thanks for the pr, i'll update the https://github.com/ironcalc/IronCalc/issues/813

PR #1306, opened 2026-07-30
update: improve testing for NORM.DIST functions

elsaminsut:
Add xlsx test and arguments test for NORM.DIST functions (NORM.DIST, NORM.INV, NORM.S.DIST and NORM.S.INV).

Includes a small fix to make these functions accept booleans as arguments.


nhatcher (approved):
Lovely, thanks!

PR #1319, opened 2026-08-02
ci(nodejs): gate publish on complete platform binaries

YashBhalodi:
## Summary
- Remove musl platform packages from `@zolidar/ironcalc-nodejs` optionalDependencies / napi targets (not built in CI).
- Add a hard gate before `npm publish` so missing `.node` artifacts fail the job instead of napi warning-and-skipping.

## Test plan
- [ ] `npm.yml` build matrix stays green
- [ ] Manual dispatch with Publish=true (or tag `v0.8.3-zolidar.1`) publishes all 11 platform packages + root to GitHub Packages


YashBhalodi:
Opened against upstream by mistake — this change belongs on zolidar/IronCalc only.

PR #1326, opened 2026-08-04
Two implicit intersection fixes

Bobstin:
Fixes two related implicit-intersection (`@`) bugs where the operator either
returned `#VALUE!`/`#ERROR!` or silently produced `0` for cases Excel evaluates
to a plain value:

1. **Cross-sheet single-cell reference.** `implicit_intersection` only returned
   a result when the consuming cell shared a row/column with the range, and its
   single-cell (1×1) shortcut was gated on the consuming cell being on the same
   sheet. So `=@Sheet2!D5` — and any OFFSET/INDIRECT result that resolves to one
   cell on another sheet, used in scalar context — failed to dereference and
   evaluated to an error. A 1×1 range is unambiguous and, per Excel, always
   dereferences to that cell's value regardless of where the formula lives.

2. **Scalar flowing through an `@` wrapper.** The importer wraps unknown-shaped
   function arguments in an automatic implicit intersection (e.g. a `CHOOSE`
   arm that is an `IF(...)` returning a value). When that child evaluated to a
   scalar rather than a range, `evaluate_node_with_reference` raised
   `#ERROR!` ("Error with Implicit Intersection"). Downstream `IFERROR` then
   masked it to `0`, so workbooks computed silently-wrong numbers rather than
   erroring visibly. The implicit intersection of a scalar is the scalar itself,
   so the value should flow through.

### DIFF
--- a/base/src/implicit_intersection.rs
+++ b/base/src/implicit_intersection.rs
@@ -5,24 +5,33 @@ use crate::{calc_result::Range, expressions::types::CellReferenceIndex};
 ///  * i_i(B5, A2:A9) -> B5
 ///  * i_i(B5, A7:A9) -> None
 ///  * i_i(B5, A2:D2) -> B2
+///
+/// Excel-parity, cross-sheet:
+///  * a 1x1 range always dereferences, regardless of the consuming cell's
+///    position or sheet (Excel: `=@Sheet2!D5` and OFFSET/INDIRECT results in
+///    scalar context return the single cell's value);
+///  * row/column-aligned intersection works across sheets — the intersection
+///    happens in the range's own coordinate space (Excel: `=Data!D1:D5`
+///    entered on another sheet's C3 returns Data!D3). The resulting reference
+///    lives on the RANGE's sheet.
 pub(crate) fn implicit_intersection(
     cell_reference: &CellReferenceIndex,
     range: &Range,
 ) -> Option<CellReferenceIndex> {
     let left = &range.left;
     let right = &range.right;
-    // A single-cell "range" intersects to itself, on its own sheet. This holds
-    // even when that sheet differs from the calling cell's — e.g. `@Sheet2!B3`,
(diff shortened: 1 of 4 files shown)


base/src/model.rs
```diff
@@ -337,11 +337,13 @@ impl<'a> Model<'a> {
...
-                    cell,
-                    format!("Error with Implicit Intersection in cell {cell:?}"),
-                ),
+                // The implicit intersection of a scalar is the scalar itself. The
+                // importer wraps any unknown-shaped function argument (e.g. a CHOOSE
+                // arm that is an IF(...) returning a value) in an automatic `@`; when
+                // the child evaluates to a value rather than a range, that value must
+                // flow through instead of erroring (previously: #ERROR! "Error with
+                // Implicit Intersection", which IFERROR then masked to 0).
+                other => other,
```
nhatcher, line 346: This is not quite right. It fixes the sympton at the wrong level. But I think the real fix is a bit more complex so I will merge this and tackle the real fix in a diferent PR.

base/src/implicit_intersection.rs
```diff
@@ -5,17 +5,34 @@ use crate::{calc_result::Range, expressions::types::CellReferenceIndex};
...
     }
+    let sheet = left.sheet;
+    // Single-cell range: always dereference (position- and sheet-independent).
+    if left.row == right.row && left.column == right.column {
+        return Some(CellReferenceIndex {
+            sheet,
+            row: left.row,
+            column: left.column,
+        });
+    }
```
nhatcher, line 35: Good catch!

nhatcher (approved):
The first fix in implicit_intersection is correct and at the right level. But the second makes of @ a noop I think it will fail with things like `=SUM(CHOOSE(1, @B1:B3))` for instnace.

Nevertheless let's merge this and I will tackle the real fix right away.

PR #1327, opened 2026-08-05
update: dark mode support

dg-ac:
This PR dark mode to IronCalc! Although there's still a few more things to do, we've started with the next changes:

- The most significant is the cleanup of multiple components that were using hardcoded colors or wrong variables. Most of the files changed are affected by this.
- Added some new color variables like `--palette-surface-elevated` or `--palette-surface-elevated-border`, to handle these that elements are visible on dark mode but not on light
- Minor fixes in the Conditional Formatting thumbnails so they are mode-sensitive
- Added a 'Dark' option to the mode toggle in Storybook
- Added an item in `utils.ts` for the grid corner
  - This brought some adjustments in `worksheetCanvas.ts`
 
Also, see here some screenshots:

image
image
image

---

Pending tasks, will come in future PRs:

- How to handle font and background color in Themes
- Named Cell styles
- Conditional Formatting
- Frozen rows and columns
- Borders
- Cells where content overflows

### DIFF
--- a/webapp/IronCalc/.storybook/preview.tsx
+++ b/webapp/IronCalc/.storybook/preview.tsx
@@ -3,7 +3,11 @@ import { useEffect, useRef, useState } from "react";
 import { I18nextProvider } from "react-i18next";
 import i18n from "../src/i18n";
 import type { PartialIronCalcThemeVariables } from "../src/theme";
-import { defaultThemeVariables, setThemeVariables } from "../src/theme/theme";
+import {
+  darkThemeVariables,
+  defaultThemeVariables,
+  setThemeVariables,
+} from "../src/theme/theme";
 import "../src/theme/theme.css";
 import "../src/index.css";
 
@@ -34,6 +38,7 @@ const crazyThemeVariables: PartialIronCalcThemeVariables = {
 
 const themes = {
   default: defaultThemeVariables,
+  dark: darkThemeVariables,
   crazy: crazyThemeVariables,
 };
 
@@ -88,6 +93,7 @@ const preview: Preview = {
         icon: "paintbrush",
         items: [
           { value: "default", title: "Default" },
+          { value: "dark", title: "Dark" },
           { value: "crazy", title: "Crazy" },
         ],
       },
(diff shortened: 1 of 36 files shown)


webapp/IronCalc/src/components/WorksheetCanvas/worksheetCanvas.ts
```diff
@@ -1191,16 +1191,16 @@ export default class WorksheetCanvas {
       }
       const selected = row >= rowStart && row <= rowEnd;
       context.fillStyle = this.theme.headerBorderColor;
-      context.fillRect(0.5, topLeftCornerY, headerColumnWidth - 1, rowHeight);
+      context.fillRect(0, topLeftCornerY, headerColumnWidth, rowHeight);
```
nhatcher: Why did you need this? Same question for the change below

webapp/IronCalc/.storybook/preview.tsx
```diff
@@ -88,6 +93,7 @@ const preview: Preview = {
         icon: "paintbrush",
         items: [
           { value: "default", title: "Default" },
+          { value: "dark", title: "Dark" },
           { value: "crazy", title: "Crazy" },
```
nhatcher: Can you remove the "Crazy" theme? Was there only for you build the dark mode.

nhatcher (changes requested):
Very nice! I have just a question and a small request

webapp/IronCalc/src/components/WorksheetCanvas/worksheetCanvas.ts
dg-ac: This change is to get rid of these borders on left and right. Couldn't find a better way to avoid this visual glitch. Would you have a suggestion? I'm sure it can be improved.

Image

nhatcher: Can you NOT do this and open an issue once this is merged?

nhatcher (changes requested):

webapp/IronCalc/src/components/WorksheetCanvas/worksheetCanvas.ts
dg-ac: Sure, I've reverted it. I'll open the issue after merge

webapp/IronCalc/.storybook/preview.tsx
dg-ac: Makes sense. Removed

nhatcher (approved):
LGTM!

PR #1341, opened 2026-08-11
XLOOKUP: support computed (array-constant) lookup and return arrays

### LINKED ISSUE #1338: XLOOKUP with a computed lookup_array returns #N/A even when the value is present
## Description

A standard spreadsheet idiom for a two-column join is to concatenate the key columns with a
separator and look up against the result:

```
A1: a   B1: x   C1: first
A2: b   B2: y   C2: second

=XLOOKUP("b|y", A1:A2&"|"&B1:B2, C1:C2)
```

The lookup array is a *computed* array of join keys — `{"a|x"; "b|y"}` — and Excel returns
`second`. In IronCalc 0.8.3, XLOOKUP with any computed `lookup_array` always returns `#N/A`,
even when the value is demonstrably present. A related failure: a computed `return_array`
(with a stored `lookup_array`) returns `#VALUE!`.

Observations that narrow the cause:

1. **The computed join-key array itself is correct** — entered as its own formula,
   `=A1:A2&"|"&B1:B2` spills `a|x`, `b|y` exactly as expected, so array evaluation is not
   the problem.
2. **MATCH and XMATCH over the identical computed array work** and find the value.
3. **VLOOKUP and HLOOKUP over the same computed array also work** — the defect is specific
   to XLOOKUP, not to lookups over computed arrays generally.
4. **Every computed form fails the same way**: text concatenation, `ROW(A1:A2)`, a boolean
   product `(A1:A2="b")*(B1:B2="y")`, and even a typed array literal `{"a";"b"}`.
5. **The `if_not_found` argument is bypassed** — `=XLOOKUP("b|y", A1:A2&"|"&B1:B2, C1:C2,
   "fallback")` returns `#N/A` rather than `"fallback"`, so the error escapes the parameter
   designed to catch it.

Unlike the functions that don't support dynamic arrays yet and refuse loudly with `#N/IMPL!`
(#1153 family), this failure is silent: `#N/A` is a plausible everyday lookup result, so
wrong answers are easy to miss — and formulas that defensively wrap the lookup
(`=IFERROR(XLOOKUP(key, computed, values), fallback)`) always take their fallback branch,
returning a clean, plausible, wrong value. For context on prevalence: in a corpus of
~18,000 real-world production formulas we analysed, the join-key idiom appears ~2,000
times (visible `#N/A`s plus ~290 formulas silently degrading through IFERROR-style
wrappers), and boolean-array lookups hit the same defect.

## Root cause (I think....so more a hypothesis)

`fn_xlookup` only accepts a **stored range** for `lookup_array`. In
`base/src/functions/xlookup.rs` (v0.8.3, `9bed9ec`) the evaluated argument is matched
against `CalcResult::Range` alone, and every other non-error result — including
`CalcResult::Array` from a computed expression — falls through to:

```rust
// xlookup.rs:382-388
error @ CalcResult::Error { .. } => error,
_ => CalcResult::Error {
    error: Error::NA,                       // ← the reported #N/A
    origin: cell,
    message: "Range expected".to_string(),
},
```

So the `#N/A` is not a failed search — it is an **argument-type rejection mislabeled as
`#N/A`**, emitted before `linear_search` ever runs. That also explains the `if_not_found`
bypass: the fallback is only consulted on a `None` search result (`xlookup.rs:302`), and
this error path never reaches the search. The `return_array` arm has the same structure
with `Error::VALUE` (`xlookup.rs:375-379`) — the `L1` facet below.

VLOOKUP works because `fn_vlookup`
(`base/src/functions/lookup_and_reference/mod.rs:576-594`) already handles both variants
via the `LookupTable` abstraction (`CalcResult::Array(array) => LookupTable::Array(array)`),
and MATCH/LOOKUP materialize either form through `as_vector`.

## Test File

Attached in the first comment below: `XLOOKUP_computed_lookup_array_v2.xlsx` — the
reproduction table's formulas in their on-disk form (`_xlfn.`-prefixed, CSE array-marked),
with Excel's cached results stamped. Verified both ways: desktop Excel opens it,
recalculates live, and confirms every value in the Excel column; `Model.fromXlsx` (0.8.3)
on the same file reproduces every failure in the IronCalc column.

## Reproduction

Sheet data: `A1:A2` = `a`, `b` · `B1:B2` = `x`, `y` · `C1:C2` = `first`, `second`

| Cell | Formula | Excel | IronCalc 0.8.3 | Notes |
|---|---|---|---|---|
| E1 | `=A1:A2&"\|"&B1:B2` | spills `a\|x`, `b\|y` | spills `a\|x`, `b\|y` | computed join-key array is correct |
| F1 | `=MATCH("b\|y", A1:A2&"\|"&B1:B2, 0)` | `2` | `2` | MATCH over the same array works |
| F2 | `=XMATCH("b\|y", A1:A2&"\|"&B1:B2)` | `2` | `2` | XMATCH works too |
| G1 | `=XLOOKUP("b\|y", A1:A2&"\|"&B1:B2, C1:C2)` | `second` | `#N/A` | **the bug** |
| H1 | `=XLOOKUP("b\|y", A1:A2&"\|"&B1:B2, C1:C2, "fallback")` | `second` | `#N/A` | `if_not_found` bypassed — this formula can never be `#N/A` in Excel |
| I1 | `=XLOOKUP(2, ROW(A1:A2), C1:C2)` | `second` | `#N/A` | numeric computed array — same failure |
| J1 | `=XLOOKUP(1, (A1:A2="b")*(B1:B2="y"), C1:C2)` | `second` | `#N/A` | boolean product — same failure |
| K1 | `=XLOOKUP("b", {"a";"b"}, C1:C2)` | `second` | `#N/A` | even a typed array literal fails |
| L1 | `=XLOOKUP("b", A1:A2, C1:C2&"")` | `second` | `#VALUE!` | computed *return* array — related failure |
| M1 | `=VLOOKUP("b\|y", A1:A2&"\|"&B1:B2, 1, FALSE)` | `b\|y` | `b\|y` | VLOOKUP over the computed array works |
| N1 | `=HLOOKUP("b\|y", TRANSPOSE(A1:A2&"\|"&B1:B2), 1, FALSE)` | `b\|y` | `b\|y` | HLOOKUP works too |

Reproduced two ways on 0.8.3: via `@ironcalc/nodejs` (`updateCellWithFormula` +
`evaluate()` + `getFormattedCellValue`) and via `Model.fromXlsx` on the attached workbook.

youdie006:
### Summary

Fixes #1338. `XLOOKUP` returns `#N/A` when `lookup_array` is a computed array (and `#VALUE!` when `return_array` is one), because it only handled arguments that evaluate to a range reference.

### Root cause

`fn_xlookup` matched `lookup_array` (args[1]) and `return_array` (args[2]) only against `CalcResult::Range`. A computed array such as `A1:A2&"|"&B1:B2` evaluates to `CalcResult::Array`, so it fell through to the catch-all arm and returned `#N/A` directly — which also bypassed `if_not_found`. A computed `return_array` hit the inner catch-all and returned `#VALUE!`.

```
=XLOOKUP("b|y", A1:A2&"|"&B1:B2, C1:C2)   → #N/A   (should find the match)
=XMATCH("b|y", A1:A2&"|"&B1:B2)           → 2      (sibling already works)
```

### Fix

Materialise both `lookup_array` and `return_array` into a flat `Vec<CalcResult>` (`xlookup_vector`), handling a range reference (via `prepare_array`, with the existing entire-column/row clamping and single-row/column validation) and an in-formula array constant (via `array_node_to_calc_result`) the same way `XMATCH` already does. The search then runs on the lookup vector and the result is taken by index from the return vector.

This unifies the previously range-only path (net fewer lines) and leaves range/range behaviour unchanged.

### Tests

`XLOOKUP` had no tests, so this adds `test_fn_xlookup.rs` covering:
- range/range exact match, not-found (`#N/A`), and `if_not_found`;
- computed `lookup_array` (the reported case), including `if_not_found`;
- computed `return_array` (`{10;20;30}`);
- approximate match (`match_mode=-1`) and binary search (`search_mode=2`).

### Verification (local, in Docker)

- `cargo test -p ironcalc_base` — 2283 + 23 tests pass, 0 failures (the four new tests included; no regressions).
- `cargo clippy -p ironcalc_base --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.

---

Disclosure: this fix was prepared with AI assistance (Claude). I reviewed it, mirrored the established `XMATCH` approach, and verified the red-green tests and full suite myself.

### DIFF
--- a/base/src/functions/xlookup.rs
+++ b/base/src/functions/xlookup.rs
@@ -31,6 +31,45 @@ enum MatchMode {
     WildcardMatch = 2,
 }
 
+/// Which of the two array arguments of XLOOKUP is being evaluated. Excel
+/// reports a different error for each: a `lookup_array` that is not a range
+/// simply finds nothing (`#N/A`), while a `return_array` that is not a range,
+/// or that does not line up with the `lookup_array`, is an invalid argument
+/// (`#VALUE!`).
+#[derive(PartialEq, Clone, Copy)]
+enum VectorArgument {
+    LookupArray,
+    ReturnArray,
+}
+
+impl VectorArgument {
+    fn not_a_vector_error(self, cell: CellReferenceIndex) -> CalcResult {
+        match self {
+            VectorArgument::LookupArray => CalcResult::Error {
+                error: Error::ERROR,
+                origin: cell,
+                message: "Second argument must be a vector".to_string(),
+            },
+            VectorArgument::ReturnArray => CalcResult::Error {
+                error: Error::VALUE,
+                origin: cell,
+                message: "Arrays must be of the same size".to_string(),
+            },
+        }
+    }
+
+    fn not_a_range_error(self, cell: CellReferenceIndex) -> CalcResult {
+        CalcResult::Error {
+            error: match self {
+                VectorArgument::LookupArray => Error::NA,
+                VectorArgument::ReturnArray => Error::VALUE,
+            },
+            origin: cell,
+            message: "Range expected".to_string(),
+        }
+    }
+}
+
 // lookup_value in array, match_mode search_mode
 fn linear_search(
     lookup_value: &CalcResult,
@@ -219,11 +258,11 @@ impl<'a> Model<'a> {
         // Materialise lookup_array (args[1]) and return_array (args[2]) into flat
         // vectors so a range reference or an in-formula array constant (e.g.
         // `A1:A2&"|"&B1:B2`) is handled the same way. See issue #1338.
-        let lookup_array = match self.xlookup_vector(&args[1], cell) {
+        let lookup_array = match self.xlookup_vector(&args[1], cell, VectorArgument::LookupArray) {
             Ok(v) => v,
             Err(e) => return e,
         };
-        let return_array = match self.xlookup_vector(&args[2], cell) {
+        let return_array = match self.xlookup_vector(&args[2], cell, VectorArgument::ReturnArray) {
             Ok(v) => v,
             Err(e) => return e,
         };
@@ -287,6 +326,7 @@ impl<'a> Model<'a> {
         &mut self,
         node: &Node,
         cell: CellReferenceIndex,
+        argument: VectorArgument,
     ) -> Result<Vec<CalcResult>, CalcResult> {
         match self.evaluate_node_in_context(node, cell) {
             CalcResult::Range { left, right } => {
@@ -295,11 +335,7 @@ impl<'a> Model<'a> {
                 } else if left.column == right.column {
                     true
                 } else {
-                    return Err(CalcResult::Error {
-                        error: Error::ERROR,
-                        origin: cell,
-                        message: "Second argument must be a vector".to_string(),
-                    });
+                    return Err(argument.not_a_vector_error(cell));
                 };
                 let mut row2 = right.row;
                 let mut column2 = right.column;
@@ -340,11 +376,7 @@ impl<'a> Model<'a> {
                 let is_row_vec = n_rows == 1;
                 let is_col_vec = n_cols == 1;
                 if n_rows == 0 || n_cols == 0 || (!is_row_vec && !is_col_vec) {
-                    return Err(CalcResult::Error {
-                        error: Error::ERROR,
-                        origin: cell,
-                        message: "Second argument must be a vector".to_string(),
-                    });
+                    return Err(argument.not_a_vector_error(cell));
                 }
                 Ok(rows
                     .iter()
@@ -353,11 +385,7 @@ impl<'a> Model<'a> {
                     .collect())
             }
             error @ CalcResult::Error { .. } => Err(error),
-            _ => Err(CalcResult::Error {
-                error: Error::NA,
-                origin: cell,
-                message: "Range expected".to_string(),
-            }),
+            _ => Err(argument.not_a_range_error(cell)),
         }
     }
 }
--- a/base/src/functions/xlookup.rs
+++ b/base/src/functions/xlookup.rs
@@ -1,7 +1,10 @@
 use crate::constants::{LAST_COLUMN, LAST_ROW};
 use crate::expressions::types::CellReferenceIndex;
 use crate::{
-    calc_result::CalcResult, expressions::parser::Node, expressions::token::Error, model::Model,
+    calc_result::CalcResult,
+    expressions::parser::{ArrayNode, Node},
+    expressions::token::Error,
+    model::Model,
 };
 
 use super::{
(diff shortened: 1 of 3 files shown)


youdie006:
Thanks for the CI signal — the failure was mine, and it was a real regression rather than a fixture quirk.

Pulling the two array arguments into a shared `xlookup_vector` helper flattened a distinction the original inline code was making on purpose. A `lookup_array` that is not a range fell through to `#N/A`, but a `return_array` that is not a range fell through to `#VALUE!`, and the shared helper gave both the `#N/A` branch.

That is what broke the two cells in the fixture:

| cell | formula | Excel | was |
|---|---|---|---|
| `Errors!H3` | `=XLOOKUP(E2,A:A,TRUE)` | `#VALUE!` | `#N/A` |
| `Errors!I3` | `=XLOOKUP(E2,A:A,3)` | `#VALUE!` | `#N/A` |

`E2` is `India`, which does exist in `A:A`, so the lookup itself succeeds — the only thing wrong is the scalar third argument. An invalid argument is `#VALUE!`, not a failed search, which is the same rule the sheet's other `#VALUE!` cases already follow (`D9` and `D11` for mismatched lengths, `H5`/`I5` for out-of-range `search_mode`).

While fixing that I found a second regression from the same collapse that no fixture covers: a two-dimensional `return_array` such as `=XLOOKUP(1,A1:A3,B1:C3)` returned `#ERROR!` "Second argument must be a vector" instead of the previous `#VALUE!` "Arrays must be of the same size". `#ERROR!` is not an xlsx error at all, so that one was worse than the reported failure.

The fix threads a small `VectorArgument` enum through the helper so each argument reports the error it reported before the refactor. `#N/A` still means "found nothing", `#VALUE!` still means "bad argument", and the array-constant support this PR adds is unchanged.

Four unit tests added in `test_fn_xlookup.rs`: scalar `return_array`, scalar `lookup_array`, mismatched sizes including the 2-D and array-constant forms, and `test_xlookup_not_found_is_still_na`, which pins plain not-found back to `#N/A` so the `#VALUE!` cases cannot swallow it. I checked they are load-bearing rather than assuming — reverting only `xlookup.rs` and keeping the tests fails two of them (`left: "#N/A" right: "#VALUE!"` and `left: "#ERROR!" right: "#VALUE!"`) while the `#N/A` guard stays green in both states.

Verification: `cargo test --workspace test_calc_LOOKUP_AND_REFERENCE_XLOOKUP` passes, `cargo test` is green (2604 passed, 0 failed, 8 ignored), and green again when merged with current `main` (2626 passed, clean merge). For reference the same `cargo test` on unmodified `main` is 2619 passed / 0 failed — the branch is 27 commits behind, which accounts for the difference rather than any test disappearing. `cargo fmt -- --check` and the clippy gate from the Makefile both pass with no warnings. I did not run the npm parts of `make lint` or the js/nodejs/python CI jobs, since the change is Rust-only.

Disclosure: I use AI assistance in my work, and I review and verify everything before it goes out.

PR #1349, opened 2026-08-11
FIX: number format date/time tokens must be case-insensitive

tommy230:
## Problem

A number format code written in upper case renders as `#VALUE!`. Excel treats
format codes as case-insensitive, and upper-case forms are common in files
written both by Excel and by third-party writers. The format lexer matches the
date/time token heads as the literal lower-case characters `'d'`, `'m'`, `'y'`,
`'s'` and consumes repeats the same way, so anything it cannot tokenise falls
through to `#VALUE!`.

### Repro

| formula | Excel | IronCalc `main` |
|---|---|---|
| `=TEXT(45000, "m/d/yyyy")` (control) | `3/15/2023` | `3/15/2023` |
| `=TEXT(45000, "M/D/YYYY")` | `3/15/2023` | `#VALUE!` |
| `=TEXT(45000, "DD-MMM-YY")` | `15-Mar-23` | `#VALUE!` |
| `=TEXT(0.5, "H:MM AM/PM")` | `12:00 PM` | `#VALUE!` |

The same applies to a cell's `numFmt` string, not just `TEXT`.

Every expected value was read back from desktop Excel (Microsoft 365).

## Fix

Match the four token heads as `'d' | 'D'`, `'m' | 'M'`, `'y' | 'Y'`,
`'s' | 'S'`, and continue the repeat loop with `eq_ignore_ascii_case` so a
mixed-case run such as `Mmm` is consumed as one three-character token rather
than as `M` followed by `mm`.

The month-versus-minute disambiguation is untouched: it depends on the
neighbouring tokens, not on case, and Excel behaves the same way.

## Tests

`base/src/test/test_format_case.rs` (new):

- `test_text_uppercase_date_tokens` — `"M/D/YYYY"` matches `"m/d/yyyy"`, and
  `"MM/DD/YYYY"`, `"YYYY"`, `"MMM"`, `"D"`, `"MMM-YY"` all produce the expected
  strings rather than an error.
- `test_text_uppercase_minutes_disambiguation` — `"hh:MM:SS"` matches
  `"hh:mm:ss"`, and `"MM/DD/YYYY hh:MM:SS"` still reads the leading `MM` as
  month and the trailing `MM` as minutes, so the fix does not disturb the
  existing rule.

Both fail on `main` and pass with this change.

### DIFF
--- a/base/src/formatter/lexer.rs
+++ b/base/src/formatter/lexer.rs
@@ -434,11 +434,15 @@ impl Lexer {
                         Token::ILLEGAL
                     }
                 }
-                'd' => {
+                'd' | 'D' => {
                     let mut d = 1;
-                    while let Some('d') = self.peek_char() {
-                        d += 1;
-                        self.read_next_char();
+                    while let Some(c) = self.peek_char() {
+                        if c.eq_ignore_ascii_case(&'d') {
+                            d += 1;
+                            self.read_next_char();
+                        } else {
+                            break;
+                        }
                     }
                     match d {
                         1 => Token::Day,
@@ -447,11 +451,15 @@ impl Lexer {
                         _ => Token::DayName,
                     }
                 }
-                'm' => {
+                'm' | 'M' => {
                     let mut m = 1;
-                    while let Some('m') = self.peek_char() {
-                        m += 1;
-                        self.read_next_char();
+                    while let Some(c) = self.peek_char() {
+                        if c.eq_ignore_ascii_case(&'m') {
+                            m += 1;
+                            self.read_next_char();
+                        } else {
+                            break;
+                        }
                     }
                     match m {
                         1 => Token::Month,       // (or minute)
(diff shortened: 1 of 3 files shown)


elsaminsut (changes requested):
Hey I tested your branch and everything seems to work well! 

To make sure everything's covered, could you add a few cases that test the opposite of the ones you added? Like for example you have: 
```
    model._set("A3", r#"=TEXT(45000,"MM/DD/YYYY")"#);
    model._set("A4", r#"=TEXT(45000,"YYYY")"#);
    model._set("A5", r#"=TEXT(45000,"MMM")"#);
    model._set("A6", r#"=TEXT(45000,"D")"#);
    model._set("A7", r#"=TEXT(45000,"MMM-YY")"#);
```
So that we could test that all these cases return the same value when using lowercase format codes (`mm/dd/yyyy`)

Once those are added, this is ready for approval!

tommy230:
Added the lowercase counterparts in 63e6681 and asserted each one matches its uppercase form. I also reran the focused test locally:

`cargo test -p ironcalc_base test_text_uppercase`

Result: 2 passed, 0 failed.

tommy230:
Added the lowercase counterparts for each of those formats, asserted equal to the uppercase results.  Thanks!

elsaminsut (approved):
Awesome, thanks!

Issue #1153, opened 2026-06-22
[templates] `EDATE`, `PPMT`, `CUMIPMT`, and `CUMPRINC` do not support dynamic arrays


dg-ac:
### Description
`EDATE` and several related financial functions (`PPMT`, `CUMIPMT`, `CUMPRINC`, `PMT`) do not support dynamic arrays. When used with an array input, they fail to spill results across the expected range.

See sheet `Amortization schedule`, columns C, F, H and I in the attached file for a reproduction case:
https://testing.ironcalc.com/?model=INx1M-18FJ1-AB8p8
[mortgage_calculator (4).xlsx](https://github.com/user-attachments/files/29205562/mortgage_calculator.4.xlsx)

### Steps to reproduce
1. Open the attached IronCalc file.
2. Navigate to the `Amortization schedule` sheet.
3. Observe column C, where `EDATE` is used with an array input.
4. Repeat for `PPMT`, `CUMIPMT`, and `CUMPRINC`.
5. Review also `PMT`, in Seetings sheet.

### Expected behavior
All four functions accept array inputs and spill results dynamically across the expected range, consistent with how other functions handle dynamic arrays in IronCalc.

### Actual behavior
The functions do not spill and return a single value or an `#N/IMPL!` error instead of the full array of results.

Note:
This may also be the reason we get a `#CIRC!` in Settings!B24

nhatcher:
Moving to v1.0.0

PR #1351, opened 2026-08-11
FIX: XLOOKUP must accept a computed lookup_array or return_array (#1338)

### LINKED ISSUE #1338: XLOOKUP with a computed lookup_array returns #N/A even when the value is present

tommy230:
Fixes #1338.

## Problem

`XLOOKUP` returns `#N/A` whenever its `lookup_array` or `return_array` is a
computed expression rather than a plain range — the two-column join
`=XLOOKUP("b|y", A1:A2&"|"&B1:B2, C1:C2)`, an array literal, an arithmetic
expression over a range. `MATCH`, `XMATCH`, `VLOOKUP` and `HLOOKUP` all handle
the same computed arrays, so the inconsistency is visible from one cell to the
next.

`fn_xlookup` matches only `CalcResult::Range` for those two arguments. Every
other non-error result, including the `CalcResult::Array` a computed expression
produces, falls through to a catch-all:

```rust
error @ CalcResult::Error { .. } => error,
_ => CalcResult::Error {
    error: Error::NA,                       // lookup_array
    origin: cell,
    message: "Range expected".to_string(),
},
```

So the `#N/A` is not a failed search. It is an argument-type rejection
mislabelled as `#N/A`, emitted before `linear_search` ever runs. That also
explains the `if_not_found` bypass reported in the issue: the fallback is only
consulted on a `None` search result, and this path never reaches the search, so
`=XLOOKUP(key, computed, values, "fallback")` returns `#N/A` — a value that
formula can never produce in Excel.

### Repro

Sheet data: `A1:A2` = `a`, `b` · `B1:B2` = `x`, `y` · `C1:C2` = `first`,
`second` · `D1:D2` = `1`, `2`

| formula | Excel | IronCalc main (measured) |
|---|---|---|
| `=XLOOKUP("b",A1:A2,C1:C2)` | `second` | `second` |
| `=XLOOKUP("b\|y",A1:A2&"\|"&B1:B2,C1:C2)` | `second` | `#N/A` |
| `=XLOOKUP("b\|y",A1:A2&"\|"&B1:B2,C1:C2,"fallback")` | `second` | `#N/A` |
| `=XLOOKUP("z\|z",A1:A2&"\|"&B1:B2,C1:C2,"fallback")` | `fallback` | `#N/A` |
| `=XLOOKUP(4,D1:D2*2,C1:C2)` | `second` | `#N/A` |
| `=XLOOKUP(1,(A1:A2="b")*(B1:B2="y"),C1:C2)` | `second` | `#N/A` |
| `=XLOOKUP("b",{"a";"b"},C1:C2)` | `second` | `#N/A` |
| `=XLOOKUP("b",A1:A2,C1:C2&"")` | `second` | `#VALUE!` |
| `=XLOOKUP("b\|y",A1:A2&"\|"&B1:B2,C1:C2&"!")` | `second!` | `#N/A` |
| `=XLOOKUP(2,D1:D2*1,C1:C2,"nf",0,2)` | `second` | `#N/A` |
| `=SUM(XLOOKUP(D1:D2,D1:D2,D1:D2))` | `3` | `#N/A` |
| `=MATCH("b\|y",A1:A2&"\|"&B1:B2,0)` | `2` | `2` |
| `=VLOOKUP("b\|y",A1:A2&"\|"&B1:B2,1,FALSE)` | `b\|y` | `b\|y` |

Every expected value was read back from desktop Excel (Microsoft 365,
16.105.3). The xlsx fixture added by this PR holds the same formulas, authored
in Excel so the expected values are Excel's own cached results; it fails on
`main` with ten diffs.

The failure is silent rather than loud. Unlike the `#N/IMPL!` family in #1153,
`#N/A` is a plausible everyday lookup result, and an `IFERROR`-wrapped lookup
degrades to a clean, plausible, wrong value.

## Fix

`lookup_array` and `return_array` are normalised into a small `Vector` type
that is either a stored range (coordinates plus length and orientation) or a
materialised array. The existing `linear_search` and the four binary searches
already operate on a `&[CalcResult]`, so they are unchanged.

- Range vectors keep their coordinates, so the **return** side is still read
  one element at a time: a whole-column `return_array` is never materialised,
  which is how `main` behaves today.
- A whole-column or whole-row reference is clamped to the sheet's used area
  exactly as before, now for both arguments rather than only `lookup_array`.
- The size check is now "same length and same orientation", which is
  equivalent to the old row-delta/column-delta comparison for two vectors.
- A computed array that is not a vector (a 2-D array) is rejected the same way
  a 2-D range already was.

The second commit vectorizes `lookup_value`: a range or array lookup value is
applied elementwise and spills an array of the same dimensions, as in Excel.
Static analysis reports the spill shape from `args[0]`. This is separable from
the fix — happy to drop it into its own PR if you prefer.

## Tests

`base/src/test/lookup_and_reference/test_fn_xlookup.rs` (new, 18 tests):
computed lookup_array (text join, numeric, boolean product, array literal),
computed return_array, both computed, `if_not_found` reached on a genuine miss
and *not* reached on a hit, row-oriented computed arrays, mismatched lengths,
binary search over a computed array, whole-column ranges, the unchanged
stored-range and scalar paths, the MATCH/XMATCH/VLOOKUP control group, and the
vectorized `lookup_value` cases.

Measured: 14 of the 18 fail on `main` and all 18 pass on this branch. The four
that pass on both are the control group (`MATCH`/`XMATCH`/`VLOOKUP` over the
same computed array), the unchanged stored-range path, the whole-column case,
and the `ROW` regression guard.

`xlsx/tests/calc_tests/LOOKUP_AND_REFERENCE/XLOOKUP_computed_arrays.xlsx` is
the same repro authored in desktop Excel.

## Note

The issue also lists `=XLOOKUP(2, ROW(A1:A2), C1:C2)`. That one is a different
gap: `ROW(range)` does not spill in IronCalc, it returns the first row as a
scalar, so the `lookup_array` there is a scalar rather than a computed array.
This PR leaves that alone; there is a regression test pinning the current
`ROW` behaviour so the distinction is on the record.


nhatcher:
Thanks! I think there is already a PR for this: #1341  Lte me analyze both :)

tommy230:
Thanks for taking a look! You're right that #1341 and this PR both fix the #1338 repro (computed `lookup_array` / `return_array`), and @youdie006 filed theirs first. To make comparing easier, here's the delta as I see it:

**Where they overlap:** both accept a `CalcResult::Array` for the lookup and return arguments instead of falling through to the `#N/A` / `#VALUE!` catch-alls.

**What this PR does differently:**

1. **Range arguments stay lazy.** Range-backed vectors keep their coordinates, so only the return element the search lands on is evaluated, rather than materialising the whole return column up front. This keeps full-column references (`XLOOKUP(x, A:A, B:B)`) at their current cost.
2. **Array `lookup_value` spills.** When `lookup_value` is itself a range or array, the lookup is applied elementwise and the result spills with the same dimensions, matching Excel. This includes the `static_analysis.rs` change so the engine knows XLOOKUP's spill shape.
3. Adds an `.xlsx` calc-test fixture (`XLOOKUP_computed_arrays.xlsx`) alongside the unit tests.

No preference on which lands as the base. If you'd rather merge #1341 for the core fix, I'm happy to rebase the lazy-range handling and the spill behaviour on top of it as a follow-up. Whatever is easiest for you to review.

PR #1352, opened 2026-08-11
feat(wasm): expose raw cell values via getCellValue

uddhav:
The wasm `Model` currently exposes `getFormattedCellValue` (the display string, which rounds numbers at 15 significant digits, matching spreadsheet display convention) and `getCellContent` (the input text/formula) - but no way to read a cell's **raw evaluated value**. The nodejs binding already has this as `Model.getCellValue` / `getCellValueByRef`, backed by `Model::get_cell_value_by_index`. Browser embedders doing numeric work (charting, exporting, differential testing against reference values) currently can't get the full-precision `f64` without going through the formatted string - e.g. `25498.569618488982` reads back as `"25498.569618489"`.

This PR:

- adds `UserModel::get_cell_value(sheet, row, column) -> Result<CellValue, String>` in `base/src/user_model/common.rs`, an `#[inline]` delegation to `Model::get_cell_value_by_index`, in the same style as the neighbouring `get_formatted_cell_value` / `get_cell_type`;
- exposes it from `bindings/wasm/src/lib.rs` as `getCellValue(sheet, row, column): string | number | boolean | null` (via `unchecked_return_type`), converting `CellValue` to `JsValue` the same way the nodejs binding maps it to `Either3` - `None -> null`;
- adds `base/src/test/user_model/test_get_cell_value.rs` covering number (full-precision `1/3`), string, boolean, empty cell (`CellValue::None`), and the wrong-sheet error path.

Verified locally: `cargo check` and `cargo clippy` clean on `ironcalc_base` and `bindings/wasm`; `cargo test -p ironcalc_base get_cell_value` passes. I didn't run a wasm32-target build locally - the change is target-independent delegation plus `JsValue` construction, but happy to iterate if CI disagrees.

Naming follows the nodejs binding (`getCellValue`); if you'd rather mirror the base name (`getCellValueByIndex`), happy to rename.

### DIFF
--- a/base/src/test/user_model/mod.rs
+++ b/base/src/test/user_model/mod.rs
@@ -17,6 +17,7 @@ mod test_dynamic_arrays;
 mod test_evaluation;
 mod test_fn_formulatext;
 mod test_general;
+mod test_get_cell_value;
 mod test_grid_lines;
 mod test_hidden_columns;
 mod test_keyboard_navigation;
--- a/base/src/test/user_model/test_get_cell_value.rs
+++ b/base/src/test/user_model/test_get_cell_value.rs
@@ -1,0 +1,31 @@
+#![allow(clippy::unwrap_used)]
+
+use crate::cell::CellValue;
+use crate::test::user_model::util::new_empty_user_model;
+
+#[test]
+fn get_cell_value_returns_raw_typed_values() {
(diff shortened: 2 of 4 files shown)


uddhav:
Apologies this PR was not intended. Closing.

PR #1359, opened 2026-08-17
UPDATE: Remove not needed deps in the xlsx package

nhatcher:
I _think_ an xlsx does not this at all.

Only Store/Deflate are used in xlsx files.

Dropping the default features (aes-crypto, bzip2, zstd) also removes C dependencies so the crate builds for wasm32-unknown-unknown.


nhatcher:
FYI: @BrianHung

PR #1363, opened 2026-08-18
feat: publish xlsx import export as a second wasm package

BrianHung:
Publishes XLSX import and export as a second WebAssembly package, so the core `@ironcalc/wasm` bundle stays the same size for projects that do not use it. This addresses the bundle size concern from #379.

### What the packages provide

`@ironcalc/wasm` gives you the core engine (`wasm_bg.wasm`).

`@ironcalc/wasm-xlsx` gives you the same engine plus XLSX import and export (`Model.fromXlsx` / `model.toXlsx`).

Both are built from the wasm crate with wasm-pack. The xlsx APIs are behind a Cargo feature, so the default package does not grow. `@ironcalc/workbook` keeps depending on the smaller package.

This is relevant to #1113 (converting xlsx in the browser). Supersedes the separate-converter approach in #381.

### Usage

```TypeScript
import init, { Model } from "@ironcalc/wasm-xlsx";

await init();

const model = Model.fromXlsx(xlsxBytes, "Workbook1", "en", "UTC", "en");
const xlsxBytesOut = model.toXlsx();
```

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -562,6 +562,12 @@ jobs:
           name: wasm-pkg
           path: bindings/wasm/pkg
           if-no-files-found: error
+      - name: Upload wasm-xlsx pkg artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: wasm-xlsx-pkg
+          path: bindings/wasm/pkg-xlsx
+          if-no-files-found: error
 
   build-test-workbook:
     name: Build and test @ironcalc/workbook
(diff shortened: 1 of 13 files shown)


nhatcher:
That's pretty neat! Thanks for all the work. I'll review and merged with my highest priority

nhatcher (approved):
LGTM!

Many thanks

ghostyfreak:
@nhatcher would be nice to have release soon :)

nhatcher:
> @nhatcher would be nice to have release soon :)

Would ~ October 15th work for you? If you need sooner I can do a special release next week. Wouldn't be a huge issue

PR #1376, opened 2026-08-19
update: in links, show pointer on hover when cmd/ctrl is pressed

dg-ac:
This PR adds the change mentioned in #1336

Hovering links while pressing `cmd` (in MacOS) and `ctrl` (in Windows and Linux) should change the cursor to a `pointer`. It still opens the link in a new tab, that hasn't changed.

### DIFF
--- a/webapp/IronCalc/src/components/WorksheetCanvas/cellLinks.ts
+++ b/webapp/IronCalc/src/components/WorksheetCanvas/cellLinks.ts
@@ -26,7 +26,18 @@ export interface CellLinksOptions {
   tooltipCell: LinkHoverCell | null;
 }
 
-// Pointer cursor while Ctrl/Cmd+click would follow a link
+// Cmd on Mac, Ctrl elsewhere (on Mac Ctrl+click is a right click)
+const isMac =
+  typeof navigator !== "undefined" && navigator.userAgent.includes("Mac");
+
+function hasLinkModifier(event: {
+  ctrlKey: boolean;
+  metaKey: boolean;
+}): boolean {
+  return isMac ? event.metaKey : event.ctrlKey;
+}
+
+// Pointer cursor while the modifier+click would follow a link
 const LINK_CURSOR_CLASS = "ic-worksheet-link-cursor";
 let linkContainer: HTMLElement | null = null;
 let linkHovered = false;
@@ -40,14 +51,14 @@ function updateLinkCursor(): void {
   );
 }
 
-// The container is never focused so Ctrl/Cmd is tracked on the window
+// The container is never focused so the modifier is tracked on the window
 function trackModifierKey(): void {
   if (modifierTracked || typeof window === "undefined") {
     return;
   }
   modifierTracked = true;
   const onKey = (event: KeyboardEvent): void => {
-    modifierDown = event.ctrlKey || event.metaKey;
+    modifierDown = hasLinkModifier(event);
     updateLinkCursor();
   };
   window.addEventListener("keydown", onKey);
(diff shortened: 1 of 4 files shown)


webapp/IronCalc/src/components/WorksheetCanvas/cellLinks.ts
```diff
@@ -26,6 +26,38 @@ export interface CellLinksOptions {
...
+// The container is never focused so Ctrl/Cmd is tracked on the window
+function trackModifierKey(): void {
+  if (modifierTracked || typeof window === "undefined") {
+    return;
+  }
+  modifierTracked = true;
+  const onKey = (event: KeyboardEvent): void => {
+    modifierDown = event.ctrlKey || event.metaKey;
+    updateLinkCursor();
+  };
```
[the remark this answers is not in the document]
dg-ac: Very nice catch, didn't know about the Ctrl+click on Mac! Fixed.

webapp/IronCalc/src/components/WorksheetCanvas/cellLinks.ts
```diff
@@ -76,15 +121,20 @@ export class CellLinks {
...
     };
     container.onpointerleave = () => {
+      linkHovered = false;
+      updateLinkCursor();
       this.options.onLinkHover?.(null);
     };
-    // Ctrl+click (Cmd+click on Mac) on a link cell follows the link. Only the
-    // primary button: on Mac Ctrl+click is a right click (context menu).
+    // Cmd+click (Ctrl+click outside Mac) on a link cell follows the link,
+    // primary button only
```
nhatcher, line 135: I really preffer the notation of Linux/Windows as the general rule and Mac as a special case.

webapp/IronCalc/src/components/WorksheetCanvas/cellLinks.ts
```diff
@@ -26,6 +26,49 @@ export interface CellLinksOptions {
...
+  const onKey = (event: KeyboardEvent): void => {
+    modifierDown = hasLinkModifier(event);
+    updateLinkCursor();
+  };
+  window.addEventListener("keydown", onKey);
+  window.addEventListener("keyup", onKey);
+  window.addEventListener("blur", () => {
+    modifierDown = false;
+    updateLinkCursor();
+  });
```
nhatcher, line 69: Are you sure this is ok? You are attching events to window, they will be there forever.

webapp/IronCalc/src/components/WorksheetCanvas/cellLinks.ts
```diff
@@ -26,6 +26,49 @@ export interface CellLinksOptions {
   tooltipCell: LinkHoverCell | null;
 }
 
+// Cmd on Mac, Ctrl elsewhere (on Mac Ctrl+click is a right click)
+const isMac =
+  typeof navigator !== "undefined" && navigator.userAgent.includes("Mac");
```
nhatcher, line 31: We really shouldn't be doing this. This is a can of worms

nhatcher:
Hi @dg-ac , you are introducing here some dangerous precedents, are you susre we even need this?

dg-ac:
@nhatcher Well, _need_ is a pretty big word... we can survive without for sure, but UX-wise it is a good practice to use `cursor: pointer` for hyperlinks. If it brings more issues than benefits we can of course close it, your call.

nhatcher:
This is really nice but a bit too much for a small thing. Let's find a better way if there is one.

PR #1393, opened 2026-08-25
update: improve testing for EOMONTH

elsaminsut:
Extend existing xlsx and unit tests for EOMONTH function. 

Includes 2 small fixes:
- Reject booleans in date argument
- Serial number 0 resolves to 1 Jan 1900 to match Excel behavior


base/src/functions/date_and_time.rs
```diff
@@ -653,9 +653,17 @@ impl<'a> Model<'a> {
             }
             Err(s) => return s,
         };
-        let date = match self.excel_date(serial_number, cell) {
-            Ok(d) => d,
-            Err(e) => return e,
+        // Serial number 0 (e.g. an empty cell or omitted argument) is not a
+        // real calendar day, but Excel still resolves it to January 1900 for the
+        // purposes of month arithmetic, so we special case it here.
+        let date = if serial_number == 0 {
```
nhatcher: Anyway you can extract this logic and explain what you are doing in a more genarl setting? Also is this done anywhere else?

nhatcher (changes requested):
Quick question

base/src/functions/date_and_time.rs
elsaminsut: Other functions treat serial date 0 as 0 January 1900 (see #1400, where I didn't try to fix it) but EOMONTH specifically resolves it to Jan 1st so that `=EOMONTH(0, 0)` returns 31, creating a 1-day mismatch.

I saw how the start serial date is defined in [base/src/formatter/dates.rs](https://github.com/ironcalc/IronCalc/blob/d65a5492514edc3dea210b15098617bb3c30417d/base/src/formatter/dates.rs#L35) and applied the same logic for a date 0.

Where should this logic live otherwise?

elsaminsut:
I removed the serial date 0 fix from this PR and wrote an issue about it (#1409) and we can discuss at a later point what happens on day 0 more generally.

nhatcher (approved):

PR #1422, opened 2026-09-17
fix: quote sheet names containing any non-identifier character

getdom:
## Problem

`name_needs_quoting` (`base/src/expressions/utils/mod.rs`) uses a denylist of characters (space, parentheses, quote, `$ , ; - + { }`). Any other special character slips through, so a sheet named `P&L` — very common in financial models — is serialized as `=P&L!A1`, which is then read back as `P & L!A1` and evaluates to `#NAME?`.

Reproduction (in-memory):

```rust
let mut m = Model::new_empty("t", "en", "UTC", "en")?;
m.add_sheet("P&L")?;
m.set_user_input(1, 1, 1, "42".into())?;
m.set_user_input(0, 1, 1, "='P&L'!A1".into())?;
assert_eq!(m.get_cell_formula(0, 1, 1)?, Some("='P&L'!A1".into())); // was: Some("=P&L!A1")
```

Loading an xlsx with a `P&L` sheet shows the same thing: every cross-sheet reference to it becomes `#NAME?` after import.

## Fix

Excel leaves a sheet name unquoted only when it is made of letters, digits, `_` and `.`, and does not start with a digit or a `.`. This switches the check to that allowlist. An empty name stays unquoted so `ADDRESS(…, "")` keeps its bare `!` prefix (existing test `test_address_empty_sheet_text`).

Adds tests for `P&L`, `Q1=Q2`, `Sheet!`, `100%`, `.hidden` (quoted) and `Été`, `_draft`, `Sheet.1` (not quoted). `cargo test -p ironcalc_base`: 2451 passed.


nhatcher (approved):
Thanks! I had this on the back of my mind for a long time, thanks for taking care of it

PR #1424, opened 2026-09-17
FIX: Enforce Linux Like endings (CR)

nhatcher:
FYI: @elsaminsut: Windows uses a different way to end lines than POSIX: https://en.wikipedia.org/wiki/Newline

This PR normalizes everythong to the [POSIX standard](https://en.wikipedia.org/wiki/POSIX)


nhatcher:
Note to others, if you have an old checkout just do:

```
git add --renormalize .
```

PR #1425, opened 2026-09-17
fix(import): a typed cell without a value is empty, not 0

getdom:
## Problem

Many writers emit a cell that only carries a style as `<c r="C10" s="1" t="n"/>` — openpyxl does it for every formatted blank cell, and LibreOffice reads such files fine. The importer keeps the declared type when there is no `<v>`, so the cell becomes `NumberCell(0)`. A model with percentage formats on blank cells shows `0.0%` everywhere instead of nothing.

Excel treats a `<c>` with no `<v>` and no inline string as blank, whatever `t` says.

## Fix

In `get_cell_from_excel`, for non-formula cells: no value and no inline string → `EmptyCell { s }`. Formula cells without a cached value keep their current path (they get evaluated anyway).

Adds a unit test covering `t` in `n`, `b`, `str`, `s`, `e`, plus the `<v>0</v>` case which must stay a number. `cargo test -p ironcalc`: 42 + 226 passed.


nhatcher (approved):
LGTM!

Interesting. I did not know that

